PR #73, opened 2022-03-02
Add POC: Embark in Racoon :)

robinandeer:
## Table of contents

1. [Summary](#summary)
1. [Background](#background)
1. [Design](#design)
1. [Alternatives considered](#alternatives-considered)

## Summary

A new approach to Embark 💎 where the business logic is centrally located on an API server.

## Background

Currently each Embark client has to implement core business logic (parsing Angel data, determining next passage, calling GraphQL queries/mutations, etc.) This logic is exactly the same no matter the client, however, there's no way to share it across platforms.

Even though an Angel story can run on any client, the behavior is not guaranteed to be the same. Does all client handle chained redirects? Do they evaluate conditionals the same way?

Frontends/clients naturally diverge when it comes to UI, navigation, and user input. However, the rest of the logic could potentially be moved elsewhere that all clients can access centrally.

Testing Embark is also a pain point since at least the web solution is so integrated that you have to render the UI to perform tests on the core logic.

## Design

The aim of this proposal is to centralize critical logic to an API server.

### Goals

- Support existing Angel data format

- Clearly separate client vs. shared logic

  - Clients should only recieve the data they actually need.

- Avoid duplicating critical business logic for different clients

  - It should be possible to automatically test stories without rendering UI

- Design a unified "Embark API" to support server, client, and a future editor!

  - If we change something in the Angel data format, we should only need to make updates in one place - the "Embark API"!

### API responsibilities

- Parse Angel JSON/XML
- Determine next passage
- Determine updates to response history
- Call GraphQL queries/mutations

### Client responsibilities

- Fetch and render the passages
- Submit user input for current passage action

### Key concepts

- Angel + Angel data: stays the same

- Embark Core: parsing, critical logic (minus persistence)

- Client REST API

- Persistence

  > The web solution uses cookie storage to persist responses but this is not required. Each client can implement their own persistence layer OR we can use something like a central database. This "persitance layer" should support 4 functions: 1) create new session/initialize history. 2) add/update history. 3) go back/pop history. 4) clear session/history.

#### Client REST API

I've design a REST API but it should be obvious how it could work in a GraphQL context.

- `POST /api/embark/start`: start new session for a given story

- `GET /api/embark/passage`: get next passage

- `POST /api/embark/submit/[passage]`: submit user input for a given passage

- `POST /api/embark/go-back`: remove last user input

#### Persistence API

This could be up to each client to implement. Current version uses cookie storage.

- `create`: create new session
- `save`: update session with given history
- `get`: retrieve session

## Alternatives considered

A 2022 maker week project was initiated to see if it was possible to use Kotlin Multiplatform to cross-compile this critical logic to each platform. However, the conclusion was the the technology is not mature enough.

## How you can test the API

1. Install [`httpie`](https://httpie.io/) - `brew install httpie`

1. Get the initial passage

    ```bash
    http --session=test http://localhost:3000/api/embark/web-onboarding-no-contents/passage
    ```

1. Submit user response to "Start" passage

    ```bash
    http --session=test --form http://localhost:3000/api/embark/web-onboarding-no-contents/Start value=true
    ```

1. Get the next passage

    ```bash
    http --session=test http://localhost:3000/api/embark/web-onboarding-no-contents/passage
    ```

1. Repeat 🤑


robinandeer:
No longer relevant :)

PR #249, opened 2022-06-22
Feat/bundles

simonauner:
## Describe your changes

This PR adds support for bundles in the mocked store.
I added Perils to the mocked product service to show how I imagine we differentiate between data from a "storyblok CMS" and PCMS - where PCMS holds the actual information about each insurance. 

So when browsing insurances we show which insurances are contained in a bundle and which Perils that belong to each insurance. When it's not a bundle we just show the Perils as they are.

https://user-images.githubusercontent.com/1343979/174958765-878e5c75-b47a-4ec3-90cf-77ef7b672dbe.mov

To explore how we work with bundles.

### DIFF
--- a/apps/store/src/services/mockCmsService.ts
+++ b/apps/store/src/services/mockCmsService.ts
@@ -11,8 +11,8 @@ export type CmsProduct = {
   displayName: string
   slug: string
   pageTitle: string
-  product: string
   form: PriceForm
+  products: string[]
 }
 
 const CMS_PRODUCTS: CmsProduct[] = [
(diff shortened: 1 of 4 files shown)


robinandeer (approved):
ship it

apps/store/src/services/mockCmsService.ts
```diff
@@ -31,9 +31,18 @@ const CMS_PRODUCTS: CmsProduct[] = [
...
-    product: 'SE_CAR',
+    products: ['SE_CAR'],
     form: SWEDEN_CAR_FORM,
   },
+  {
+    name: 'SE_HOME_ACCIDENT_BUNDLE',
+    market: MarketLabel.SE,
+    displayName: 'Home and accident insurance', // TODO: should be a translation key (or translated from BE)
+    slug: 'home-accident',
+    pageTitle: 'Home & accident insurance | Hedvig', // TODO: should be a translation key (or translated from BE)
```
gustaveen, line 42: I would say the `pageTitle` should come from the CMS right?

gustaveen (approved):
👌

apps/store/src/services/mockCmsService.ts
simonauner: Yes, that's how I see it as well, as long as they have the same locale-per-market setup... If they only have the market dimension it needs to be a translation key.

PR #250, opened 2022-06-23
Feat/new bundle handling

simonauner:
## Describe your changes

1. [Add some global styles to list and anchor](https://github.com/HedvigInsurance/racoon/commit/64171e0aa079595e8748f37d8e95e4204348cfda)
2. [Add Text component](https://github.com/HedvigInsurance/racoon/commit/0a94a72a1e1b160074bdb1d1fbebabf2fabf9437)
3. [Consider bundles to be separate products](https://github.com/HedvigInsurance/racoon/commit/aed2a3b99254e2743f75da646efa463448ea2c9a)
    To solve the issue of "multiple insurances in a product" I added `insurances` property to a Product. This is what we today consider to be a "PCMS Insurance"

1. It makes it much easier to see what you code instead of showing everything as plain text
2. I just want to have smaller text sometimes
3. This is the latest results from our discussions - consider a bundle to be a product in itself, just like everything else.  

https://user-images.githubusercontent.com/1343979/175278430-e7714d86-6fd5-42e1-8428-c5313b7af491.mov

### DIFF
--- a/apps/store/src/components/Text/Text.tsx
+++ b/apps/store/src/components/Text/Text.tsx
@@ -1,0 +1,24 @@
+import styled from '@emotion/styled'
+import { ReactNode } from 'react'
+
+type FontSize = 's' | 'm' | 'l' | 'xl'
+
+type TypographyProps = {
+  size: FontSize
+  children: ReactNode
+}
+
+const fontSizes: Record<FontSize, string> = {
+  s: '0.875rem',
+  m: '1rem',
+  l: '1.125rem',
+  xl: '1.25rem',
+}
+
+const TextElement = styled.span(({ size }: TypographyProps) => ({
+  fontSize: fontSizes[size],
+}))
+
+export const Text = ({ size = 'm', children }: TypographyProps) => {
+  return <TextElement size={size}>{children}</TextElement>
+}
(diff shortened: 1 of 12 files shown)


apps/store/src/components/Text/Text.tsx
```diff
@@ -0,0 +1,24 @@
...
+  m: '1rem',
+  l: '1.125rem',
+  xl: '1.25rem',
+}
+
+const TextElement = styled.span(({ size }: TypographyProps) => ({
+  fontSize: fontSizes[size],
+}))
+
+export const Text = ({ size = 'm', children }: TypographyProps) => {
```
guilhermespopolin, line 22: 🤔 If you providing a default value for `size` prop with the idea of enabling the usage of `<Text>` without the need of passing that prop, then maybe `size` should be defined as option in `TypographyProps`.

guilhermespopolin (approved):
Thanks for helping the review process by splitting the added changes into their respective commits :)

apps/store/src/components/Text/Text.tsx
simonauner: But it _is_ in `TypograpyProps` 🙃

PR #253, opened 2022-06-27
GRW-1161 / Fix SEO bugs in onboarding

amandaflod:
## Describe your changes

To get the correct Html lang attribute, I had to map over the locale from Next. 

[GRW-1161]: https://hedvig.atlassian.net/browse/GRW-1161?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/onboarding/src/pages/_document.tsx
+++ b/apps/onboarding/src/pages/_document.tsx
@@ -1,11 +1,22 @@
 import { fonts } from '@hedviginsurance/brand'
 import Document, { Head, Html, Main, NextScript } from 'next/document'
+import { locales, LocaleLabel } from '@/lib/l10n/locales'
 import { gtmDevScript, gtmProdScript } from '@/services/analytics/gtm'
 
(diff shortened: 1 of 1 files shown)


gustaveen (approved):
Great job Amanda! ⭐️  Much better than my solution :D

gustaveen:
Or maybe I was a bit too quick :D This was the problem I ran into as well. It get's the right value on initial load but if we change locale client-side (by clicking the link in the header), it sets the wrong htlm lang value. So I think we need to make sure it updates correctly then as well. Any ideas @robinandeer ? 


https://user-images.githubusercontent.com/6661511/175914563-40ff3c27-6628-496f-9a2d-a1875e323ee4.mov

amandaflod:
Oh no! I hadn't tested that 😅 Will have a think 🤔

PR #255, opened 2022-06-27
Fix / Move quote cart creation to client

robinandeer:
## Describe your changes

Remove code in middleware.

The middleware implementation fails to often. Safer to do it client side.
That will also give us faster loading of the page.


guilhermespopolin (approved):
If you don't mind, could you explain shortly why it fails more often when executed in the middleware?

PR #258, opened 2022-06-30
refact(CartList): removes component level integration between product and cms prodcut

guilhermespopolin:
## Describe your changes

Read title ☝🏻 

The idea was to get ride of the code that connects info between `product` and `cmsProduct` from a component level. That's because [`getProductByMarketAndProduct`](https://github.com/HedvigInsurance/racoon/compare/refact/cart-list?expand=1#diff-c29933a30df352a916eeeb7754b07a04f2c954421e829fc29f2c0401bf55eb87L45) represents what would be one day asynchronous code and we probably don't want to fetch data from a component level when it's not necessary.


robinandeer (approved):
Very nice :D

PR #262, opened 2022-07-01
Discussion: adoption of msw

guilhermespopolin:
## What

* Refactored a `mockCmsService` into a real `cms` service that fetches data from an API
* Mocked a `cms` api which will intercept requests to the CMS API and return mocked data.
* Added a simple integration test for _Product Page_; That page level component makes usage of `cms` service to fetch some data about products. However the same mock information is used for the test.

## Why

This is about an experimentation of using `msw` and API mocks during the development process of _Coverage Store_. 
The reasoning behind can splitted as follow:

### Experimenting

We could use it in order to experiment API design decisions before the real API gets implemented. By mocking an a completely new API or changes on an existing one we could get a sense of how the application would behave had it been written differently, without having to actually write/rewrite it.

### Development

Established an agreement between FE and BE teams about the API schema, we could work in parallel with BE team by mocking an API that follows it. In that way we'd be writing application code that fetches data as the API already exists while the BE works on the implementation of the real one. As soon the real API becomes available the transition between the mocked one to the real one would require minimal code changes since both API share the same schema.

### Debug

Most of the client-side issues are related to data. Invalid API call, missing error handling, or receiving an unexpected response—such scenarios happen daily during both development and production. API mocking allows developers to model an exact API interaction that causes an issue. It's efficient in both debugging and fixing the problem, as one can mock a successful scenario easily as well.

More info about `msw` can be found in the oficial [docs](https://mswjs.io/).


apps/store/src/services/cms/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from './types'
```
robinandeer, line 1: nit: Didn't we agree to skip the "index.ts" re-export modules?

robinandeer (approved):
Great stuff!

[lackoftests](https://app.graphite.dev/macros/HedvigInsurance)

guilhermespopolin:
> Great stuff!
> 
> [lackoftests](https://app.graphite.dev/macros/HedvigInsurance)

Yeah, it's just I don't exactly what to test with the addition of msw. I mean, it adds a mock server. We could use that mock server for testing but at this moment all of our data fetching comes from Next page components (`getServerSideProps`and so on). So I have added a new dependency to support the testing of Next page components: [next-page-tester](https://www.npmjs.com/package/next-page-tester). But at the end I was testing Next implementation details which is not nice. 

I can write integration tests for `ProductPage` page directly by passing data as props by cutting the middleman: `[product].tsx` and by doing it I don't need  a mock server since no request will be made. But since I'd wouldn't rely on msw there's no point of adding it as part of this PR.

But maybe I'm missing something 🤔 Do you have a suggestion on for what to write tests at this moment?

apps/store/src/services/cms/index.ts
guilhermespopolin: That's true. It's a bad habit of mine

PR #279, opened 2022-07-12
Feat/storyblok setup continue

gustaveen:
## Describe your changes
* Add Storyblok preview mode
* Add initStoryblok func
* Add editor.html to fix http issue for using storyblok admin v2
* Render bloks from Storyblok

[GRW-1260]

[GRW-1260]: https://hedvig.atlassian.net/browse/GRW-1260?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/store/src/services/storyblok.ts
```diff
@@ -0,0 +1,34 @@
...
+    version: preview ? 'draft' : 'published',
+  })
+  return data.story
+}
+
+export const getAllLinks = async () => {
+  const storyblokApi = getStoryblokApi()
+  const { data } = await storyblokApi.get('cdn/links/')
+  return data.links
+}
```
gustaveen, line 34: Should we namespace all functions in a service? @robinandeer It looks like we mix a it so far

robinandeer (approved):

apps/store/src/services/storyblok.ts
robinandeer: Hmm like I mentioned this morning, as long as we keep client vs. server components separate so we can support code splitting I think that might be nice :)

apps/store/public/editor.html
```diff
@@ -0,0 +1,13 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <title>Storyblok Admin</title>
+  </head>
+  <body>
+    <div id="app"></div>
+    <script type="text/javascript">
+      STORYBLOK_PREVIEW_URL = 'http://localhost:3000/'
```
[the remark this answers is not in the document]
gustaveen: True that, let's have a look

PR #281, opened 2022-07-14
Dynamic localhost port for editor.html

gustaveen:
## Describe your changes
Dynamic localhost port for editor.html


robinandeer (approved):
Nice, didn't think about that solution

PR #292, opened 2022-07-15
Add new Heading component with variants

gustaveen:
## Describe your changes
New Heading component to support the new fonts. Since the serif font is actually different fonts depending on size and letter-spacing varies between standard and serif, I combined them into variants.

We could split up the `fontSize` (where we pass in the theme fontSize values instead) and `fontFamily` into two props. But all sizes shouldn't be supported in the serif fonts (the serif font is not supposed to be used smaller than 20px). 

https://user-images.githubusercontent.com/6661511/179250740-377acd3a-4b21-4350-8850-a7eb24134f24.mov

typography

### 
Next steps:
- Add media query support

### DIFF
--- a/packages/ui/src/components/HeadingNew/HeadingNew.tsx
+++ b/packages/ui/src/components/HeadingNew/HeadingNew.tsx
@@ -8,27 +8,23 @@ export type HeadingProps = Margins & {
   as: 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6'
   color?: 'dark' | 'light'
   children: React.ReactNode
-  variant: HeadingVariant
+  variant?: HeadingVariant
 }
 
 type HeadingBaseProps = Pick<HeadingProps, 'color' | 'variant'> & Margins
 
-const HeadingBase = styled.h2<HeadingBaseProps>(({ theme, color, variant, ...props }) => ({
-  color: getColor(color),
-  fontFamily: theme.fonts.heading,
-  fontWeight: 400,
-  lineHeight: 1.2,
-  ...getMargins(props),
-  ...getHeadingVariant(variant, theme),
-}))
+const HeadingBase = styled.h2<HeadingBaseProps>(
+  ({ theme, color, variant = 'standard.32', ...props }) => ({
+    color: getColor(color),
+    fontFamily: theme.fonts.heading,
+    fontWeight: 400,
+    lineHeight: 1.2,
+    ...getMargins(props),
+    ...getHeadingVariant(variant, theme),
+  }),
+)
 
-export const Heading = ({
-  as,
-  color,
-  children,
-  variant = 'standard.32',
-  ...rest
-}: HeadingProps) => {
+export const Heading = ({ as, color, children, variant, ...rest }: HeadingProps) => {
   return (
     <HeadingBase as={as} color={color} variant={variant} {...rest}>
       {children}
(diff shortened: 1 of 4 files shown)


packages/ui/src/components/HeadingNew/HeadingNew.tsx
```diff
@@ -0,0 +1,37 @@
...
+import React from 'react'
+import { getColor } from 'ui'
+import { getMargins, Margins } from '../../lib/margins'
+import { getHeadingVariant, HeadingVariant } from './HeadingNew.helpers'
+
+export type HeadingProps = Margins & {
+  as: 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6'
+  color?: 'dark' | 'light'
+  children: React.ReactNode
+  variant: HeadingVariant
```
robinandeer, line 11: Shouldn't this be optional? It has a default value when destructured it seems 🤔

packages/ui/src/components/HeadingNew/HeadingNew.tsx
```diff
@@ -0,0 +1,37 @@
+import styled from '@emotion/styled'
+import React from 'react'
+import { getColor } from 'ui'
+import { getMargins, Margins } from '../../lib/margins'
```
robinandeer, line 4: nit: can this be changes to an "alias"-import?

robinandeer (approved):
Very nice!

I wouldn't mind if you renamed the existing Heading component to "HeadingOld" actually 😉

guilhermespopolin (approved):
I was checking the references for the old `Heading` and this seems to be the places we're using it:

image

I understand it's 12 files to change but maybe it would be worth to updated `Heading` instead of creating a new one. Only 4 of the occurrences are present in something we have released today, which is `onboarding`. Most of them are present in `store` which is on its initial phase so I guess even though being 12 files it's low risk, sort of speak.

packages/ui/src/components/HeadingNew/HeadingNew.tsx
gustaveen: True, good catch! 👍

gustaveen:
> I was checking the references for the old `Heading` and this seems to be the places we're using it:
> 
> I understand it's 12 files to change but maybe it would be worth to updated `Heading` instead of creating a new one. Only 4 of the occurrences are present in something we have released today, which is `onboarding`. Most of them are present in `store` which is on its initial phase so I guess even though being 12 files it's low risk, sort of speak.

Right you are @guilhermespopolin 🙌 My thought was to switch it out in a separate PR :) One thing that we might need to add is support to scale over breakpoints before we switch it out in onboarding. So what do you think, should I fix that before merging and switch it out all occurrences or do it in a new PR :) @guilhermespopolin

guilhermespopolin:
> > I was checking the references for the old `Heading` and this seems to be the places we're using it:
> > I understand it's 12 files to change but maybe it would be worth to updated `Heading` instead of creating a new one. Only 4 of the occurrences are present in something we have released today, which is `onboarding`. Most of them are present in `store` which is on its initial phase so I guess even though being 12 files it's low risk, sort of speak.
> 
> Right you are @guilhermespopolin 🙌 My thought was to switch it out in a separate PR :) One thing that we might need to add is support to scale over breakpoints before we switch it out in onboarding. So what do you think, should I fix that before merging and switch it out all occurrences or do it in a new PR :) @guilhermespopolin

In a different PR sounds better :)

PR #297, opened 2022-07-19
Feat / Improve lint setup

gustaveen:
## Describe your changes
Improve typescript linting (to warn for unused-vars and such) and fix warnings and errors

Improve code styling

Stuff to improve
- Eslint doesn't seem to highlight warnings in VSCode still. Does that work for you?

### How it looks in racoon (only ts warning)
Screenshot 2022-07-19 at 13 34 31

### In market-web (eslint warning)
Screenshot 2022-07-19 at 13 43 14

### DIFF
--- a/packages/scripts/.eslintrc.js
+++ b/packages/scripts/.eslintrc.js
@@ -1,6 +1,7 @@
 module.exports = {
-  extends: ['next/core-web-vitals', 'prettier'],
-  plugins: ['import', 'testing-library', 'jest'],
+  parser: '@typescript-eslint/parser',
+  extends: ['next/core-web-vitals', 'prettier', 'plugin:@typescript-eslint/recommended'],
+  plugins: ['import', 'testing-library', 'jest', '@typescript-eslint'],
   settings: {
     next: {
       rootDir: ['apps/*/', 'packages/*/'],
--- a/packages/scripts/.eslintrc.js
+++ b/packages/scripts/.eslintrc.js
@@ -2,11 +2,11 @@ module.exports = {
   parser: '@typescript-eslint/parser',
   extends: [
     'next/core-web-vitals',
+    'eslint:recommended',
     'plugin:@typescript-eslint/recommended',
-    'plugin:@typescript-eslint/eslint-recommended',
-    'prettier',
+    'prettier', // Uses eslint-config-prettier to turn off all rules that are unnecessary or might conflict with Prettier
   ],
-  plugins: ['import', 'testing-library', 'jest', '@typescript-eslint', 'prettier'],
+  plugins: ['import', 'testing-library', 'jest', '@typescript-eslint'],
   settings: {
     next: {
       rootDir: ['apps/*/', 'packages/*/'],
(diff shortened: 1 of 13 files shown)


packages/scripts/.eslintrc.js
```diff
@@ -1,6 +1,12 @@
...
-  extends: ['next/core-web-vitals', 'prettier'],
-  plugins: ['import', 'testing-library', 'jest'],
+  parser: '@typescript-eslint/parser',
+  extends: [
+    'next/core-web-vitals',
+    'plugin:@typescript-eslint/recommended',
+    'plugin:@typescript-eslint/eslint-recommended',
+    'prettier',
+  ],
+  plugins: ['import', 'testing-library', 'jest', '@typescript-eslint', 'prettier'],
```
robinandeer, line 9: Hmm should we really run prettier as an eslint plugin?

Seems like this is not recommended by prettier:
https://prettier.io/docs/en/integrating-with-linters.html#notes

It has worked well for me so far -- do we need it as a plugin for some reason?

robinandeer (commented):
...otherwise looks great! Thanks for looking into this 🙏

packages/scripts/.eslintrc.js
gustaveen: Ah right you are, didn't really grasp what implemented what package. Feels like shooting in the blind setting up eslint configs sometimes haha!

Ah so you get the eslint warnings in VSCode @robinandeer ? If so it's probably some local setting I have

gustaveen:
> ...otherwise looks great! Thanks for looking into this 🙏

Thanks! Me thinking " this will just take 10 minutes" LOL

packages/scripts/.eslintrc.js
guilhermespopolin: For me it wasn't showing warnings. Then I've opened one particular project in vscode instead the whole monorepo: `code racoon/apps/store` and it magically started to display warnings 🤷🏻‍♂️

robinandeer: oh whuutt 😅 

well I just mean that the prettier formatting has worked -- I also noticed the issue not seeing warnings :/

packages/scripts/.eslintrc.js
```diff
@@ -37,6 +43,11 @@ module.exports = {
         pathGroupsExcludedImportTypes: ['internal'],
       },
     ],
+    '@typescript-eslint/no-empty-function': 'off',
+    '@typescript-eslint/ban-ts-comment': 'off',
+    '@typescript-eslint/no-explicit-any': 'off',
+    '@typescript-eslint/no-empty-interface': 'off',
+    '@typescript-eslint/no-non-null-assertion': 'off',
```
robinandeer, line 50: nit: I would be okey leaving this on ;)

robinandeer (approved):
[birdbox](https://app.graphite.dev/macros/HedvigInsurance)

packages/scripts/.eslintrc.js
gustaveen: For sure, I think thats should be the aim. 🙌 I turned some of them off since it outputs warnings/errors on some of the store code given that we have taken some shortcuts when prototyping :) @robinandeer

PR #298, opened 2022-07-19
Feature / Setup GraphQL codegen

robinandeer:
## Describe your changes

Here we go again with all the seemingly irrelevant dev dependencies that we just need to run the codegen 🙄

Setup non-Apollo GraphQL client.
Add .env file with defaults.

Setup is similar to onboarding except:

- The graphql endpoint is referenced by env var in the codegen config. This requires us to use a `.env` file. In Next.js this is the file we should use to set default environment variables so I think it fits pretty well :)
- I put a "graphql" folder outside "src" to group all mutation/query definitions -- good/bad?

Type safety.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,7 +26,6 @@ yarn-error.log*
 tsconfig.tsbuildinfo
 
 # local env files
-.env
 .env.local
 .env.development.local
 .env.test.local
--- a/apps/store/.env
+++ b/apps/store/.env
@@ -1,0 +1,3 @@
+# Default environment variables
+
+NEXT_PUBLIC_GRAPHQL_ENDPOINT=https://apollo-router.dev.hedvigit.com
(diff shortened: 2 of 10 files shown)


guilhermespopolin (approved):
Nice work :)
About the placement for `graphql` folder I think I'd keep it under `src` as we have for `onboarding`. There's no real argument against it; it's just personal preference 😅 
But honestly, both options are fine 😄

gustaveen (approved):
Great! I think we can put `/graphql` under `src` or do we feel like we want to separate it from the rest of the codebase?

robinandeer (commented):
I moved it outside of `/src` since I didn't think it's actual source code 🤔

It's more like config we use to generate the source code -- but I don't mind moving it inside src again :)

robinandeer (commented):
Okey I will move it in my next PR!

robinandeer:
**Graphite Merge Job**

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/298?utm_source=stack-comment).


Job Reference: Jl1125ldcu9HBt83oi9D

PR #301, opened 2022-07-20
Feature / Add cart notification component

robinandeer:
## Describe your changes

Add purchase notification (replaces mini cart).
Add dependency on radix dialog.

I used a dialog / modal since I think that's what it really is. I was considering a toast but I think this is more accurate.
I didn't add the close button/icon just since the designs are very early stage and it works to "click outside".

I also didn't setup a very component to pull everything together in favor of a more flexible approach with individual components. When we notice we have a stable component that's well defined we can build one.

[GRW-1265]: https://hedvig.atlassian.net/browse/GRW-1265?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/store/src/components/CartNotification/Dialog.tsx
+++ b/apps/store/src/components/CartNotification/Dialog.tsx
@@ -1,0 +1,55 @@
+import { keyframes } from '@emotion/react'
+import styled from '@emotion/styled'
+import * as DialogPrimitive from '@radix-ui/react-dialog'
+
+const overlayShow = keyframes({
+  '0%': { opacity: 0 },
+  '100%': { opacity: 1 },
+})
+
+const contentShow = keyframes({
+  '0%': { opacity: 0, transform: 'scale(.96)' },
+  '100%': { opacity: 1, transform: 'scale(1)' },
+})
+
+const StyledOverlay = styled(DialogPrimitive.Overlay)(({ theme }) => ({
+  backgroundColor: 'rgba(0, 0, 0, 0.6)',
+  position: 'fixed',
+  inset: 0,
+  '@media (prefers-reduced-motion: no-preference)': {
+    animation: `${overlayShow} 150ms cubic-bezier(0.16, 1, 0.3, 1) forwards`,
+  },
+}))
+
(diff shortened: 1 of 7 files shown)


robinandeer:
* main
  * **PR #301** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/301" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/301?utm_source=stack-comment).

apps/store/src/components/CartNotification/Dialog.tsx
```diff
@@ -0,0 +1,55 @@
...
+const contentShow = keyframes({
+  '0%': { opacity: 0, transform: 'scale(.96)' },
+  '100%': { opacity: 1, transform: 'scale(1)' },
+})
+
+const StyledOverlay = styled(DialogPrimitive.Overlay)(({ theme }) => ({
+  backgroundColor: 'rgba(0, 0, 0, 0.6)',
+  position: 'fixed',
+  inset: 0,
+  '@media (prefers-reduced-motion: no-preference)': {
```
guilhermespopolin, line 19: There are some implementations for CSS reset that disables animation globally so you don't need to thing checking if you're able to use them. The following snippet was taken from [here](https://piccalil.li/blog/a-modern-css-reset/)

```
/* Remove all animations, transitions and smooth scroll for people that prefer not to see them */
@media (prefers-reduced-motion: reduce) {
  *,
  *::before,
  *::after {
    animation-duration: 0.01ms !important;
    animation-iteration-count: 1 !important;
    transition-duration: 0.01ms !important;
    scroll-behavior: auto !important;
  }
}
```

Maybe we can think about adding it in our `<GlobalStyles />`. What do you thinK?

guilhermespopolin (approved):
Nice. It's awesome working with UI primitives like radix 🚀

robinandeer (commented):
@guilhermespopolin yeah I really like it too! They should be part of the web platform 😌

PR #309, opened 2022-07-22
 added cart card

dendoesit:
## Describe your changes

Added cart card in storybook

For the cart page.


packages/ui/src/components/CartCard/CartCard.tsx
```diff
@@ -0,0 +1,59 @@
+import styled from '@emotion/styled'
+import { colorsV3 } from '@hedviginsurance/brand'
+
+const ProductCard = styled.div({
+  padding: '0 1rem 1.5rem 1rem',
```
robinandeer, line 5: We have spacing tokens in the theme now :)

packages/ui/src/components/CartCard/CartCard.tsx
```diff
@@ -0,0 +1,59 @@
...
+  display: 'block',
+  width: '100%',
+  alignItems: 'center',
+  fontSize: '1rem',
+})
+
+const TitleElement = styled.div({})
+const IconElement = styled.div(
+  {
+    backgroundColor: colorsV3.gray300,
```
robinandeer, line 22: Get these directly from the theme instead like you do below in the border style!

packages/ui/src/components/CartCard/CartCard.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+const RemoveButton = styled.button({
+  marginTop: '2rem',
+  textDecoration: 'underline',
+  fontSize: '0.925rem',
+})
+
+export type CartCardProps = {
+  title?: string
+  price?: string
```
robinandeer, line 43: Hehe how are they both optional? What's the point of rendering it without any of that information? 😅

robinandeer (commented):
Great job! Although this seems more like a store component rather than something for the UI kit, don't you think?

I would put it under the store app for now.

packages/ui/src/components/CartCard/CartCard.text.tsx
```diff
@@ -0,0 +1,15 @@
+import { screen } from '@testing-library/react'
```
guilhermespopolin, line 1: I believe that file should be called `CartCard.test.tsx`, right?! 😄

guilhermespopolin (approved):
Nice :)
Just added a comment about the test file and I agree with Robin about `CartCard` into `store`.

packages/ui/src/components/CartCard/CartCard.tsx
dendoesit: oops, wanted only the price to be optional 😄

packages/ui/src/components/CartCard/CartCard.tsx
dendoesit: is there a way we can do the same ?

packages/ui/src/components/CartCard/CartCard.text.tsx
dendoesit: 😂

PR #327, opened 2022-07-28
 cart page

dendoesit:
## Describe your changes

Added cart page content.


apps/store/src/components/CartList/CartList.tsx
```diff
@@ -1,82 +1,33 @@
...
-import Link from 'next/link'
-import { useContext } from 'react'
-import { Button, Space, CrossIcon } from 'ui'
-import { Text } from '@/components/Text/Text'
-import { PageLink } from '@/lib/PageLink'
-import { CartContext } from '@/services/mockCartService'
-import { ProductNames } from '@/services/mockProductService'
+import { Space } from 'ui'
+import { CartCard } from '../CartCard/CartCard'
+import { ProductData } from '../CheckoutPaymentPage/CheckoutPaymentPage.types'
```
robinandeer, line 4: nit: avoid cross dependencies between components

better to create a duplicate type OR move this to global scope.

apps/store/src/components/CartList/CartList.tsx
```diff
@@ -1,82 +1,33 @@
 import styled from '@emotion/styled'
-import Link from 'next/link'
-import { useContext } from 'react'
-import { Button, Space, CrossIcon } from 'ui'
-import { Text } from '@/components/Text/Text'
-import { PageLink } from '@/lib/PageLink'
-import { CartContext } from '@/services/mockCartService'
-import { ProductNames } from '@/services/mockProductService'
+import { Space } from 'ui'
+import { CartCard } from '../CartCard/CartCard'
```
robinandeer, line 3: nit: use `@/components/CartCard/CartCard` import syntax

apps/store/src/components/CartList/CartList.tsx
```diff
@@ -1,82 +1,33 @@
...
-  const cartContext = useContext(CartContext)
-
-  if (!cartContext) {
-    throw new Error('ProductPage cannot be rendered outside CartContext')
-  }
-
-  const { cart, removeItem } = cartContext
-
+export const CartList = ( {products} : Array<ProductData>) => {
+  console.log(products)
```
robinandeer, line 18: nit: remove

apps/store/src/components/CartPage/CartPage.tsx
```diff
@@ -4,6 +4,7 @@ import { Heading, Space } from 'ui'
 import { CartList } from '@/components/CartList/CartList'
 import { PageLink } from '@/lib/PageLink'
 import { CartContext } from '@/services/mockCartService'
+import { PriceBreakdown } from '../CheckoutPaymentPage/PriceBreakdown'
```
robinandeer, line 7: nit: same, we should aim to either duplicate / make global components over importing between components

That way we can easily remove a component without worrying about breaking other things :)

robinandeer (commented):
Good job!

I would consider grouping all these components under e.g. `components/CartPage` since I assume they are not meant to be reused on other pages, right?

apps/store/src/components/CartPage/CartPage.tsx
dendoesit: hmm, I feel like duplicating would go against the DRY principle. 
The PriceBreakdown component is present on the Cart + Checkout pages so I feel it should be a global component.

apps/store/src/components/CartList/CartList.tsx
dendoesit: cross dependencies refers to this import 
`{ ProductData } from '../CheckoutPaymentPage/CheckoutPaymentPage.types'
`

Checkout and Cart should have the same types for products correct ? @robinandeer

robinandeer: Possibly, but then we should move them to some global scope rather than creating inter-dependencies :)

apps/store/src/components/CartPage/CartPage.tsx
```diff
@@ -14,26 +15,28 @@ export const CartPage = () => {
 
   const { cart } = cartContext
```
guilhermespopolin, line 16: I guess `cart` is not being used anymore. Maybe we can remove it.

apps/store/src/components/CartPage/CartPageProps.types.ts
```diff
@@ -0,0 +1,5 @@
+export type ProductData = { name: string; cost: number; currency: string }
+
+export type CartPageProps = {
```
guilhermespopolin, line 3: Nit:
I'd put that Inside `CartPageProps`

guilhermespopolin (approved):

PR #342, opened 2022-08-01
fix(store: CartNotificationHeader): remove missing ref warning

guilhermespopolin:
## Describe your changes

Even though not describe in the documentation, it seems Dialog title components should be able to accept a ref. Since our custom component used [here](https://github.com/HedvigInsurance/racoon/blob/main/apps/store/src/components/CartNotification/CartToast.tsx#L45) a warning is raised.

Screenshot 2022-08-01 at 13 47 49

### DIFF
--- a/apps/store/src/components/CartNotification/Header.tsx
+++ b/apps/store/src/components/CartNotification/Header.tsx
@@ -1,16 +1,20 @@
 import styled from '@emotion/styled'
+import React from 'react'
 
 type HeaderProps = {
   children: string
 }
 
-export const Header = ({ children }: HeaderProps) => {
-  return (
-    <Wrapper>
-      <StyledTitle>{children}</StyledTitle>
-    </Wrapper>
-  )
-}
+export const Header = React.forwardRef<HTMLDivElement, HeaderProps>(
+  ({ children }, forwardedRef) => {
+    return (
+      <Wrapper ref={forwardedRef}>
+        <StyledTitle>{children}</StyledTitle>
+      </Wrapper>
+    )
+  },
+)
+Header.displayName = 'Header'
 
 const ROW_HEIGHT = '4.5rem'
 


robinandeer (commented):
Hmm, I don't quite understand where this comes from tho.. we never pass ref to this component, no?

guilhermespopolin:
> Hmm, I don't quite understand where this comes from tho.. we never pass ref to this component, no?

I'm sorry, I've put the wrong link in the description. As one can see [here](https://github.com/HedvigInsurance/racoon/blob/main/apps/store/src/components/CartNotification/CartToast.tsx#L45) we're passing `<CartNotification.Header>` as `Title` prop. `<CartNotification.Content>` then [renders](https://github.com/HedvigInsurance/racoon/blob/main/apps/store/src/components/CartNotification/CartNotification.tsx#L51) it as radix `<DialogPrimitive.Title asChild>{Title}</DialogPrimitive.Title asChild>`. The issue is that radix `<DialogPrimitive.Title>` will pass in a ref to its child component but in that case it will be `<CartNotification.Header>` which doesn't accept one, so a warning will be raised.

I guess radix uses that ref in order to focus Dialog's title when it get's opened due accessibility reasons.

robinandeer (approved):
Aha I see, I wonder if it has anything to do with this "asChild" prop that I added?

https://github.com/HedvigInsurance/racoon/blob/main/apps/store/src/components/CartNotification/CartNotification.tsx#L51

If you want you can try to remove that and try again. Otherwise I approve this change!

guilhermespopolin:
> Aha I see, I wonder if it has anything to do with this "asChild" prop that I added?
> 
> https://github.com/HedvigInsurance/racoon/blob/main/apps/store/src/components/CartNotification/CartNotification.tsx#L51
> 
> If you want you can try to remove that and try again. Otherwise I approve this change!

I've tested and removing `asChild` also solves the warning issue. The only tradeoff tho is the generated markup:

**With `asChild`**

Screenshot 2022-08-01 at 16 07 55

**Without `asChild`** 

Screenshot 2022-08-01 at 16 06 45

As one can see, by just removing `asChild` prop the warning goes away but our custom component gets rendered inside a  `<h2>`. I assume that would bot be ideal for accessibility reasons but I've not concrete arguments against it 😅 
I mean, it just doesn't feels right I guess.

PR #345, opened 2022-08-02
feat(GRW-1306): added base product snap component

guilhermespopolin:
## Describe your changes

Adds a base component for implementing _Top Picks_ insurance list for Promise Listing Page. I've decided to call it Snap because the way I see it's not exactly a carousel.

Screenshot 2022-08-02 at 13 44 47

In order to test it, please checkout the branch and verify it with storybook

image

## Jira issue(s): [GRW-1306](https://hedvig.atlassian.net/browse/GRW-1306)

### DIFF
--- a/apps/store/src/components/ProductCardSnap/ProductCardSnap.stories.tsx
+++ b/apps/store/src/components/ProductCardSnap/ProductCardSnap.stories.tsx
@@ -1,0 +1,49 @@
+import { INITIAL_VIEWPORTS } from '@storybook/addon-viewport'
+import { ComponentMeta, ComponentStory } from '@storybook/react'
+import { ProductCard } from '@/components/ProductCard/ProductCard'
+import { ProductCardSnap } from './ProductCardSnap'
+
+console.log(INITIAL_VIEWPORTS)
+
+export default {
+  title: 'ProductCardSnap',
+  component: ProductCardSnap,
+  parameters: {
+    viewport: {
+      viewports: INITIAL_VIEWPORTS,
+      defaultViewport: 'iphonese2',
+    },
+  },
+} as ComponentMeta<typeof ProductCardSnap>
+
+const Template: ComponentStory<typeof ProductCardSnap> = (props) => <ProductCardSnap {...props} />
+
+export const Default = Template.bind({})
+Default.args = {
+  children: [
+    <ProductCard
+      key="home"
+      title="Hedvig Home"
+      subtitle="Complete coverage for your home"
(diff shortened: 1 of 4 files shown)


apps/store/src/components/ProductCardSnap/ProductCardSnap.stories.tsx
```diff
@@ -0,0 +1,49 @@
+import { INITIAL_VIEWPORTS } from '@storybook/addon-viewport'
+import { ComponentMeta, ComponentStory } from '@storybook/react'
+import { ProductCard } from '@/components/ProductCard/ProductCard'
+import { ProductCardSnap } from './ProductCardSnap'
+
+console.log(INITIAL_VIEWPORTS)
```
robinandeer, line 6: delete ;)

[deletstar.jpg)](https://app.graphite.dev/macros/HedvigInsurance)

apps/store/src/components/ProductCardSnap/ProductCardSnap.tsx
```diff
@@ -0,0 +1,27 @@
+import styled from '@emotion/styled'
+import React from 'react'
```
robinandeer, line 2: nit: import not needed with React 17

robinandeer (commented):
Nice work! Haven't tested locally but I think it looks good

My only question would be if `ProductCardSnap` needs to care about what goes inside of it?

Could we just call it e.g. `Slideshow`?

apps/store/src/components/ProductCardSnap/ProductCardSnap.stories.tsx
guilhermespopolin: 🤦🏻

apps/store/src/components/ProductCardSnap/ProductCardSnap.tsx
guilhermespopolin: I see. But if I remove it then it complains about the line where I call `React.Children`
`'React' refers to a UMD global, but the current file is a module. Consider adding an import instead.`

robinandeer (approved):
Looks great to me! The only "nit" left is to fix unify how to spell: "SlideShow" or "Slideshow" 😅 (folder vs. module name)

guilhermespopolin:
> Looks great to me! The only "nit" left is to fix unify how to spell: "SlideShow" or "Slideshow" 😅 (folder vs. module name)

Gonna fix it

PR #346, opened 2022-08-02
refact(store): updated ProductCardSnap

guilhermespopolin:
## Describe your changes

Update `ProductCardSnap` so it uses `ProductCardSnap/ProductCard` component

image

## Jira issue(s): [GRW-1305](https://hedvig.atlassian.net/browse/GRW-1305)

This was a very good use case for graphite. But for some weird reason I can't install the cli 🤷🏻


apps/store/src/components/ProductCardSnap/ProductCard.tsx
```diff
@@ -0,0 +1,57 @@
+import styled from '@emotion/styled'
```
guilhermespopolin, line 1: I guess it wasn't worth to create a new one. The functionality needed is almost exactly the one implemented in `components/ProductCard/ProductCard`. The only difference is the values of `PRODUCT_CARD_IMAGE_WIDTH_SMALL` and `PRODUCT_CARD_IMAGE_HEIGHT_SMALL` env vars.

robinandeer: I think it's fine to duplicate things I this point, then we can look at extracting reusable parts

apps/store/src/components/SlideShow/Slideshow.tsx
```diff
@@ -1,15 +1,24 @@
...
 }
 
-export const Slideshow = ({ children }: SlideshowProps) => {
+export const Slideshow = ({ products }: SlideshowProps) => {
   return (
     <ScollableContainer>
-      {React.Children.map(children, (child, index) => (
-        <ScrollableItem key={index}>{child}</ScrollableItem>
+      {products.map(({ title, subtitle, image }, index) => (
+        <ScrollableItem key={`${title}-${index}`}>
```
robinandeer, line 19: nit: why not just "title"?

guilhermespopolin:
For some weird reason slide show component folder name differs on this PR and my local project 😠 
In this PR one can see it's called `SlideShow` while locally it's called `Slideshow` which should be the correct name.

image

apps/store/src/components/SlideShow/Slideshow.stories.tsx
```diff
@@ -18,30 +17,26 @@ const Template: ComponentStory<typeof Slideshow> = (props) => <Slideshow {...pro
...
+    {
+      title: 'Hedvig Car',
+      subtitle: 'Complete coverage for your car',
+      image: { src: 'https://via.placeholder.com/327x400' },
+    },
+    {
+      title: 'Hedvig Accident',
+      subtitle: 'Complete coverage for when you need most',
+      image: { src: 'https://via.placeholder.com/327x400' },
+    },
```
robinandeer, line 40: Just curious, why do you prefer this API compared to the previous one? :)

I think the previous one with the components is more flexible and easier to understand. It also allows the "Slideshow" to handle any kind of content.

robinandeer (commented):
> For some weird reason slide show component folder name differs on this PR and my local project

You have to call some special git commands to make changes to files and folders when only changing case: https://stackoverflow.com/questions/17683458/how-do-i-commit-case-sensitive-only-filename-changes-in-git

I think it has to do with macOS FS being case insensitive.

apps/store/src/components/SlideShow/Slideshow.stories.tsx
guilhermespopolin: You're right. The issue is that my head was contaminated with the idea of how this component would be used. I mean, I was kind of using the same _API_ one would use in CMS to change what would be displayed. But at the end that responsibility should falls under the Block component, right?!

Will change it back with the previous API 👍🏻

guilhermespopolin: However, by making `Slideshow` handle any kind of content I guess it wouldn't make sense to keep a `ProductCard` inside `Slideshow` folder anymore, right?!

apps/store/src/components/Slideshow/Slideshow.stories.tsx
```diff
@@ -0,0 +1,27 @@
...
+  component: Slideshow,
+  parameters: {
+    viewport: {
+      viewports: INITIAL_VIEWPORTS,
+      defaultViewport: 'iphonese2',
+    },
+  },
+} as ComponentMeta<typeof Slideshow>
+
+export const Default: ComponentStory<typeof Slideshow> = () => (
```
robinandeer, line 20: Nice, I will start using this format -- then we don't need to export types!

guilhermespopolin:
> Nice work, I like this API much better -- let's keep iterating on a name for the component that makes sense for us and the designers but we can do that later :)

Yes. Naming it is hard 😢

apps/store/src/components/TopPickCard/TopPickCard.tsx
```diff
@@ -0,0 +1,57 @@
...
+  width: PRODUCT_CARD_IMAGE_WIDTH_SMALL,
+})
+
+const ImageWrapper = styled.div(() => ({
+  position: 'relative',
+  height: PRODUCT_CARD_IMAGE_HEIGHT_SMALL,
+  width: '100%',
+}))
+
+const Title = styled.p(({ theme }) => ({
```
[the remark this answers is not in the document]
guilhermespopolin: Yes. Good catch

PR #349, opened 2022-08-02
 adapt cart page to design

dendoesit:
## Describe your changes
Bring cart page more to design.


apps/store/package.json
```diff
@@ -7,7 +7,7 @@
     "build": "next build",
     "start": "node --require ./server-preload.js ./node_modules/.bin/next start -p $PORT",
     "test": "jest --passWithNoTests --coverage",
-    "lint": "TIMING=1 next lint",
+    "lint": " next lint",
```
dendoesit, line 10: @robinandeer I saw that you added the TIMING=1 in a previous PR. Is it mandatory ? It currently breaks the yarn dev script on windows.

robinandeer: Sorry @dendoesit, I keep forgetting about that... It enables caching of the linting step so would be nice to keep but we should make it work cross-platform

Feel free to remove it in the mean time. Maybe we could try this: https://stackoverflow.com/questions/25112510/how-to-set-environment-variables-from-within-package-json/36373133#36373133

apps/store/src/components/CartPage/CartPage.tsx
```diff
@@ -12,16 +13,36 @@ export const CartPage = () => {
...
+        <Button fullWidth={true}>
+          <Link href={PageLink.cartReview()}>Check Out</Link>
+        </Button>
+      </Footer>
+    </CartLayout>
   )
 }
+
+const CartLayout = styled.div(({ theme }) => ({
+  marginTop: theme.space[8],
```
robinandeer, line 33: I suggest we avoid margins inside components -- better to move that concern to the parent scope and use something like `Space / SpaceFlex`

apps/store/src/components/LayoutWithMenu/LayoutWithMenu.tsx
```diff
@@ -37,4 +38,5 @@ const FixedHeader = styled.header({
   top: 0,
   left: 0,
   right: 0,
+  backgroundColor: theme.colors.white,
```
robinandeer, line 41: I think the idea is that content bleeds into the header -- like it just floats on top of everything else from looking at the design.

apps/store/src/components/CartCard/CartCard.tsx
```diff
@@ -48,20 +49,20 @@ export type CartCardProps = {
...
         <IconElement />
         <HeaderElement>
           {title && <TitleElement>{title}</TitleElement>}
-          <RemoveButton>Remove</RemoveButton>
+          <RemoveButton onClick={() => handleRemoveProduct()}>Remove</RemoveButton>
         </HeaderElement>
-        {price &&
-          <ExtraElement>{currencyFormatter.format(price)}/mo.</ExtraElement>
-}
+        {price && <ExtraElement>{currencyFormatter.format(price)}/mo.</ExtraElement>}
```
robinandeer, line 65: Is there a reason why we would display the `CartCard` without a price?

robinandeer (commented):
Sorry @dendoesit, I keep forgetting about that... It enables caching of the linting step so would be nice to keep but we should make it work cross-platform

Feel free to remove it in the mean time. Maybe we could try this: https://stackoverflow.com/questions/25112510/how-to-set-environment-variables-from-within-package-json/36373133#36373133

apps/store/src/components/CartPage/CartPage.tsx
dendoesit: I was thinking that the cartLayout is the parent component of the CartPage. 
I really could not figure out how Space works. It seems to only add margin-top if there is already a previous element before.

apps/store/package.json
dendoesit: yeah that seems to do it. I added it to the PR.

apps/store/package.json
```diff
@@ -7,7 +7,7 @@
     "build": "next build",
     "start": "node --require ./server-preload.js ./node_modules/.bin/next start -p $PORT",
     "test": "jest --passWithNoTests --coverage",
-    "lint": "TIMING=1 next lint",
+    "lint": "export TIMING=1 || set TIMING=1 &&  next lint",
```
robinandeer, line 10: that's awesome if this works

```suggestion
    "lint": "export TIMING=1 || set TIMING=1 && next lint",
```

guilhermespopolin (approved):

apps/store/src/components/CartCard/CartCard.tsx
```diff
@@ -39,26 +39,30 @@ const RemoveButton = styled.button({
...
       <Content>
         <IconElement />
         <HeaderElement>
-          {title && <TitleElement>{title}</TitleElement>}
-          <RemoveButton>Remove</RemoveButton>
+          <TitleElement>{title}</TitleElement>
+          <RemoveButton onClick={() => handleRemoveProduct()}>Remove</RemoveButton>
         </HeaderElement>
-        {price && <ExtraElement>{currencyFormatter.format(price)}/mo.</ExtraElement>}
+        <ExtraElement>{currencyFormatter.format(price)}/mo.</ExtraElement>
```
robinandeer, line 65: nit: could be easier to just pass price in as a string here (including the `/mo.` part

robinandeer (approved):

PR #356, opened 2022-08-03
Remove brand repo dependency 💅

robinandeer:
## Describe your changes

Remove brand from ui + onboarding workspaces.
Refactor font preloading.
Replace use of colorsV3 with getColor utility from ui package.
Add UIColor type.
Refactor @font-face and move to global styles definition.
Refactor getColor utility to not lose type safety.
Move font definitions from brand repo to ui package.

The brand repo is deprecated so we should move stuff to the UI package instead.

I wasn't super careful with the market-web stuff but that should be removed in the future so I think that okey (sorry @gustaveen 🙏)

### DIFF
--- a/packages/ui/src/lib/theme/theme.test.ts
+++ b/packages/ui/src/lib/theme/theme.test.ts
@@ -1,18 +1,14 @@
-import { colorsV3 } from '@hedviginsurance/brand'
+import { colors } from './colors'
 import { getColor } from './theme'
 
 describe('theme', () => {
   describe('getColor', () => {
-    it('should return a color when found in colorsV3', () => {
-      expect(getColor('gray100')).toEqual(colorsV3.gray100)
+    it('should return a UI Color when found', () => {
+      expect(getColor('gray100')).toEqual(colors.gray100)
     })
 
     it('should return a simplified color when found in simplifiedColors', () => {
-      expect(getColor('lavender')).toEqual(colorsV3.purple500)
-    })
-
-    it('should default to currentColor when no other colors can be found', () => {
-      expect(getColor('smurf')).toEqual('currentColor')
+      expect(getColor('lavender')).toEqual(colors.purple500)
     })
   })
 })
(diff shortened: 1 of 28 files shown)


robinandeer:
* main
  * **PR #356** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/356" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/356?utm_source=stack-comment).

packages/ui/src/lib/theme/theme.ts
```diff
@@ -30,6 +30,6 @@ declare module '@emotion/react' {
   export interface Theme extends CustomTheme {}
 }
 
-export const getColor = (color?: string) => {
-  return theme.colors[color as keyof typeof theme.colors] || 'currentColor'
+export const getColor = (color?: UIColor) => {
```
alebedev, line 33: Is passing in undefined a valid usage?  We can make code easier to debug if that would be TS error

packages/ui/src/lib/theme/theme.test.ts
```diff
@@ -1,17 +1,18 @@
...
+      expect(getColor('gray100')).toEqual(colors.gray100)
     })
 
     it('should return a simplified color when found in simplifiedColors', () => {
-      expect(getColor('lavender')).toEqual(colorsV3.purple500)
+      expect(getColor('lavender')).toEqual(colors.purple500)
     })
 
     it('should default to currentColor when no other colors can be found', () => {
+      // @ts-ignore - this is a test
```
alebedev, line 15: @ts-expect-error would be better - unlike @ts-ignore, you'll get notified if line below no longer generates error.  This ensures we only keep valid error ignore comments

alebedev (approved):
Looks great!

packages/ui/src/lib/theme/theme.test.ts
robinandeer: [cool](https://app.graphite.dev/macros/HedvigInsurance)

robinandeer (commented):
@alebedev I removed the fallback to `currentColor` and the test for it.

I think it's not really expected behaviour and like you said - this makes for a clear API

PR #358, opened 2022-08-04
feat(GRW-1318): enable storyblok for store page (promise listing page)

guilhermespopolin:
## Describe your changes

* Connect _store_ page with storyblok
* Minor style updates for `<Slideshow>`
* Removes module used for mocking cms


apps/store/src/components/StorePage/StorePage.tsx
```diff
@@ -1,30 +1,13 @@
...
-            </Space>
-          </ul>
-        )}
-        <div>
-          <Link href={PageLink.cart()}>Go to cart</Link>
-        </div>
-      </Space>
+      <Head>
+        <title>Store</title>
+      </Head>
```
robinandeer, line 9: I suggest we move this to under /pages since it's next.js specific

Would be nice if code in "components" would be as portable as possible, don't you agree?

apps/store/src/pages/store.tsx
```diff
@@ -1,22 +1,22 @@
...
 
-export const getServerSideProps: GetServerSideProps<StorePageProps> = async (context) => {
-  const localeData = getLocale(context.locale ?? context.defaultLocale)
+  return <StorePage {...story} />
+}
 
-  const products = await CmsService.getProductsByMarket(localeData.marketLabel)
+export const getServerSideProps: GetServerSideProps<StoryblokPageProps> = async (context) => {
+  const story = await getStoreStory(context.preview)
+  console.log(story)
```
robinandeer, line 15: remove

robinandeer (approved):
Looks good overall to me

This could just be a content page rendered by `[...slug].tsx` but since it's so central I don't mind defining it explicitly here.

[goodidea](https://app.graphite.dev/macros/HedvigInsurance)

apps/store/src/components/StorePage/StorePage.tsx
guilhermespopolin: Yes, ofc. My bad.

guilhermespopolin:
> Looks good overall to me
> 
> This could just be a content page rendered by `[...slug].tsx` but since it's so central I don't mind defining it explicitly here.
> 
> [goodidea](https://app.graphite.dev/macros/HedvigInsurance) [ goodidea ](https://app.graphite.dev/macros/HedvigInsurance) [ ](https://app.graphite.dev/macros/HedvigInsurance)

Oh, that's right. I haven't realised that. Let me check if it would works fine by removing `pages/store`

PR #362, opened 2022-08-04
Feature/remove product

dendoesit:
Added modal to remove products from cart.

This is blocked until the cart design updates pr is merged.


apps/store/package.json
```diff
@@ -7,7 +7,7 @@
     "build": "next build",
     "start": "node --require ./server-preload.js ./node_modules/.bin/next start -p $PORT",
     "test": "jest --passWithNoTests --coverage",
-    "lint": "TIMING=1 next lint",
```
alebedev, line 10: There's cross-env nom package for that - ```cross-env TIMING=1 next lint```

Do you want to try it?

apps/store/src/components/CartPage/CartPage.tsx
```diff
@@ -12,16 +13,34 @@ export const CartPage = () => {
...
+        </Heading>
+      </PageHeader>
       <CartList products={products} />
-      <PriceBreakdown currency="SEK" products={products} cost={cost} />
-      <Button fullWidth={true}>
-        <Link href={PageLink.cartReview()}>Check Out</Link>
-      </Button>
-    </>
+      <Footer>
+        <PriceBreakdown currency="SEK" products={products} cost={cost} />
```
alebedev, line 25: Can we move default currency to different level?  Probably another task, just caught my eye now.  I doubt we want to have currency default at presentation level, does not seem scaleable

apps/store/src/components/CartCard/CartCard.tsx
```diff
@@ -39,26 +44,53 @@ const RemoveButton = styled.button({
...
         <HeaderElement>
-          {title && <TitleElement>{title}</TitleElement>}
-          <RemoveButton>Remove</RemoveButton>
+          <TitleElement>{title}</TitleElement>
+
+          <CartNotification.Root>
+            <CartNotification.Trigger asChild>
+              <RemoveButton>Remove</RemoveButton>
+            </CartNotification.Trigger>
+            <CartNotification.Content Title={<Header>Remove insurance ?</Header>}>
```
alebedev, line 74: What's our policy on localisation here?  When it's OK to hardcode English text?

alebedev (commented):
Left some questions, none of them critical for the review

guilhermespopolin (commented):
I'd suggest display the buttons in the center of the modal dialog

image

apps/store/src/components/CartCard/CartCard.tsx
dendoesit: I believe we will get to the point where we will pass lokalize text keys but for now English text should be fine for presentation.

apps/store/package.json
dendoesit: I have managed to make it work using this:     "lint": "export TIMING=1 || set TIMING=1 &&  next lint",
but will give the cross-env a try.

apps/store/src/components/CartPage/CartPage.tsx
dendoesit: yeah, will definetely be part of a different task. should probably take currency out of url maybe.

dendoesit:
great catch @guilhermespopolin . updated.

guilhermespopolin (approved):

alebedev (approved):

PR #365, opened 2022-08-05
Add Tabs + Tab blocks to CMS setup

robinandeer:
## Describe your changes

Add Tabs + Tab blocks to CMS setup

Add types module for shared types between blocks.

Used on the product page.

I considered adding it to the product page content type but seems a bit too early - this is more flexible.

[GRW-1320]: https://hedvig.atlassian.net/browse/GRW-1320?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/store/src/blocks/TabsBlock.tsx
+++ b/apps/store/src/blocks/TabsBlock.tsx
@@ -1,0 +1,51 @@
+import { SbBlokData, StoryblokComponent, storyblokEditable } from '@storyblok/react'
+import { useMemo } from 'react'
+import * as Tabs from '@/components/ProductPage/Tabs'
+import { SbBaseBlockProps, StoryblokBlockName } from '@/services/storyblok/storyblok'
+
+type TabsBlockProps = SbBaseBlockProps<{
+  items: Array<SbBlokData>
+}>
+
+type TabBlockFields = {
+  title: string
+  body: Array<SbBlokData>
+}
+
+type TabBlockData = SbBlokData & TabBlockFields
+
+type TabBlockProps = SbBaseBlockProps<TabBlockFields>
+
+export const TabsBlock = ({ blok }: TabsBlockProps) => {
+  const tabBlocks = useMemo(() => blok.items.filter(isTabBlock), [blok.items])
+  const firstTabValue = blok.items[0]?._uid
+
+  return (
+    <Tabs.Tabs defaultValue={firstTabValue} {...storyblokEditable(blok)}>
+      <Tabs.TabsList>
+        {tabBlocks.map((tabBlock) => (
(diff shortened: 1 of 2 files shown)


robinandeer:
* main
  * **PR #365** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/365" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/365?utm_source=stack-comment).

apps/store/src/blocks/TabsBlock.tsx
```diff
@@ -0,0 +1,25 @@
...
+      <Tabs.TabsList>
+        {blok.items.map((item) => (
+          <Tabs.TabsTrigger key={item._uid} value={item._uid}>
+            {item.title}
+          </Tabs.TabsTrigger>
+        ))}
+      </Tabs.TabsList>
+      {blok.items.map((nestedBlock) => (
+        <StoryblokComponent key={nestedBlock._uid} blok={nestedBlock} />
+      ))}
```
guilhermespopolin, line 22: [Discussion]
Maybe it's just the way _Storyblok_ works but I get the sense that this solution might be too implicit and therefore hard to understand. The issue here is that, even tho we can define in _Storyblok_ that only `TabBlock` should be accepted as children of `TabsBlock`, that relationship is not well expressed in the code. E.g: here we're rendering `<StoryblokComponent />` which can be any block, but at the end it will became a `TabBlock`.

My first thought would be changed it for `<TabBlock>` and don't register a component for it inside `storybook` service.

```suggestion
      {blok.items.map((nestedBlock) => (
        <TabBlok key={nestedBlock._uid} blok={nestedBlock} />
      ))}
```
But to be honest I don't thing I've a solution for it. I've just thought it would be a good thing to discuss and I was hoping to see other opinions about it :)

guilhermespopolin (approved):

apps/store/src/blocks/TabsBlock.tsx
robinandeer: @guilhermespopolin very good point, I'm also not very happy with this setup.

@alebedev and I had similar discussion around the accordion. But here it becomes even more obvious.

I'm leaning towards your idea if just using the `TabBlock` inside the `TabsBlock` instead of using `StoryblokComponent` - you made a good case!

alebedev (approved):

apps/store/src/blocks/TabsBlock.tsx
guilhermespopolin: Thanks :)
I'm glad I could help

PR #366, opened 2022-08-08
Feature/GRW-1330 yarn3 poc

alebedev:
## Yarn v3 migration POC

Update racoon to yarn v3 to see how it works for us

Research which of the promised benefits apply to us
- Faster installs
- Smaller node_modules
- Future-proof

[GRW-1330]: https://hedvig.atlassian.net/browse/GRW-1330?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


package.json
```diff
@@ -33,6 +33,7 @@
     ]
   },
   "devDependencies": {
+    "eslint": "8.21.0",
```
robinandeer, line 36: hmm do we need this globally?

.yarnrc.yml
```diff
@@ -0,0 +1,13 @@
+# Configuration docs: https://yarnpkg.com/configuration/yarnrc
+
+# Use exact versions of new dependencies by default
+defaultSemverRangePrefix: ''
+
+# TODO: Experiment with pnpm or zero-install linkers
+nodeLinker: node-modules
+
+plugins:
+  - path: .yarn/plugins/@yarnpkg/plugin-interactive-tools.cjs
```
robinandeer, line 10: What does this do? Do we need it or should we start without it?

robinandeer (commented):
Looks good to me, if this is all the changes we need to do, why not?

I have heard really good things about pnpm so maybe we should look into how that would work as well, what do you say?

apps/store/package.json
```diff
@@ -7,7 +7,7 @@
     "build": "next build",
     "start": "node --require ./server-preload.js ./node_modules/.bin/next start -p $PORT",
     "test": "jest --passWithNoTests --coverage",
-    "lint": "export TIMING=1 || set TIMING=1 && next lint",
+    "lint": "TIMING=1 next lint",
```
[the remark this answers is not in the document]
alebedev: It does not with yarn v3 - they provide shell wrapper to handle env variables on windows.  But since we don't need TIMING at all, I can remove it

package.json
```diff
@@ -22,7 +22,7 @@
     "storybook:build": "turbo run storybook:build",
     "chromatic": "turbo run chromatic"
   },
-  "packageManager": "yarn@1.22.17",
+  "packageManager": "yarn@3.2.2",
```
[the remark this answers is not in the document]
alebedev: Good catch!  It can be safely removed from yarnrc and .yarn/releases can be also removed since we're using corepack

.yarnrc.yml
alebedev: This supports ```yarn upgrade-interactive``` -- they moved it from core (v1) to plugin (v3).  Works better for monorepos too, compared to v1

I think we need it for developer experience - faster and safer upgrades are good

package.json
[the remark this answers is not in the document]
alebedev: ... and this breaks our CI.  Rolled back, added TODO to check later

package.json
alebedev: We need it in every package that references eslint in its scripts.  Did not find how to remove duplicate dependencies yet.  At least they are now always in sync when using upgrade-interactive

.yarnrc.yml
robinandeer: Sounds good! Maybe add comment about that directly in this file?

alebedev: done

robinandeer (approved):
Looks good to me, haven't tested it locally yet tho :)

dendoesit:
tried to test it locally, but did we want to change the url ? seems to start http://localhost:3001/se-en/new-member now and before it was on 8040.

robinandeer:
I also get the correct port :)

Looks really good :)

PR #367, opened 2022-08-08
feat(GRW-1323): added base perils component

guilhermespopolin:

### DIFF
--- a/apps/store/src/components/Accordion/Accordion.tsx
+++ b/apps/store/src/components/Accordion/Accordion.tsx
@@ -1,6 +1,6 @@
 import styled from '@emotion/styled'
 import * as AccordionPrimitives from '@radix-ui/react-accordion'
-import { PropsWithChildren } from 'react'
+import { PropsWithChildren, ReactElement } from 'react'
 import { ChevronIcon } from 'ui'
 
 export const Root = AccordionPrimitives.Root
@@ -21,12 +21,15 @@ const Trigger = styled(AccordionPrimitives.Trigger)(() => ({
   justifyContent: 'space-between',
 }))
 
-export const HeaderWithTrigger = ({ children }: PropsWithChildren<unknown>) => {
+export const HeaderWithTrigger = ({
+  children,
+  icon = <TriggerIcon size="1rem" />,
+}: PropsWithChildren<unknown> & { icon?: ReactElement }) => {
   return (
     <Header>
       <Trigger>
         {children}
-        <TriggerIcon size="1rem" />
+        {icon}
       </Trigger>
     </Header>
   )
--- a/apps/store/src/components/Perils/Perils.tsx
+++ b/apps/store/src/components/Perils/Perils.tsx
@@ -1,0 +1,112 @@
+import styled from '@emotion/styled'
+import React, { useState } from 'react'
+import { CrossIcon } from 'ui'
+import * as Accordion from '@/components/Accordion/Accordion'
+import { SpaceFlex } from '@/components/SpaceFlex/SpaceFlex'
+import { CheckIcon } from './CheckIcon'
+import { MinusIcon } from './MinusIcon'
+import { PlusIcon } from './PlusIcon'
+
+type Peril = {
+  icon: React.ReactNode
+  name: string
+  description: string
+  covered: Array<string>
+  notCovered: Array<string>
+}
+
+type PerilsProps = {
+  perils: Array<Peril>
+}
+
+export const Perils = ({ perils }: PerilsProps) => {
+  const [openedItems, setOpenedItems] = useState<Array<string>>()
+
+  return (
+    <Accordion.Root
+      type="multiple"
+      value={openedItems}
+      onValueChange={(value) => {
+        setOpenedItems(value)
+      }}
+    >
+      {perils.map(({ icon, name, description, covered, notCovered }) => {
+        const isOpened = openedItems?.includes(name)
+
+        return (
+          <Accordion.Item key={name} value={name}>
+            <Accordion.HeaderWithTrigger icon={isOpened ? <MinusIcon /> : <PlusIcon />}>
+              <SpaceFlex space={0.5}>
+                {icon}
+                {name}
+              </SpaceFlex>
+            </Accordion.HeaderWithTrigger>
+            <Accordion.Content>
+              <ContentWrapper>
+                <p>{description}</p>
+                {covered.length > 0 && (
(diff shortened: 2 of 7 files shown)


guilhermespopolin:
* main
  * **PR #367** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/367" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/367?utm_source=stack-comment).

apps/store/src/components/Accordion/Accordion.tsx
```diff
@@ -21,12 +21,15 @@ const Trigger = styled(AccordionPrimitives.Trigger)(() => ({
   justifyContent: 'space-between',
 }))
 
-export const HeaderWithTrigger = ({ children }: PropsWithChildren<unknown>) => {
+export const HeaderWithTrigger = ({
+  children,
+  icon = <TriggerIcon size="1rem" />,
+}: PropsWithChildren<unknown> & { icon?: ReactElement }) => {
```
alebedev, line 27: Would separate Props type be more readable?

apps/store/src/components/Perils/Perils.tsx
```diff
@@ -0,0 +1,112 @@
...
+  covered: Array<string>
+  notCovered: Array<string>
+}
+
+type PerilsProps = {
+  perils: Array<Peril>
+}
+
+export const Perils = ({ perils }: PerilsProps) => {
+  const [openedItems, setOpenedItems] = useState<Array<string>>()
```
alebedev, line 23: I know we're talking about low number of items in most cases, but using Set would be more efficient.  ```openedItems.includes``` on every item means quadratic complexity

apps/store/src/components/Perils/Perils.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+  return (
+    <Accordion.Root
+      type="multiple"
+      value={openedItems}
+      onValueChange={(value) => {
```
alebedev, line 29: useCallback here may help with number of re-renders.  Could you profile and see if this is the case?

apps/store/src/components/Perils/Perils.tsx
```diff
@@ -0,0 +1,112 @@
...
+                        <CoverageListItem key={`${itemCovered}-${index}`}>
+                          <CheckIcon color="currentColor" size="10px" />
+                          {itemCovered}
+                        </CoverageListItem>
+                      ))}
+                    </CoverageList>
+                  </CoverageListWrapper>
+                )}
+                {notCovered.length > 0 && (
+                  <CoverageListWrapper>
```
alebedev, line 61: Looks like copy-pasted block.  Can we extract component here?

Also, shouldn't ```covered.map``` be ```notCovered.map```?

alebedev (commented):
Looks good!  Let's discuss data structure and efficiency comments if you have questions

apps/store/src/components/Perils/Perils.tsx
guilhermespopolin: 🤔 I see. I understand that [`Set.has` is faster than `Array.prototype.includes`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set#performance) but I guess on this specific case adding it might just complicate things since we don't have a performance issue yet. But maybe I've missing something. The following is how I understood how a `Set` would help:

```suggestion
  const [openedItems, setOpenedItems] = useState<Set<string>>(new Set())

  const handleValueChange = useCallback((value: Array<string>) => {
    setOpenedItems(new Set(value))
  }, [])

  return (
    <Accordion.Root
      type="multiple"
      value={Array.from(openedItems)}
      onValueChange={handleValueChange}
    >
        {perils.map(({ icon, name, description, covered, notCovered }) => {
            const isOpened = openedItems.has(name)
```

As one can see, I'd need to create a new `Set` so React re-renders the component and then convert that `Set` into a regular array in order to be compliance with `Accordion.Root` API. If I got correctly we'd be moving from O(n^2) --> O(n), which is fair. In any case, I don't mind adding it :)

alebedev: Forgot we need to pass down value as well.  to/from conversion on every change does not seem ergonomic.   

It's easier to keep state an array and use set only for `isOpen`. Something like
```
const openedSet = new Set(openedItems)

perils.map((...) => {
  const isOpened = openedSet.has(name)
})
```

apps/store/src/components/Perils/Perils.tsx
```diff
@@ -0,0 +1,112 @@
...
+import { CrossIcon } from 'ui'
+import * as Accordion from '@/components/Accordion/Accordion'
+import { SpaceFlex } from '@/components/SpaceFlex/SpaceFlex'
+import { CheckIcon } from './CheckIcon'
+import { MinusIcon } from './MinusIcon'
+import { PlusIcon } from './PlusIcon'
+
+type Peril = {
+  icon: React.ReactNode
+  name: string
```
alebedev, line 12: The way we're using Peril, name need to be unique and it's not obvious from component APi.  With real backend, we'll have some unique ID, I assume.  Can we add it now?

guilhermespopolin: I guess It's a fair addition :) Will do

apps/store/src/components/Perils/Perils.tsx
```diff
@@ -29,11 +30,11 @@ export const Perils = ({ perils }: PerilsProps) => {
         setOpenedItems(value)
       }}
     >
-      {perils.map(({ icon, name, description, covered, notCovered }) => {
+      {perils.map(({ id, icon, name, description, covered, notCovered }) => {
         const isOpened = openedItems?.includes(name)
 
         return (
-          <Accordion.Item key={name} value={name}>
+          <Accordion.Item key={id} value={name}>
```
alebedev, line 35: what would happen if there's duplicate name in the list?  Do we need `value={id}` to prevent potential issues?

guilhermespopolin: 🤔 I guess that reasonable to expect that those names would be different. I believe that receiving them with duplicates indicates an issue that lies where they're from.

alebedev: OK, let's go with this assumption

alebedev (approved):

PR #370, opened 2022-08-08
Add ShopSession query

robinandeer:
## Describe your changes

Add ShopSession query.

Update mocks to cover the new query.

Update ShopSessionService to use the new query.

## Justify why they are needed

The new query will be used in places where we expect a ShopSession to already exist.
For example in the checkout.

## Checklist before requesting a review

- [ ] I have performed a self-review of my code
- [ ] Will this be part of a product update? If yes, please write one phrase about this update.

[GRW-1322]: https://hedvig.atlassian.net/browse/GRW-1322?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #371, opened 2022-08-08
Add API endpoint for cart review form

robinandeer:
## Describe your changes

Add API endpoint for cart review form

Add skeleton start date update function in cart service.

Ability to update start dates.

### DIFF
--- a/apps/store/src/components/CartReviewPage/useHandleSubmitCartReview.ts
+++ b/apps/store/src/components/CartReviewPage/useHandleSubmitCartReview.ts
@@ -1,0 +1,34 @@
+import { FormEvent, useState } from 'react'
+import { PageLink } from '@/lib/PageLink'
+
+type Params = {
+  onSuccess: () => void
+}
+
+export const useHandleSubmitCartReview = ({ onSuccess }: Params) => {
+  const [status, setStatus] = useState<'idle' | 'submitting'>('idle')
+
+  const handleSubmit = async (event: FormEvent<HTMLFormElement>) => {
+    event.preventDefault()
+
+    if (status === 'submitting') return
+
+    const formData = new FormData(event.currentTarget)
+    const data = Object.fromEntries(formData.entries())
+
+    setStatus('submitting')
+
+    try {
+      await fetch(PageLink.apiCartReview(), {
+        method: 'PATCH',
+        body: JSON.stringify(data),
+        headers: { 'Content-Type': 'application/json' },
+      })
+      onSuccess()
+    } finally {
+      setStatus('idle')
+    }
+  }
+
+  return [handleSubmit, status] as const
+}
--- a/apps/store/src/components/CartReviewPage/CartReviewPage.constants.ts
+++ b/apps/store/src/components/CartReviewPage/CartReviewPage.constants.ts
@@ -1,0 +1,1 @@
+export const LINE_PREFIX = 'line-'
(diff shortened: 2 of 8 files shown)


robinandeer:
* main
  * **PR #368** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/368" target="_blank">Graphite</a> 
    * **PR #370** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/370" target="_blank">Graphite</a> 
      * **PR #371** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/371" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/371?utm_source=stack-comment).

apps/store/src/components/CartReviewPage/useHandleSubmitCartReview.ts
```diff
@@ -0,0 +1,34 @@
...
+        body: JSON.stringify(data),
+        headers: { 'Content-Type': 'application/json' },
+      })
+      onSuccess()
+    } finally {
+      setStatus('idle')
+    }
+  }
+
+  return [handleSubmit, status] as const
```
KarnellSchultz, line 33: 🧶  Wondering if we're able to get better intellisense from returning an object here.

My thinking is that when a hook returns an array I'd expect the first and second items to be something like `[state, setState]` and with this returning two different things I'd likely end up navigating to the file where this hook lives to find out what to expect in return. 

Perhaps that's just my workflow and I'm overlooking something there.

KarnellSchultz (approved):

apps/store/src/components/CartReviewPage/useHandleSubmitCartReview.ts
alebedev: We can also see if explicitly declared tuple return type gives us good intelligence

https://www.typescriptlang.org/docs/handbook/2/objects.html#tuple-types

robinandeer: Right, I think this essentially turns the return type into a tuple tho

I see your point @KarnellSchultz. What I like about this is that we let the "consumer" of the hook decide what to name the variables.

Otherwise we have to come up with good names that work in any context. But perhaps we can do that since it's only an internal hook.

robinandeer: That's one thing I don't really like with Apollo Client hooks at least

apps/store/src/components/CartReviewPage/useHandleSubmitCartReview.ts
```diff
@@ -0,0 +1,34 @@
...
+import { PageLink } from '@/lib/PageLink'
+
+type Params = {
+  onSuccess: () => void
+}
+
+export const useHandleSubmitCartReview = ({ onSuccess }: Params) => {
+  const [status, setStatus] = useState<'idle' | 'submitting'>('idle')
+
+  const handleSubmit = async (event: FormEvent<HTMLFormElement>) => {
```
alebedev, line 11: Do we need useCallback to keep submit handler stable?

alebedev (approved):
Looks good!

KarnellSchultz:
Let's get it merged ✅

PR #379, opened 2022-08-10
Refactor useShopSession hook

robinandeer:
## Describe your changes

Replace `useCart` with `useShopSession` hook.

Create a `ShopSessionContext` to avoid duplicate requests.

Avoid running ShopSession queries in SSR mode.

Implement "get or create shop session" client side.

Fix bug in shop session query mock.

Remove unused shop session helper.

Add option to get and save shop session ID from service.

Fix bug in shop session get method, return null if not found.

I believe we should aim to use Apollo Client for all client side fetching => or use something else :)

This approach will let us keep an active query on the page that we can invalidate e.g. when an item is added to cart so it updates automatically.

Had to add `eslint-import-resolver-typescript` since husky and VSCode were reporting errors related to `order/import`.


robinandeer:
* main
  * **PR #379** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/379" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/379?utm_source=stack-comment).

apps/store/src/services/shopSession/useShopSession.ts
```diff
@@ -0,0 +1,73 @@
...
+    shopSessionId,
+    onCompleted: ({ shopSession }) => {
+      if (shopSession.countryCode !== countryCode) createShopSession()
+    },
+  })
+
+  return useMemo(() => {
+    const shopSessionCountryCode = result.data?.shopSession?.countryCode
+    return {
+      ...result,
```
alebedev, line 38: I see we may be returning result from query while waiting for mutation.  I'll see what can be improved

apps/store/src/services/shopSession/ShopSessionService.ts
```diff
@@ -15,10 +15,20 @@ export class ShopSessionService {
     private readonly apolloClient: ApolloClient<unknown>,
   ) {}
 
+  public get shopSessionId() {
```
alebedev, line 18: Nit: `getShopSessionId()` ?

This is currently the only getter in racoon.  I feed that getters/setters make code less obvious without much improvement in developer experience

apps/store/src/services/shopSession/ShopSessionService.ts
```diff
@@ -15,10 +15,20 @@ export class ShopSessionService {
...
+    return this.persister.fetch()
+  }
+
+  public save(shopSessionId: string) {
```
alebedev, line 22: Nit: `saveShopSessionId(..)`?  We're not saving whole session

apps/store/src/services/shopSession/useShopSession.ts
alebedev: Checked Apollo API, this part is fine

robinandeer: 👍

apps/store/src/services/shopSession/ShopSessionService.ts
robinandeer: ahh right, yeah we can change that :)

KarnellSchultz (approved):
LGTM 🍒

PR #380, opened 2022-08-10
GRW-1330/fix/update cypress scripts for yarn3

alebedev:

After upgrade yarn is stricter with env variables and never-existing $PORT became an error.  We can work without it

[GRW-1330]: https://hedvig.atlassian.net/browse/GRW-1330?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/onboarding/package.json
+++ b/apps/onboarding/package.json
@@ -5,10 +5,10 @@
   "scripts": {
     "dev": "next dev --port 3001",
     "build": "next build",
-    "start": "node --require ./server-preload.js ./node_modules/.bin/next start -p $PORT",
+    "start": "node --require ./server-preload.js ./node_modules/.bin/next start",
     "test": "jest --passWithNoTests",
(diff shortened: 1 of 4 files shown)


apps/onboarding/package.json
```diff
@@ -5,10 +5,10 @@
   "scripts": {
     "dev": "next dev --port 3001",
     "build": "next build",
-    "start": "node --require ./server-preload.js ./node_modules/.bin/next start -p $PORT",
+    "start": "node --require ./server-preload.js ./node_modules/.bin/next start",
```
robinandeer, line 8: Have you checked that we don't need this for the deploy to work?

robinandeer (approved):
If this doesn't break deploy, look good!

apps/onboarding/package.json
alebedev: Looks good in preview app

PR #383, opened 2022-08-11
GRW-1339 onboarding codegen on-demand

KarnellSchultz:
## Describe your changes

Changes only for `onboarding`

Changed the `dev` command to depend on the `codegen` step so that each time the dev command is run the codegen step happens before it.

Added the generated files to the `.gitignore`.

From a Slack conversation here: [slack](https://hedviginsurance.slack.com/archives/G01MARCFUUW/p1660121749295119)

The changes were made to make the apps more consistent with one another. Since the `store` was the only app that had this setup previously.


turbo.json
```diff
@@ -46,14 +46,14 @@
       "dependsOn": ["^build"]
     },
     "onboarding#test": {
-      "outputs": ["covereage/**"],
+      "outputs": ["coverage/**"],
       "dependsOn": ["^build"]
     },
     "test": {
-      "outputs": ["covereage/**"]
+      "outputs": ["coverage/**"]
```
robinandeer, line 53: nice find 😅

turbo.json
```diff
@@ -98,10 +102,15 @@
       "dependsOn": ["codegen", "^build"],
       "outputs": []
     },
+    "onboarding#codegen": {
+      "inputs": ["src/graphql/*.graphql"],
+      "outputs": ["src/services/**/generated.ts"],
```
robinandeer, line 107: This outputs also to `src/services/apollo/types.ts`

turbo.json
```diff
@@ -46,14 +46,14 @@
       "dependsOn": ["^build"]
```
robinandeer, line 46: <i class='graphite__hidden'>[Re: line 11]<br/></i>

I wonder if we need to depend on the codegen here as well 🤔<span class='graphite__hidden'><br/><br/>See this comment inline on <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/383?utm_source=unchanged-line-comment">Graphite</a>.</span>

turbo.json
KarnellSchultz: good catch

thought about making it a glob like this but wonder if there might be some other files here that unintentially match the glob and cause issues. 

`"outputs": ["src/services/**/{generated,types}.ts"],`

any thoughts on that?

turbo.json
KarnellSchultz: Seems like a good idea. 

Especially since we're going to be swapping this with the real API soon and likely copy/paste over these settings.

turbo.json
```diff
@@ -43,17 +43,17 @@
     },
     "fake-api#build": {
       "outputs": [".next/**"],
-      "dependsOn": ["^build"]
+      "dependsOn": ["^build", "codegen"]
```
robinandeer, line 46: I don't think this needs "codegen" :)

KarnellSchultz: ahhh i misunderstood your previous comment about `line:11`. 

I see now you mean that `onboarding` `build` step _should_  have `codegen` as well. 👍

guilhermespopolin (approved):

PR #386, opened 2022-08-11
GRW-1348 / Feature / Update button block to match design

robinandeer:
## Describe your changes

Center button horizontally.
Add variant option to CMS.

### DIFF
--- a/apps/store/src/blocks/ButtonBlock.tsx
+++ b/apps/store/src/blocks/ButtonBlock.tsx
@@ -1,22 +1,27 @@
+import styled from '@emotion/styled'
 import { storyblokEditable } from '@storyblok/react'
-import { LinkButton } from 'ui'
+import { ButtonVariant, LinkButton } from 'ui'
 import { LinkField, SbBaseBlockProps } from '@/services/storyblok/storyblok'
 
 type ButtonBlockProps = SbBaseBlockProps<{
   text: string
   link: LinkField
+  variant: ButtonVariant
 }>
 
 export const ButtonBlock = ({ blok }: ButtonBlockProps) => {
   return (
-    <LinkButton
-      href={blok.link.cached_url}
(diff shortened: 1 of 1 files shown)


robinandeer:
* main
  * **PR #386** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/386" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/386?utm_source=stack-comment).

apps/store/src/blocks/ButtonBlock.tsx
```diff
@@ -1,22 +1,27 @@
...
-    <LinkButton
-      href={blok.link.cached_url}
-      color="dark"
-      variant="filled"
-      size="lg"
-      {...storyblokEditable(blok)}
-    >
-      {blok.text}
-    </LinkButton>
+    <Wrapper {...storyblokEditable(blok)}>
```
KarnellSchultz, line 14: Are you able to use this `Wrapper` before it's being declared on line 22? 

I'd expect the `Wrapper` component to have to be above the `ButtonBlock

KarnellSchultz (approved):

apps/store/src/blocks/ButtonBlock.tsx
KarnellSchultz: I see now that i'm just used to having this [ESLint rule](https://eslint.org/docs/latest/rules/no-use-before-define) on

`no-use-before-define`

robinandeer: Right, I kind of like being able to put these "helper functions" below the main export since they are not so important.

It works in this case since `Wrapper` is defined and then only later used inside the function when it's being called.

PR #388, opened 2022-08-12
GRW-1342 / Fix support block ssr bug

amandaflod:
## Describe your changes

Added an 'undefined' check to useBreakpoint hook to work around the SSR problems. In the future we should maybe try to handle responsiveness in CSS instead as this fix is not perfect 😅


robinandeer (changes requested):
Looks good, but this can't be based on latest main commit because we already have `eslint-import-resolver-typescript` as a dependency there

Perhaps you can try to rebase and request review again? :)

alebedev (approved):
Looks good

amandaflod:
Thank you @robinandeer ! I was so sure my branch was up to date but you guys are just so fast! 🤠🏎 Have fixed it now!

robinandeer (approved):

PR #391, opened 2022-08-15
Add ImageBlock to CMS setup

robinandeer:

Primitive component.

I'm starting to realize that adding padding to each primitive block makes it harder to nest inside each other. Perhaps we can discuss how to handle various layout concerns @gustaveen?

### DIFF
--- a/apps/store/src/blocks/ImageBlock.tsx
+++ b/apps/store/src/blocks/ImageBlock.tsx
@@ -1,0 +1,47 @@
+import styled from '@emotion/styled'
+import { SbBlokData, StoryblokComponent, storyblokEditable } from '@storyblok/react'
+import Image from 'next/image'
+import { SbBaseBlockProps, StoryblokImage } from '@/services/storyblok/storyblok'
+
+type ImageBlockProps = SbBaseBlockProps<{
+  image: StoryblokImage
+  fullBleed?: boolean
+  body?: Array<SbBlokData>
+}>
+
+export const ImageBlock = ({ blok }: ImageBlockProps) => {
+  const sizeProps = getSizeFromURL(blok.image.filename)
+
+  return (
+    <Wrapper {...storyblokEditable(blok)} margins={blok.fullBleed ? false : true}>
+      <Image src={blok.image.filename} {...sizeProps} alt={blok.image.alt} />
+      <BodyWrapper>
+        {blok.body?.map((nestedBlock) => (
+          <StoryblokComponent key={nestedBlock._uid} blok={nestedBlock} />
+        ))}
+      </BodyWrapper>
+    </Wrapper>
+  )
+}
+
+const Wrapper = styled.div<{ margins: boolean }>(({ theme, margins = true }) => ({
+  paddingLeft: margins ? theme.space[4] : 0,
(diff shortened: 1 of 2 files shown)


robinandeer:
* main
  * **PR #391** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/391" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/391?utm_source=stack-comment).

apps/store/src/services/storyblok/storyblok.ts
```diff
@@ -80,6 +81,7 @@ export enum StoryblokBlockName {
   Content = 'content',
   CheckList = 'checkList',
   Text = 'text',
+  Image = 'image',
```
KarnellSchultz, line 84: 🧶  Nothing to do with your changes

What do you think about putting this enum and the object on `line:88" in alphabetical order?

Thinking as these things grow they're going to become unwieldy.

robinandeer: aha yeah for sure, good idea!

apps/store/src/blocks/ImageBlock.tsx
```diff
@@ -0,0 +1,47 @@
...
+  image: StoryblokImage
+  fullBleed?: boolean
+  body?: Array<SbBlokData>
+}>
+
+export const ImageBlock = ({ blok }: ImageBlockProps) => {
+  const sizeProps = getSizeFromURL(blok.image.filename)
+
+  return (
+    <Wrapper {...storyblokEditable(blok)} margins={blok.fullBleed ? false : true}>
```
KarnellSchultz, line 16: Thinking this might be a bit more easy to read.

```suggestion
    <Wrapper {...storyblokEditable(blok)} margins={!blok.fullBleed>
```

KarnellSchultz (approved):
Looks good 👍 

one thing. I tried this out in Storyblok and was not able to see the Image blok. Maybe I was doing something wrong though.

apps/store/src/blocks/ImageBlock.tsx
```diff
@@ -0,0 +1,47 @@
...
+      <Image src={blok.image.filename} {...sizeProps} alt={blok.image.alt} />
+      <BodyWrapper>
+        {blok.body?.map((nestedBlock) => (
```
gustaveen, line 19: My gut feeling was that the image block would not have nested blocks, is there a need for it or maybe it will be down the line? Currently we have support for a caption field in `market-web`

gustaveen (commented):
Awesome! 🙌 Was just curious if we needed nested blocks on images. 
Regarding the layout I think we could work with some layout block. So for repetitive blocks we have a more generic parent block to set layout and spacing. I  have some ideas I want to run by Mange (he's on holiday still) and Zak but I'll talk to Zak and I can draft out something :)

apps/store/src/blocks/ImageBlock.tsx
gustaveen: Saw the design with text on image so let's go with this and adjust if needed when we have hi-fi design 👍

PR #392, opened 2022-08-15
Update README with how to get started in store app

simonauner:
## Describe your changes

Just a small change but maybe a larger discussion point on what to have in the README or in Notion...

IMHO a repository should contain all the information needed to run the project. Obviously, if we have sensitive information it is better to place it in a wiki, but all other info should be derived from the repo itself. Otherwise it's a guessing game which wiki page to go to in order to find the needed info.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@ Proof of concept monorepo to support all purchase journey related projects.
 ### Prerequisites
 
 - [Node.js](https://nodejs.org/en/)
-- [Yarn v1](https://yarnpkg.com)
+- [Yarn v3](https://yarnpkg.com)
 
 ### Install Node.js with `nvm`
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -7,7 +7,7 @@ Proof of concept monorepo to support all purchase journey related projects.
 ### Prerequisites
 
 - [Node.js](https://nodejs.org/en/)
-- [Yarn v1](https://yarnpkg.com)
+- [Yarn v3](https://yarnpkg.com)
```
simonauner, line 10: These are "prerequisites" but then we have details on how to install Node.js below? So if a dev follows these points by the number they will first install Node.js from the nodejs.org homepage, and then be told to install via NVM... So perhaps remove these entirely?

PR #393, opened 2022-08-15
Update yarn version in README 

gustaveen:
Keep readme up to date

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,16 +4,15 @@ Proof of concept monorepo to support all purchase journey related projects.
 
 ## Getting Started
 
-### Prerequisites
-
-- [Node.js](https://nodejs.org/en/)
-- [Yarn v1](https://yarnpkg.com)
-
 ### Install Node.js with `nvm`
 
 1. Install `nvm` [https://github.com/creationix/nvm](https://github.com/creationix/nvm)
 1. `nvm install` (installs the node version specified in `.nvmrc`)
 1. `nvm use` (use version from `.nvmrc`)
+
+### Packaga manager
+
+We use `yarn` as package manager. [Corepack](https://github.com/nodejs/corepack) installs the specified version defined in `package.json`
 
 ### Install dependencies
 


README.md
```diff
@@ -7,7 +7,7 @@ Proof of concept monorepo to support all purchase journey related projects.
 ### Prerequisites
 
 - [Node.js](https://nodejs.org/en/)
-- [Yarn v1](https://yarnpkg.com)
+- [Yarn v3](https://yarnpkg.com)
```
simonauner, line 10: Hehe, I was looking at this as well!

I see it as a problem that these two are pre-requisites and then below we instruct to install Node.js with NVM. Should we perhaps remove this section and add Yarn after installing Node.js with NVM? Otherwise the user might do a "native" install of Node.js above, and then have to install with NVM.

gustaveen: Hehe nice! 🙌 @simonauner 
True, that might be better. Maybe @alebedev and @KarnellSchultz could chime in that have done it from scratch recently, if it was confusing. 

Possibly add a section for YVM, or that's what I use at least :)

simonauner: Yep, I use both YVM and NVM 😁

alebedev: Since we're using node v16 with core pack, we don't need to install yarn or yvm explicitly - corepack will provide package manager automatically

gustaveen: sweet, didn't know that @alebedev 👌

KarnellSchultz: it's a bit confusion for sure. I had to ask myself today did i need to update my yarn to yarn v3.

perhaps we should mention core-pack in the readme as well?

README.md
```diff
@@ -4,17 +4,20 @@ Proof of concept monorepo to support all purchase journey related projects.
...
-
 ### Install Node.js with `nvm`
 
 1. Install `nvm` [https://github.com/creationix/nvm](https://github.com/creationix/nvm)
 1. `nvm install` (installs the node version specified in `.nvmrc`)
 1. `nvm use` (use version from `.nvmrc`)
 
+### Yarn version
+
+Install and use the specified yarn version
```
gustaveen, line 15: Did I understand it correctly that we could use `yvm` out of the box then? @alebedev

alebedev: We could but it works without yvm just as well.  corepack + package.json packageManager field ensure we get required version of yarn

gustaveen: Cool! Then let's remove this step if it's redundant 👍

alebedev (approved):
Looks good!

PR #395, opened 2022-08-15
GRW-1357 / Remove market-web from racoon

gustaveen:
## Describe your changes
- Remove market-web from racoon
- Remove racoon-market-web from Vercel

The initial plan was to move market-web to racoon. This was initiated before we started working on coverage store, that ultimately will replace market-web. So let's not keep it lying around 🧹


robinandeer (approved):

alebedev (approved):
Looks good

Are there any places outside of GitHub (notion, vercel, etc) that should be updated as well?

simonauner:
Nice. You can probably remove some of the menu stuff from `packages/ui` as well unless we want to have the same behaviour in the store?

gustaveen:
> Nice. You can probably remove some of the menu stuff from `packages/ui` as well unless we want to have the same behaviour in the store?

Ah yes, thanks! I'll have a look, I think it will differ a bit from that one so we could probably remove it 👍

PR #396, opened 2022-08-15
GRW-1351 / Add Timeline component

robinandeer:

Based on: https://mui.com/material-ui/react-timeline/

We deviated slightly from the Figma designs but it should be good enough for the prototype.

Co-authored with @simonauner.

[GRW-1351]: https://hedvig.atlassian.net/browse/GRW-1351?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/store/src/components/Timeline/CheckCircleIcon.tsx
+++ b/apps/store/src/components/Timeline/CheckCircleIcon.tsx
@@ -1,0 +1,41 @@
+import styled from '@emotion/styled'
+import { theme } from 'ui'
+
+type Props = {
+  color?: string
+  size?: number | string
+  strokeWidth?: number | string
+  className?: string
+}
+
+export const CheckCircleIcon = ({
+  color = theme.colors.gray900,
+  strokeWidth = 2,
+  size = 24,
+  className,
+}: Props) => (
+  <StyledSvg
+    className={className}
+    size={size}
+    viewBox="0 0 24 24"
+    fill="none"
+    xmlns="http://www.w3.org/2000/svg"
+  >
+    <path
+      d="M22 11.08V12C21.9988 14.1564 21.3005 16.2547 20.0093 17.9818C18.7182 19.709 16.9033 20.9725 14.8354 21.5839C12.7674 22.1953 10.5573 22.1219 8.53447 21.3746C6.51168 20.6273 4.78465 19.2461 3.61096 17.4371C2.43727 15.628 1.87979 13.4881 2.02168 11.3363C2.16356 9.18455 2.99721 7.13631 4.39828 5.49706C5.79935 3.85781 7.69279 2.71537 9.79619 2.24013C11.8996 1.7649 14.1003 1.98232 16.07 2.85999"
+      stroke={color}
+      strokeWidth={strokeWidth}
+      strokeLinecap="round"
+      strokeLinejoin="round"
+    />
+    <path
+      d="M22 4L12 14.01L9 11.01"
+      stroke={color}
+      strokeWidth={strokeWidth}
+      strokeLinecap="round"
+      strokeLinejoin="round"
+    />
+  </StyledSvg>
+)
+
+const StyledSvg = styled.svg<Props>(({ size }) => ({ width: size, height: size }))
--- a/apps/store/src/components/Timeline/Timeline.tsx
+++ b/apps/store/src/components/Timeline/Timeline.tsx
@@ -1,0 +1,49 @@
+import styled from '@emotion/styled'
+import { CheckCircleIcon } from './CheckCircleIcon'
+
+const ICON_SIZE = '2.5rem'
+
+export const Root = styled.ul({
+  display: 'flex',
+  flexDirection: 'column',
+})
+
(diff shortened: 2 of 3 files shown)


robinandeer:
* main
  * **PR #396** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/396" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/396?utm_source=stack-comment).

apps/store/src/components/Timeline/CheckCircleIcon.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+export const CheckCircleIcon = ({
+  color = theme.colors.gray900,
+  size = 24,
+  width = 2,
+  className,
+}: Props) => (
+  <StyledSvg
+    className={className}
+    size={size}
```
simonauner, line 19: Can't `size` be replaced here with `width` and `height` from `StyledSvg` and just have this element as a regular `svg`? Seems like a simplification. `width` can be `strokeWidth` perhaps? But I think that strokes generally should be replaced by paths since it's difficult to scale strokes automatically.

apps/store/src/components/Timeline/Timeline.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+export const Icon = () => {
+  return (
+    <IconWrapper>
+      <CheckCircleIcon size={ICON_SIZE} width={1.5} />
+    </IconWrapper>
+  )
+}
+
+export const IconWrapper = styled.div(({ theme }) => ({
```
simonauner, line 27: Doesn't have to be exported, right?

simonauner (commented):
Just some comments, I won't approve since I'm the co-author 🙃

apps/store/src/components/Timeline/CheckCircleIcon.tsx
robinandeer: Good idea! I will change to that. I think we still need the emotion wrapper since width/height attributes don't support rems...

Not sure how to proceed with replacing stroke widths -- perhaps we can ask the designers when we get the proper icons.

gustaveen: In general I think the aim should be that we don't need to handle `strokeWidth`. So the svgs should be exported as a joined path, that way it scales accordingly. But as you said, we can handle that when we have finalised icons 👍

apps/store/src/components/Timeline/Timeline.tsx
```diff
@@ -0,0 +1,51 @@
...
+const ICON_SIZE = '2.5rem'
+
+export const Root = styled.ul({
+  display: 'flex',
+  flexDirection: 'column',
+})
+
+export const Item = styled.li(({ theme }) => ({
+  position: 'relative',
+
```
gustaveen, line 13: nit: remove line-break in css

apps/store/src/components/Timeline/Timeline.tsx
```diff
@@ -0,0 +1,51 @@
...
+const IconWrapper = styled.div(({ theme }) => ({
+  paddingTop: theme.space[1],
+  paddingBottom: theme.space[1],
+}))
+
+export const Separator = styled.div({
+  display: 'flex',
+  flexDirection: 'column',
+
```
gustaveen, line 35: nit: remove line-break in css

apps/store/src/components/Timeline/CheckCircleIcon.tsx
```diff
@@ -0,0 +1,44 @@
+import styled from '@emotion/styled'
+import { theme } from 'ui'
+
+type Props = {
+  color?: string
+  width?: number | string
+  height?: number | string
```
gustaveen, line 7: As @simonauner said, we could use size for icons with a square bounding box :)

gustaveen (commented):
Great stuff guys! 🙌

apps/store/src/components/Timeline/CheckCircleIcon.tsx
robinandeer: wait what, didn't @simonauner say to replace `size` with `width + height`? 😅

simonauner: Hm... Maybe :D What I meant was that we could send in `size` and use it for `width`/ `height` CSS props

simonauner: I guess I confused HTML `width`/`height` on SVGs with Emotion's `styled` that sets it as CSS...

robinandeer: Aha oh then I got confused also. But yes, I think we need to pass it to css/Emotion in the end. The width/height attributes are just the HTML ones.

simonauner (approved):

PR #398, opened 2022-08-16
Add base Table component

robinandeer:

Confirmation page.

### DIFF
--- a/apps/store/src/components/Table/Table.tsx
+++ b/apps/store/src/components/Table/Table.tsx
@@ -1,0 +1,24 @@
+import styled from '@emotion/styled'
+
+type TableProps = { layout?: 'fixed' | 'auto' }
+
+export const Root = styled.table<TableProps>(({ layout }) => ({
+  width: '100%',
+  borderCollapse: 'collapse',
+  tableLayout: layout,
+}))
+
+export const Head = styled.thead({})
+
+export const Body = styled.tbody({})
+
+export const Row = styled.tr({})
+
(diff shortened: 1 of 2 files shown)


robinandeer:
* main
  * **PR #398** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/398" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/398?utm_source=stack-comment).

apps/store/src/components/Table/Table.tsx
```diff
@@ -0,0 +1,24 @@
+import styled from '@emotion/styled'
+
+type TableProps = { layout?: 'fixed' | 'auto' }
```
simonauner, line 3: Should one of these be a default perhaps so one doesn't have to decide which one to use in every situation?

simonauner (approved):
Nice with something to build on. Just had one thought that needs to be decided now, but perhaps as a question of component API design - do we want sane (some) defaults?

apps/store/src/components/Table/Table.tsx
robinandeer: "auto" is already the default in browsers

https://developer.mozilla.org/en-US/docs/Web/CSS/table-layout

simonauner: Oh, sorry, didn't spot the  `? ` and it was `fixed` in both story examples. Good 👍

PR #401, opened 2022-08-16
Grw 1335/refactor shop session hook

alebedev:
## Refactor shop session hook

- Improve types
- Return full Apollo results
- Prevent double creation of session from the client


apps/store/src/services/shopSession/ShopSessionContext.tsx
```diff
@@ -46,55 +35,60 @@ export const ShopSessionProvider = ({ children }: PropsWithChildren<unknown>) =>
...
+    ...rest,
   })
 }
 
 type UseShopSessionParams = {
   shopSessionId: string | null
-  onCompleted: (data: ShopSessionQuery) => void
-  onError: (error: ApolloError) => void
-}
+} & Omit<QueryHookOptions<ShopSessionQuery, Exact<{ shopSessionId: string }>>, 'variables'>
```
robinandeer, line 91: 😅 can we break this up -- I find it impossible to read

apps/store/src/services/shopSession/ShopSessionContext.tsx
```diff
@@ -46,55 +35,60 @@ export const ShopSessionProvider = ({ children }: PropsWithChildren<unknown>) =>
...
 }
 
 export const useShopSession = () => {
-  return useContext(ShopSessionContext)
+  const queryResult = useContext(ShopSessionContext)
+  if (!queryResult) {
+    throw new Error(
+      'useShopSession called from outside ShopSessionContextProvider, no value in context',
+    )
+  }
```
robinandeer, line 64: nit: is there any way to avoid this?

It can be useful to be able to call this outside a provider: https://reactjs.org/docs/context.html#reactcreatecontext

Or perhaps it's a good thing that we don't allow it?

apps/store/src/services/shopSession/ShopSessionContext.tsx
```diff
@@ -46,55 +35,60 @@ export const ShopSessionProvider = ({ children }: PropsWithChildren<unknown>) =>
...
 }
 
-const useShopSessionQuery = ({ shopSessionId, onCompleted, onError }: UseShopSessionParams) => {
+const useShopSessionQuery = ({ shopSessionId, ...rest }: UseShopSessionParams) => {
   return useShopSessionApolloQuery({
     variables: shopSessionId ? { shopSessionId } : undefined,
     skip: !shopSessionId,
-    onCompleted,
-    onError,
     ssr: false,
```
robinandeer, line 84: We could add a comment here why we don't allow SSR :)

Like: "Only intended to run client-side, prefetch and pass to Apollo Cache if fetched on server."

apps/store/src/services/shopSession/ShopSessionContext.tsx
alebedev: I see it more as a feature and not a bug.  You'll get safer type by throwing when component is used outside of context (most likely programmer error)

Got the idea here: https://kentcdodds.com/blog/how-to-use-react-context-effectively#the-custom-consumer-hook

apps/store/src/services/shopSession/ShopSessionContext.tsx
```diff
@@ -1,43 +1,32 @@
...
-})
+export const ShopSessionContext = createContext<ShopSessionQueryResult | null>(null)
 
 export const ShopSessionProvider = ({ children }: PropsWithChildren<unknown>) => {
   const { countryCode } = useCurrentLocale()
   const shopSessionService = useShopSessionService()
+  const shopSessionId = shopSessionService.shopSessionId()
 
-  const createShopSession = useCreateShopSession({
+  const [createShopSession, mutationResult] = useCreateShopSession({
```
guilhermespopolin, line 22: Suggestion: Remove `useCreateShopSession` and call `useShopSessionCreateMutation` directly here.

```
  const [createShopSession, mutationResult] = useShopSessionCreateMutation({
    variables: { countryCode },
    onCompleted: ({ shopSessionCreate }) => {
      shopSessionService.save(shopSessionCreate)
    },
  })
```

apps/store/src/services/shopSession/ShopSessionContext.tsx
robinandeer: Yeah, I'm a fan of that article also :D

I'm always reminded of this tweet from Seb Markbåge formerly of the React Core team:

https://twitter.com/sebmarkbage/status/1219829297034092545

He considers it an anti-pattern even if it's a very common pattern in the industry. I take it as an opportunity to think: is there a way to provide a meaningful default? Or can this be solved another way? :)

apps/store/src/services/shopSession/ShopSessionContext.tsx
```diff
@@ -46,55 +35,60 @@ export const ShopSessionProvider = ({ children }: PropsWithChildren<unknown>) =>
...
 
 export const useShopSession = () => {
-  return useContext(ShopSessionContext)
+  const queryResult = useContext(ShopSessionContext)
+  if (!queryResult) {
+    throw new Error(
+      'useShopSession called from outside ShopSessionContextProvider, no value in context',
+    )
+  }
+  return queryResult
```
guilhermespopolin, line 64: Suggestion:

```suggestion
  return queryResult?.data?.shopSession
```

I think it would make the hook more _ergonometric_ by not having to do `data.shopSession` every place we use it.

alebedev: We want to keep an option to use other properties of returned object, like `refetch`, `loading`, etc

Do you think it makes sense to destructure the data and keep everything else the same?  What worries me is the amount of type declarations to maintain to keep this code type-safe

apps/store/src/services/shopSession/ShopSessionContext.tsx
alebedev: Good point, helper function should either do more or be shorter to justify its existence.  Fixed

apps/store/src/services/shopSession/ShopSessionContext.tsx
robinandeer: Hmm I think we can just use `ShopSessionQuery['shopSession']`

Perhaps that would be worth it.

guilhermespopolin: I see. But since you've mentioned the increasing of type declarations, was you considering into create a new type that defines the return of `useShopSession`? Or do I misunderstood? 🤔

alebedev: We could do something like this
```
  const { data, ...other } = useShopSessionApolloQuery({
    variables: shopSessionId ? { shopSessionId } : undefined,
    skip: !shopSessionId,
    // Only intended to run client-side, prefetch and pass to Apollo Cache if fetched on server.
    ssr: false,
    ...rest,
  })
  return {
    ...other,
    shopSession: data?.shopSession,
  }
```

It actually infers return type, no need to declare anything

What do you think?  @guilhermespopolin @robinandeer

apps/store/src/services/shopSession/ShopSessionContext.tsx
alebedev: Since we're returning full Apollo query result, that object would be hard to emulate with trivial default.  Non-trivial default like making a query and skipping it looks fishy

Any suggestions?

apps/store/src/services/shopSession/ShopSessionContext.tsx
alebedev: This requires us to make context value type more complex though
```
type ShopSessionQueryResult = Omit<
  QueryResult<ShopSessionQuery, ShopSessionQueryVariables>,
  'data'
> & { shopSession: ShopSessionQuery['shopSession'] }
```

apps/store/src/services/shopSession/ShopSessionContext.tsx
robinandeer: nah I think this is fine -- let's just keep an eye out :)

apps/store/src/services/shopSession/ShopSessionContext.tsx
robinandeer: Not too bad, we could give it a go -- but you can also do that in another PR if you like!

guilhermespopolin: Oh, I see. I guess we could keep `ShopSessionQueryResult` as we had before and make the transformation on `useShopSession`

```javascript
type ShopSessionQueryResult = QueryResult<ShopSessionQuery, ShopSessionQueryVariables>

...

export const useShopSession = () => {
  const queryResult = useContext(ShopSessionContext)

  if (!queryResult) {
    throw new Error(
      'useShopSession called from outside ShopSessionContextProvider, no value in context',
    )
  }

  const { data, ...other } = queryResult
  return {
    shopSession: data?.shopSession ?? null,
    ...other,
  }
}

...

const { shopSession } = useShopSession()
```

In that way we don't need to create complex type for the provider and the return type of `useShopSession` can be inferred. But either way is fine by me 😄

guilhermespopolin (approved):

apps/store/src/services/shopSession/ShopSessionContext.tsx
alebedev: Sounds great, I'll move transform logic to useShopSession.  Makes each of the hooks cleaner

apps/store/src/services/shopSession/ShopSessionContext.tsx
```diff
@@ -46,54 +34,49 @@ export const ShopSessionProvider = ({ children }: PropsWithChildren<unknown>) =>
...
       createShopSession()
     },
   })
 
-  const value = useMemo<ShopSessionContextValue>(() => {
-    const shopSessionCountryCode = result.data?.shopSession?.countryCode
-    return {
-      shopSession: shopSessionCountryCode !== countryCode ? null : result.data?.shopSession ?? null,
-      loading: result.loading,
+  // Has to be wrapped to prevent duplicate execution (Apollo quirk leads do duplicate execution when called directly from render)
```
robinandeer, line 42: I don't know if this is needed, I think it makes sense to put it into useEffect

robinandeer (approved):
Looks good, lets try it out :D

apps/store/src/services/shopSession/ShopSessionContext.tsx
alebedev: Let's leave it a memory for some painful debugging for now :)

guilhermespopolin (approved):
Well done :)

PR #406, opened 2022-08-17
Ignore shop session when countryCode does not match current locale

alebedev:

We may get old session that does not match current country.  In this case new session is created and while we wait, old session data should not be displayed in UI

### DIFF
--- a/apps/store/src/services/shopSession/ShopSessionContext.tsx
+++ b/apps/store/src/services/shopSession/ShopSessionContext.tsx
@@ -51,16 +51,25 @@ export const ShopSessionProvider = ({ children }: PropsWithChildren<unknown>) =>
 }
 
 export const useShopSession = () => {
+  const { countryCode } = useCurrentLocale()
+
   const queryResult = useContext(ShopSessionContext)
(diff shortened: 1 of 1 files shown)


apps/store/src/services/shopSession/ShopSessionContext.tsx
```diff
@@ -51,16 +51,25 @@ export const ShopSessionProvider = ({ children }: PropsWithChildren<unknown>) =>
...
+
+  let shopSession = data?.shopSession
+  if (shopSession?.countryCode !== countryCode) {
+    // Ignore session from different country.  This probably means old session was fetched and new one is being created
+    shopSession = undefined
+  }
   return {
     ...other,
-    shopSession: data?.shopSession,
+    shopSession,
```
robinandeer, line 72: nit: I would just write this in one line

```suggestion
shopSession: data?.shopSession.countryCode !== countryCode ? undefined : data?.shopSession
```

robinandeer (approved):

robinandeer (commented):
@alebedev we can merge this, right?

PR #407, opened 2022-08-17
GRW-1359 / Add global story

gustaveen:
## Describe your changes
Add global story data to pages

To be able to fetch content from a the content type we call `global` where we will store header, menus, footer etc
Let me know if you have other ideas how to handle it :) 

### DIFF
--- a/apps/store/src/pages/[...slug].tsx
+++ b/apps/store/src/pages/[...slug].tsx
@@ -1,7 +1,7 @@
-import { StoryblokComponent, useStoryblokState } from '@storyblok/react'
 import type { GetStaticPaths, GetStaticProps, NextPageWithLayout } from 'next'
 import Head from 'next/head'
 import { LayoutWithMenu } from '@/components/LayoutWithMenu/LayoutWithMenu'
+import { Page } from '@/components/Page/Page'
 import {
   getAllLinks,
   getStoryBySlug,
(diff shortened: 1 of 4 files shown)


gustaveen:
Not sure why the build fails though, I could build it all locally 🧐

apps/store/src/services/storyblok/storyblok.ts
```diff
@@ -59,6 +60,8 @@ export type ProductStory = StoryData & {
   }
 }
 
+type GlobalStory = StoryData
```
gustaveen, line 63: Here we can add the fields you created for the global content type @KarnellSchultz 🙌

apps/store/src/pages/[...slug].tsx
```diff
@@ -33,10 +33,12 @@ export const getStaticProps: GetStaticProps<StoryblokPageProps, StoryblokQueryPa
 }) => {
   const slug = params?.slug ? params.slug.join('/') : 'home'
   const story = await getStoryBySlug(slug, preview)
+  const globalStory = await getGlobalStory(preview)
```
Comment (robinandeer) [inline apps/store/src/pages/[...slug].tsx:36]: nit: this could be run in parallel

```suggestion
const [story, globalStory] = await Promise.all([
  getStoryBySlug(slug, preview)
  getGlobalStory(preview)
])
```

robinandeer (approved):
Interesting :) will be cool to see how this works

gustaveen:
> Interesting :) will be cool to see how this works

it's one way of doing it so yeah let's see if it works good for us :)

apps/store/src/services/storyblok/storyblok.ts
KarnellSchultz: Nice!

PR #409, opened 2022-08-18
Add App Store badges

robinandeer:
## Describe your changes

Add lokalised app store badges.

I used the official resources from Google and Apple. Google only gives PNG and Apple offers SVG. I used the preferred color (black).

For some strange reason the English Google Play store badge has a different icons to the rest 🤷‍♂️

It would be better to load resources on demand but that's an optimization we can do later.


robinandeer:
* main
  * **PR #409** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/409" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/409?utm_source=stack-comment).

apps/store/src/components/AppStoreBadge/AppStoreBadge.tsx
```diff
@@ -0,0 +1,50 @@
...
+  [Locale.DaDk]: GoogleDaDk,
+  [Locale.EnDk]: GoogleEn,
+  [Locale.NbNo]: GoogleNbNo,
+  [Locale.EnNo]: GoogleEn,
+}
+
+export const AppStoreBadge = ({ type, locale }: AppStoreBadgeProps) => {
+  if (type === 'apple') {
+    const Component = APPLE_BADGE[locale]
+    return <Component />
```
simonauner, line 38: Will there be any error if we introduce a new locale that has no badge? Will it be a runtime error or can we detect it and throw or just return null?

simonauner (approved):
Nice. I guess the English Google Play icon has the new Play store design, it was recently updated.

https://9to5google.com/2022/07/27/google-play-new-logo-2022/

apps/store/src/components/AppStoreBadge/AppStoreBadge.tsx
robinandeer: Hmm, yeah perhaps I can add a fallback.

Usually when we add support for a new locale/country it's a big project where we go over the whole flow/code base so pretty sure we would catch this :)

simonauner: Yeah, it's no biggie

PR #413, opened 2022-08-19
Grw-1359 topMenu navigation in Storyblock

KarnellSchultz:
## Describe your changes

Created a `topMenu` storyblok component. This allows users to create a custom menu with nested links. 

The topMenu lives in the new `global` components

https://user-images.githubusercontent.com/50870173/185606002-2d7a00a0-68f9-4e36-9380-60e64addb1af.mov


apps/store/src/pages/[...slug].tsx
```diff
@@ -56,9 +56,9 @@ export const getStaticPaths: GetStaticPaths = async () => {
     .filter((linkKey) => !links[linkKey].slug.startsWith('products/'))
     .forEach((linkKey) => {
       const slug = links[linkKey].slug
-      const splittedSlug = slug.split('/')
+      const splitSlug = slug.split('/')
```
Comment (KarnellSchultz) [inline apps/store/src/pages/[...slug].tsx:59]: a typo

apps/store/src/services/storyblok/storyblok.ts
```diff
@@ -57,58 +58,64 @@ export type ProductStory = StoryData & {
...
     priceFormTemplateId: string
     body: Array<SbBlokData>
+    global: Array<SbBlokData>
   }
 }
 
 type GlobalStory = StoryData
 
 export enum StoryblokBlockName {
+  Accordion = 'accordion',
```
KarnellSchultz, line 68: alphabetise these objects

apps/store/src/pages/[...slug].tsx
Comment (simonauner) [reply apps/store/src/pages/[...slug].tsx:59]: very great English :v:

apps/store/src/services/storyblok/storyblok.ts
```diff
@@ -57,58 +58,64 @@ export type ProductStory = StoryData & {
...
-    [StoryblokBlockName.ProductSlideshow]: ProductSlideshowBlock,
-    [StoryblokBlockName.Accordion]: AccordionBlock,
-    [StoryblokBlockName.AccordionItem]: AccordionItemBlock,
     [StoryblokBlockName.Tabs]: TabsBlock,
-    [StoryblokBlockName.ContactSupport]: ContactSupportBlock,
-    [StoryblokBlockName.Content]: ContentBlock,
-    [StoryblokBlockName.CheckList]: CheckListBlock,
     [StoryblokBlockName.Text]: TextBlock,
-    [StoryblokBlockName.Image]: ImageBlock,
+    [StoryblokBlockName.TopMenu]: TopMenuBlock,
```
robinandeer, line 117: I wonder if we should add this here since it's not a reusable block 🤔

apps/store/src/components/Page/Page.tsx
```diff
@@ -1,7 +1,16 @@
...
+const isStoryData = (global: StoryData | boolean): global is StoryData => {
+  return (global as StoryData).content !== undefined
+}
+
+export const Page = ({ story: initialStory, globalStory }: StoryblokPageProps) => {
   const story = useStoryblokState(initialStory)
-  return <StoryblokComponent blok={story.content} />
+  return (
+    <>
+      {isStoryData(globalStory) && <StoryblokComponent blok={globalStory.content} />}
```
robinandeer, line 12: Perhaps we don't need to use `StoryblokComponent` since we know it can only be `TopMenu`, or?

apps/store/src/blocks/TopMenuBlock.tsx
```diff
@@ -0,0 +1,208 @@
...
+import { ArrowForwardIcon, CrossIcon, theme } from 'ui'
+import { LinkField, SbBaseBlockProps } from '@/services/storyblok/storyblok'
+import { MenuIcon } from '../components/TopMenu/MenuIcon'
+import { ShoppingCartMenuItem } from '../components/TopMenu/ShoppingCartMenuItem'
+
+const MENU_BAR_HEIGHT = '3.75rem'
+
+type NestedNavContainerBlockProps = SbBaseBlockProps<{
+  name: string
+  nav_item: SbBlokData[]
```
robinandeer, line 16: any reason not to use camelCase?

apps/store/src/blocks/TopMenuBlock.tsx
```diff
@@ -0,0 +1,208 @@
...
+}
+
+export const NestedNavContainerBlock = ({ blok, ...rest }: NestedNavContainerBlockProps) => {
+  const { activeItem } = rest
+  return (
+    <NavigationMenuPrimitive.Item key={blok._uid} value={blok.name}>
+      <NavigationTrigger>
+        {blok.name}
+        {activeItem === `${blok.name}` ? (
+          <CrossIcon size="16px" />
```
robinandeer, line 29: nit: use rems

apps/store/src/blocks/TopMenuBlock.tsx
```diff
@@ -0,0 +1,208 @@
...
+        ) : (
+          <ArrowForwardIcon size="16px" />
+        )}
+      </NavigationTrigger>
+      <NavigationMenuPrimitive.Content>
+        <NavigationMenuPrimitive.Sub defaultValue={blok.name}>
+          <NavigationSecondaryList>
+            {blok.nav_item &&
+              blok.nav_item.map((nestedBlok) => (
```
robinandeer, line 38: nit: should this be "navItems" (plural)?

apps/store/src/blocks/TopMenuBlock.tsx
```diff
@@ -0,0 +1,208 @@
...
+}> & {
+  closeDialog: () => void
+}
+
+export const NavItemBlock = ({ blok, ...rest }: NavItemBlockProps) => {
+  const { closeDialog } = rest
+
+  return (
+    <NavigationMenuPrimitive.Item key={blok._uid} value={blok.name}>
+      <NavigationLink href={blok.link?.url} onSelect={closeDialog}>
```
robinandeer, line 60: What is the logic of having an optional "link"? Should we not make it mandatory?

apps/store/src/blocks/TopMenuBlock.tsx
```diff
@@ -0,0 +1,208 @@
...
+  const [open, setOpen] = useState(false)
+
+  const closeDialog = useCallback(() => {
+    setOpen(false)
+    setActiveItem('')
+  }, [])
+
+  return (
+    <Wrapper>
+      <DialogPrimitive.Root open={open} onOpenChange={() => setOpen((prevOpen) => !prevOpen)}>
```
robinandeer, line 81: Just a question: do we need to make this a controlled component? Or can we just skip the state stuff?

robinandeer (commented):
Great stuff :D :D 

Sorry for the many comments hehe

apps/store/src/blocks/TopMenuBlock.tsx
KarnellSchultz: Nope. I was following a storybook guide and they used snake_case so it ended up in my code too. Going to swap it out everywhere.

apps/store/src/services/storyblok/storyblok.ts
KarnellSchultz: I _think_ this needs to be here so the component shows up within StoryBlok. As in SB would not know what the `TopMenuBlok` was if it was not here. 

Maybe I'm misunderstanding that part though.

apps/store/src/blocks/TopMenuBlock.tsx
KarnellSchultz: Instead of copying all the code over from the actual TopMenu component and in the spirit of DRY I've imported all of the styles and the non specific 'block' parts to this component.

apps/store/src/blocks/TopMenuBlock.tsx
KarnellSchultz: Made the link required here.

apps/store/src/blocks/TopMenuBlock.tsx
KarnellSchultz: This was all copied directly from the TopMenu component. Perhaps if we decide to refactor the component itself it'd be worth looking into those things?

apps/store/src/components/Page/Page.tsx
KarnellSchultz: @gustaveen Want to get your opinion as well. 

I think if we don't use the StoryblokComponent here then the block would not be usable within the 'global' components we're creating. Also in Storyblok you would not be able to add / remove the block as it'd be a fixed component as we had before.

apps/store/src/blocks/TopMenuBlock.tsx
robinandeer: hmm that sounds good but should we really us "16px" here instead of "1rem"?

apps/store/src/blocks/TopMenuBlock.tsx
robinandeer: 👍

robinandeer (approved):

apps/store/src/blocks/TopMenuBlock.tsx
```diff
@@ -0,0 +1,119 @@
...
+  NavigationLink,
+  NavigationPrimaryList,
+  NavigationSecondaryList,
+  NavigationTrigger,
+  ToggleMenu,
+  Wrapper,
+} from '@/components/TopMenu/TopMenu'
+import { LinkField, SbBaseBlockProps } from '@/services/storyblok/storyblok'
+import { MenuIcon } from '../components/TopMenu/MenuIcon'
+import { ShoppingCartMenuItem } from '../components/TopMenu/ShoppingCartMenuItem'
```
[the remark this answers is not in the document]
KarnellSchultz: Ahhh thanks! I needed that 🧠

apps/store/src/blocks/TopMenuBlock.tsx
KarnellSchultz: I missed this 'px' use. 🔧 switching to rems ✅

apps/store/src/components/Page/Page.tsx
gustaveen: Good question, I think these things will come down to case by case usage. If we use a sub-block and it only allows for one block type, we could probably be specific in those cases what block we want to render. But I'm still a bit unsure if it would impact how the block is editable in Storyblok 🤔 Let's try it like this first 👍

apps/store/src/blocks/TopMenuBlock.tsx
```diff
@@ -0,0 +1,119 @@
...
+        </NavigationMenuPrimitive.Sub>
+      </NavigationMenuPrimitive.Content>
+    </NavigationMenuPrimitive.Item>
+  )
+}
+
+type TopMenuBlockProps = SbBaseBlockProps<{
+  navMenuContainer: SbBlokData[]
+}>
+export const TopMenuBlock = ({ blok }: TopMenuBlockProps) => {
```
gustaveen, line 78: nit: Thinking about this component it would make sense to call it `HeaderBlock` IMO, since it holds more than just a menu. But let's nail the naming later 🙌

gustaveen (approved):
Awesome work with the menu @KarnellSchultz 🙌

I know it can be a bit confusing handling the structure for this in Storyblok :D 

Let's merge this and iterate. Just some general things could be to break up the components in `TopMenBlock` into separate files

apps/store/src/blocks/TopMenuBlock.tsx
KarnellSchultz: Thanks 👍 I'll change the name and get this thing merged.

PR #414, opened 2022-08-22
Add winston for logging to Datadog

simonauner:
## Describe your changes

Added Winston.js for logging to Datadog. Winston is set up with a log forwarder that sends all logs to Datadog. We do however already have Datadog set up as an integration (https://vercel.com/hedvig/racoon/settings/integrations) which already forwards all (Next.js) logs to Datadog. It does seem that we do not have correct environment variables set for that forwarder, because we fail to get which environment it's running on (local/dev/prod) and the service name is a bit iffy, see below:

Screenshot 2022-08-22 at 13 30 56

I've tried to fix that with setting correct environment variables in Vercel, and if that works maybe we don't need Winston as a separate logger, we can just use `console.log`. That's why I'm adding double logs to this PR, so that we can see how it looks with the Datadog integration vs Winston (`pages/new-member/index.tsx`).

Another thing that we should evaluate and maybe ditch is the monkey-patching of logs and forwarding to Datadog in `server-preload.js` - it's _a lot_ of noise. This is just `info` level and higher and I think it's already a mess trying to find things. Request logs should be `trace` or something, but I haven't looked into if that can be configured.

image

Some scenarios that might pan out:

1. We do _not_ need monkey patching of logging in `server-preload.js` and `console.log` is forwarded correctly. Then Winston.js can be removed from this PR and the monkey patching also.
2. We _do_ need monkey patching of logging in `server-preload.js` and `console.log` is forwarded correctly. Then Winston.js can be removed from this PR.
3. `console.log` isn't forwarded correctly. We need Winston.js - maybe the monkey patching can be removed?

We only need one logging solution, and we need it to be consistent between Logs, APM and RUM - meaning that it's the same service name and application ID across all three Datadog "ingestion points".

The reason for choosing Winston.js is that it's in the Datadog documentation and was easy to configure with forwarding to Datadog servers. https://docs.datadoghq.com/logs/log_collection/nodejs/?tab=winston30


apps/onboarding/src/logger.js
```diff
@@ -0,0 +1,26 @@
...
+  }),
+)
+
+logger.add(
+  new transports.Console({
+    format: format.simple(),
+  }),
+)
+
+module.exports = logger
```
robinandeer, line 26: hmm why not `export` or `export default`?

apps/onboarding/src/pages/api/pages/start.ts
```diff
@@ -1,6 +1,7 @@
 import type { NextApiRequest, NextApiResponse } from 'next'
 import { handleStartPageForm } from '@/components/StartPage/StartPage.action'
 import { getFormData } from '@/lib/get-form-data'
+import logger from 'logger'
```
robinandeer, line 4: I would prefer to alias this import so we don't mix it up with external dependencies :)

Like: `import logger from '@/logger'

apps/onboarding/src/logger.js
simonauner: Just lazy copy-pasting from Datadog's tutorial, but I'll fix it

apps/onboarding/src/pages/api/pages/start.ts
simonauner: Ok, fixed 👍

simonauner:
After deployment it is clear to see that `console.log` isn't picked up by the Datadog integration in Vercel. The only log that shows is the one from Winston.js. 

image

I will continue to remove the `server-preload.js` as it doesn't seem to work. I believe that the logs that are being forwarded to Datadog is Vercel's own logs, not the ones we try to add.

simonauner:
Since the code in this PR is not really the same as it started out with I'll open a new PR with a clean slate.

PR #418, opened 2022-08-22
Fix campaign code middleware bug

robinandeer:
## Describe your changes

Remove async keyword from sync function.

Add log message.

Bug.

### DIFF
--- a/apps/onboarding/src/middleware.ts
+++ b/apps/onboarding/src/middleware.ts
@@ -30,7 +30,7 @@ const guessLocale = (request: NextRequest) => {
   return null
 }
 
-const localeRedirectMiddleware = async (request: NextRequest) => {
+const localeRedirectMiddleware = (request: NextRequest) => {
   const shouldHandleLocale =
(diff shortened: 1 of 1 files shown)


robinandeer:
* main
  * **PR #418** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/418" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/418?utm_source=stack-comment).

alebedev (approved):
There's a way to catch things like this with typescript.  If you declare return type explicitly it'll act as an assertion for what you're actually returning.  For example, original bug is caught
by following declaration
```
export function middleware(request: NextRequest): NextResponse {
  ...
}

gustaveen (approved):
🕵️‍♂️

robinandeer (commented):
@alebedev very true! I will add that, thx :D

PR #421, opened 2022-08-23
Add Pino for logging to Datadog

simonauner:
So, my draft PR got a bit messy with all the commits and re-bases https://github.com/HedvigInsurance/racoon/pull/414, starting from a clean slate here.

## Describe your changes

Added Pino set up with a log forwarder/transport that sends logs to Datadog.

Had to add `topLevelAwait` allow Pino's init function to work. 

Removed `server-preload.js` which doesn't seem to work anymore. In it we monkey-patched `console.log` and added Datadogs tracing lib `dd-trace`, but it wasn't applied anyway, so might as well remove it to remove confusion about how we log.

We already have Datadog set up as an integration in Vercel (https://vercel.com/hedvig/racoon/settings/integrations) which already forwards all Next.js logs to Datadog. This does not cover `console.log`s though - all we get is request logs:

Screenshot 2022-08-23 at 13 05 55

With Pino in place we have a way to "debug" server code in prod/staging.


apps/onboarding/package.json
```diff
@@ -37,7 +36,8 @@
     "marked": "4.0.18",
     "next": "12.2.2",
     "next-i18next": "^11.0.0",
-    "pino": "^7.10.0",
+    "pino-datadog": "https://github.com/simonauner/pino-datadog#rename-hostname-to-host",
```
simonauner, line 39: Currently running a customized version of `pino-datadog` because of https://github.com/ovhemert/pino-datadog/pull/89

apps/onboarding/public/mockServiceWorker.js
```diff
@@ -128,7 +128,7 @@ self.addEventListener('fetch', function (event) {
       // At this point, any exception indicates an issue with the original request/response.
       console.error(
         `\
-[MSW] Caught an exception from the "%s %s" request (%s). This is probably not a problem with Mock Service Worker. There is likely an additional logging output above.`,
+ [MSW] Caught an exception from the "%s %s" request (%s). This is probably not a problem with Mock Service Worker. There is likely an additional logging output above.`,
```
simonauner, line 131: I accidentally opened this file and saved it, and prettier performed its magic 🤷

apps/onboarding/src/services/datadog.ts
```diff
@@ -9,9 +9,9 @@ export const initRum = () => {
       applicationId: APPLICATION_ID,
       clientToken: CLIENT_TOKEN,
       site: 'datadoghq.eu',
-      service: 'racoon-onboarding',
+      service: process.env.NEXT_PUBLIC_DATADOG_SERVICE_NAME,
       version: process.env.NEXT_PUBLIC_DATADOG_VERSION,
-      env: process.env.NEXT_PUBLIC_DATADOG_ENV || 'dev',
+      env: process.env.NEXT_PUBLIC_DATADOG_ENV || 'local',
```
simonauner, line 14: `dev` is sometimes referred to as our staging environment, `local` cannot be misinterpreted as anything other than "running on my machine"?

apps/onboarding/src/services/logger.js
```diff
@@ -0,0 +1,16 @@
+import datadog from 'pino-datadog'
+import pinoms from 'pino-multi-stream'
+
+const writeStream = await datadog.createWriteStream({
+  apiKey: process.env.DATADOG_API_KEY,
+  hostname: process.env.VERCEL_URL || 'localhost',
+  // URL will be racoon-onboarding.vercel.app when deployed on Vercel, but empty string when running locally.
+  // NEXT_PUBLIC_DATADOG_SERVICE_NAME will be racoon-onboarding, so this is just a way to set a meaningful
+  // service name when running locally.
+  service: process.env.VERCEL_URL || `${process.env.NEXT_PUBLIC_DATADOG_SERVICE_NAME}.local`,
```
simonauner, line 10: Let's see what `process.env.VERCEL_URL` translates to when deployed to staging and prod. It might be that it's the machine-name, not the URL, and then this will have to be changed to something else.

robinandeer (approved):
No major comments from me

simonauner:
Ok, so apparently there is a new Pino version out which doesn't require `pino-multi-stream`.... Will work on that instead: https://github.com/HedvigInsurance/racoon/pull/422

PR #422, opened 2022-08-23
Grw 1388/feat/logging3

simonauner:


simonauner:
This is how we _should_ log to Datadog, but it crashes when running it on Vercel. Works well locally though. To be revisited in the future.

PR #425, opened 2022-08-23
Simplify block name mapping, put block names into block component files

alebedev:
- Less boilerplate
- Block declarations are more self-contained now

[GRW-1403]: https://hedvig.atlassian.net/browse/GRW-1403?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/store/src/blocks/CheckListBlock.tsx
+++ b/apps/store/src/blocks/CheckListBlock.tsx
@@ -3,13 +3,14 @@ import { SbBlokData, storyblokEditable } from '@storyblok/react'
 import { useMemo } from 'react'
 import { TextBlock, TextBlockProps } from '@/blocks/TextBlock'
 import * as CheckList from '@/components/CheckList/CheckList'
-import { SbBaseBlockProps, StoryblokBlockName } from '@/services/storyblok/storyblok'
+import { SbBaseBlockProps } from '@/services/storyblok/storyblok'
 
 type CheckListBlockProps = SbBaseBlockProps<{
   items: Array<SbBlokData>
 }>
 
 export const CheckListBlock = ({ blok }: CheckListBlockProps) => {
+  // TODO: Ensure on storyblok level (acceptable child type)
   const items = useMemo(() => blok.items.filter(isTextBlock), [blok.items])
 
   return (
@@ -24,9 +25,10 @@ export const CheckListBlock = ({ blok }: CheckListBlockProps) => {
     </Wrapper>
   )
 }
+CheckListBlock.blockName = 'checkList'
 
 const isTextBlock = (blok: SbBlokData): blok is TextBlockProps['blok'] => {
-  return blok.component === StoryblokBlockName.Text
+  return blok.component === TextBlock.blockName
 }
 
 const Wrapper = styled.div(({ theme }) => ({
--- a/apps/store/src/services/storyblok/storyblok.ts
+++ b/apps/store/src/services/storyblok/storyblok.ts
@@ -19,6 +19,7 @@ import { TabsBlock } from '@/blocks/TabsBlock'
 import { TextBlock } from '@/blocks/TextBlock'
 import { NavItemBlock, NestedNavContainerBlock, HeaderBlock } from '@/blocks/TopMenuBlock'
 import { TopPickCardBlock } from '@/blocks/TopPickCardBlock'
+import { ReactNode } from 'react'
 
 export type SbBaseBlockProps<T> = {
   blok: SbBlokData & T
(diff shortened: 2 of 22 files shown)


apps/store/src/services/storyblok/storyblok.ts
```diff
@@ -64,58 +65,40 @@ export type ProductStory = StoryData & {
...
 
 export const initStoryblok = () => {
+  const blockComponents: Array<NamedBlock> = [
+    AccordionBlock,
+    AccordionItemBlock,
+    ButtonBlock,
+    CheckListBlock,
+    ContactSupportBlock,
+    ContentBlock,
+    // TODO: Header vs Heading is easy to confuse.  Discuss with team if we should rename one of these
```
gustaveen, line 80: IMO I think this is fine since thet represent the elements they render, `header` and `h1` (heading) :)

apps/store/src/services/storyblok/storyblok.ts
```diff
@@ -64,58 +65,40 @@ export type ProductStory = StoryData & {
...
+    ProductCardBlock,
+    ProductGridBlock,
+    ProductSlideshowBlock,
+    ProductSummaryBlock,
+    SpacerBlock,
+    TabsBlock,
+    TextBlock,
+    TopPickCardBlock,
+  ]
+  const blockAliases = { product: PageBlock }
```
simonauner, line 98: Is this because `PageBlock` component should be rendered both when we have a `page` and `product` string?

robinandeer (approved):

gustaveen (approved):
LGTM! 🙌 Personally I don't mind having them declared in one file, but could be nice to co-locate them with the component

simonauner (approved):
Nice! It's great that the name is in the block component's file, makes it easier to maintain 👍

apps/store/src/services/storyblok/storyblok.ts
```diff
@@ -64,58 +65,40 @@ export type ProductStory = StoryData & {
...
-    [StoryblokBlockName.ProductGrid]: ProductGridBlock,
-    [StoryblokBlockName.ProductSlideshow]: ProductSlideshowBlock,
-    [StoryblokBlockName.ProductSummary]: ProductSummaryBlock,
-    [StoryblokBlockName.Spacer]: SpacerBlock,
-    [StoryblokBlockName.Tabs]: TabsBlock,
-    [StoryblokBlockName.Text]: TextBlock,
-    [StoryblokBlockName.TopMenu]: HeaderBlock,
-    [StoryblokBlockName.TopPickCard]: TopPickCardBlock,
+    ...Object.fromEntries(blockComponents.map((x) => [x.blockName, x])),
+    ...blockAliases,
```
KarnellSchultz, line 101: nit: could we give `x` a more descriptive name?

```suggestion
    ...Object.fromEntries(blockComponents.map((blockComponent) => [blockComponent.blockName, blockComponent])),
    ...blockAliases,
```

apps/store/src/services/storyblok/storyblok.ts
robinandeer: Yeah at least for now, but if we start to lock down the product page more I guess we would use a separate CMS content type

apps/store/src/blocks/CheckListBlock.tsx
```diff
@@ -3,13 +3,14 @@ import { SbBlokData, storyblokEditable } from '@storyblok/react'
...
 import * as CheckList from '@/components/CheckList/CheckList'
-import { SbBaseBlockProps, StoryblokBlockName } from '@/services/storyblok/storyblok'
+import { SbBaseBlockProps } from '@/services/storyblok/storyblok'
 
 type CheckListBlockProps = SbBaseBlockProps<{
   items: Array<SbBlokData>
 }>
 
 export const CheckListBlock = ({ blok }: CheckListBlockProps) => {
+  // TODO: Ensure on storyblok level (acceptable child type)
```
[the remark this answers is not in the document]
alebedev: This is separate discussion from this PR I guess.  We want to check "child block type" invariant somewhere, should it be just one place or both CMS and code?

apps/store/src/services/storyblok/storyblok.ts
alebedev: Maybe Header -> PageHeader?

simonauner: I agree with naming it more explicitly. CardHeader, PageHeader, etc.

gustaveen: Works for me 👍 We should probably have the same prefix for footer then :)

alebedev: Will fix in separate PR

PR #426, opened 2022-08-24
Fix calculating price for Swedish apartment

robinandeer:
## Describe your changes

Add hidden input field option.

Restrict default form value to string (for now).

Refactor combineFormTemplate to avoid ts-ignore comments.

Update form template for Swedish apartment.

Fix bug in ServerCookiePersister to allow calling save multiple times.

Make calculating and adding home insurance to the cart possible!

### DIFF
--- a/apps/store/src/services/persister/ServerCookiePersister.ts
+++ b/apps/store/src/services/persister/ServerCookiePersister.ts
@@ -9,7 +9,9 @@ export class ServerCookiePersister implements SimplePersister {
   ) {}
 
   public save(id: string) {
-    this.response.setHeader('Set-Cookie', [`${this.cookieKey}=${id}; Path=/`])
+    const rawHeader = this.response.getHeader('Set-Cookie')
+    const existingHeaders = Array.isArray(rawHeader) ? rawHeader : []
+    this.response.setHeader('Set-Cookie', [...existingHeaders, `${this.cookieKey}=${id}; Path=/`])
   }
 
   public fetch() {
--- a/apps/store/src/components/PriceCalculatorForm/DynamicField.tsx
+++ b/apps/store/src/components/PriceCalculatorForm/DynamicField.tsx
@@ -46,6 +46,9 @@ export const DynamicField = (props: Props) => {
         />
       )
 
+    case 'hidden':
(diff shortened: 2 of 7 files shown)


robinandeer:
* main
  * **PR #426** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/426" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/426?utm_source=stack-comment).

apps/store/src/services/persister/ServerCookiePersister.ts
```diff
@@ -9,7 +9,9 @@ export class ServerCookiePersister implements SimplePersister {
   ) {}
 
   public save(id: string) {
-    this.response.setHeader('Set-Cookie', [`${this.cookieKey}=${id}; Path=/`])
+    const rawHeader = this.response.getHeader('Set-Cookie')
```
alebedev, line 12: There's another case to think of - what if we already have another cookie with cookieKey before setting new value?

This is quite confusing from what I've found: https://stackoverflow.com/questions/4056306/how-to-handle-multiple-cookies-with-the-same-name

Maybe we need some higher level abstraction on top of raw cookies to handle that.  Or at least check for old value to overwrite it instead of appending

simonauner (approved):
Nice! I guess one of the challenges of handling forms and typing is that it's all strings, right? Maybe unless we use `<input type="number">`, but that is a bit iffy IIRC so I'm not suggesting that we use it. Just a mental note 🙃

apps/store/src/services/persister/ServerCookiePersister.ts
robinandeer: Yeah that's true. I think this is a worthwhile improvement for now though :)

Next already has support for managing cookies but for some reason it works only in API routes...

robinandeer (commented):
@simonauner yeah, I just found out you can use `.valueAsNumber` on number-inputs (https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement#valueasnumber)

Let's see when we get to more complex input types like extra buildings. Hopefully we can let the browser serialize most things and validate in an API route I'm thinking :)

PR #429, opened 2022-08-24
GRW-1403/chore/refactor-sublock-filter

alebedev:
- Throw error in development
- Safely hide incorrect child blocks in production mode
- Avoid extra indirection of StoryblockComponent

[GRW-1403]: https://hedvig.atlassian.net/browse/GRW-1403?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/store/src/services/storyblok/blocksOfType.ts
```diff
@@ -0,0 +1,18 @@
+import { SbBlokData } from '@storyblok/react'
+
+export function blocksOfType<BlockData extends SbBlokData>(
```
robinandeer, line 3: nit: use `const blocksOfType = () =>` syntax :)

I would consider renaming the function something like `filterBlocksByType` maybe?

apps/store/src/blocks/AccordionBlock.tsx
```diff
@@ -19,15 +20,14 @@ export const AccordionBlock = ({ blok }: Props) => {
           </StyledHeading>
         )}
         <div>
-          {blok.items.map((nestedBlock) => (
-            <StoryblokComponent key={nestedBlock._uid} blok={nestedBlock} />
+          {blocksOfType(blok.items, AccordionItemBlock.blockName).map((nestedBlock) => (
```
robinandeer, line 23: nit: memoize this with `useMemo`

robinandeer (commented):
Nice work :D

This looks like a helper function so perhaps we should try to follow the [file structure conventions](https://www.joshwcomeau.com/react/file-structure/) => `Storyblok.helpers.ts`

apps/store/src/blocks/AccordionBlock.tsx
```diff
@@ -1,12 +1,13 @@
...
 import { Heading, Space } from 'ui'
+import { AccordionItemBlock, AccordionItemBlockProps } from '@/blocks/AccordionItemBlock'
 import * as Accordion from '@/components/Accordion/Accordion'
 import { SbBaseBlockProps } from '@/services/storyblok/storyblok'
+import { blocksOfType } from '@/services/storyblok/Storyblok.helpers'
 
 type Props = SbBaseBlockProps<{
   title?: string
-  items: Array<SbBlokData>
+  items: Array<AccordionItemBlockProps['blok']>
```
gustaveen, line 10: Great!

apps/store/src/blocks/AccordionBlock.tsx
gustaveen, line 23: nit: As an alternative, you could filter this before the return 
```suggestion
          const accordions = blocksOfType(blok.items, AccordionItemBlock.blockName)
          return (
            accordions.map((nestedBlock) =>
```

gustaveen (commented):
Great stuff Aleks! 🙌 +1 on on Robins comment of the function name

apps/store/src/blocks/AccordionBlock.tsx
alebedev: Memoizing does not help with re-renders here.  And performance cost of filtering an array is trivial unless it's very large and we've seen the cost when profiling

alebedev: fixed

apps/store/src/services/storyblok/blocksOfType.ts
alebedev: fixed

apps/store/src/services/storyblok/Storyblok.helpers.ts
```diff
@@ -0,0 +1,18 @@
+import { SbBlokData } from '@storyblok/react'
+
+export const filterByBlockType = <BlockData extends SbBlokData>(
+  blocks: BlockData[],
+  targetType: string,
+): BlockData[] => {
+  const result: BlockData[] = []
```
simonauner, line 7: Perhaps it makes no difference, but could also be:

```
const result = blocks.filter(block => block.component === targetType)
if (process.env.NODE_ENV !== 'production' && blocks.some(block => block.component !== targetType) {
    throw
}
```

apps/store/src/blocks/AccordionItemBlock.tsx
```diff
@@ -2,12 +2,12 @@ import { getStoryblokApi, Richtext, storyblokEditable } from '@storyblok/react'
 import * as Accordion from '@/components/Accordion/Accordion'
 import { SbBaseBlockProps } from '@/services/storyblok/storyblok'
 
-export type Props = SbBaseBlockProps<{
+export type AccordionItemBlockProps = SbBaseBlockProps<{
```
simonauner, line 5: 😌

simonauner (approved):
Sweet 👍

apps/store/src/services/storyblok/Storyblok.helpers.ts
alebedev: Decided not to do this to avoid extra array pass on each render when developing.  Resulting code is somewhat less expressive but still easy to understand

apps/store/src/blocks/AccordionBlock.tsx
```diff
@@ -1,15 +1,17 @@
...
+import { filterByBlockType } from '@/services/storyblok/Storyblok.helpers'
 
 type Props = SbBaseBlockProps<{
   title?: string
-  items: Array<SbBlokData>
+  items: Array<AccordionItemBlockProps['blok']>
 }>
 
 export const AccordionBlock = ({ blok }: Props) => {
+  const accordions = filterByBlockType(blok.items, AccordionItemBlock.blockName)
```
robinandeer, line 14: nit: rename to `accordionItems`, no?

robinandeer (approved):

PR #432, opened 2022-08-24
Cart page: get or create shop session

robinandeer:

I think it makes sense that we don't fail the page just because there's no shop session.

### DIFF
--- a/apps/store/src/pages/cart/index.tsx
+++ b/apps/store/src/pages/cart/index.tsx
@@ -2,8 +2,10 @@ import type { NextPageWithLayout, GetServerSideProps } from 'next'
 import { CartPage } from '@/components/CartPage/CartPage'
 import { CartPageProps } from '@/components/CartPage/CartPageProps.types'
 import { LayoutWithMenu } from '@/components/LayoutWithMenu/LayoutWithMenu'
+import { getLocale } from '@/lib/l10n/getLocale'
 import { initializeApollo } from '@/services/apollo/client'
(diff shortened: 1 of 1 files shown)


apps/store/src/pages/cart/index.tsx
```diff
@@ -2,19 +2,23 @@ import type { NextPageWithLayout, GetServerSideProps } from 'next'
...
+import { isCountryCode } from '@/utils/isCountryCode'
 
 const NextCartPage: NextPageWithLayout<CartPageProps> = (props) => {
   return <CartPage {...props} />
 }
 
 export const getServerSideProps: GetServerSideProps<CartPageProps> = async (context) => {
   const { req, res } = context
+  const { marketLabel: countryCode } = getLocale(context.locale ?? context.defaultLocale)
+  if (!isCountryCode(countryCode)) return { notFound: true }
```
alebedev, line 17: Do we need to do this for every page?  If so, is there a way to do this check only in one place

alebedev (approved):

apps/store/src/pages/cart/index.tsx
simonauner: If it's needed only for `getShopSessionServerSide` it looks sort of like a "session guard" on server side could be useful 👌

robinandeer:
**Graphite Merge Job**

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/432?utm_source=stack-comment).


Job Reference: 5MjvrvcnL6kdNl4uI6Pd

PR #441, opened 2022-08-26
GRW-1402 TopMenu floating over content

KarnellSchultz:
## Describe your changes

Made the top menu stick to the top of the page and float over the content under it.

floating-image

See video below for more context.

https://user-images.githubusercontent.com/50870173/186854034-f6cb40e6-d109-494a-b294-00f7c133cbd2.mov

Previously the top menu was only shown at the top of the page. When the user scrolled away from the top they were no longer able to see the top menu. 

[GRW-1402]: https://hedvig.atlassian.net/browse/GRW-1402?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/store/src/components/TopMenu/TopMenu.tsx
+++ b/apps/store/src/components/TopMenu/TopMenu.tsx
@@ -91,11 +91,11 @@ export const TopMenu = () => {
             </NavigationPrimaryList>
           </Navigation>
 
-          <DialogPrimitive.DialogClose asChild>
+          <DialogCloseIcon asChild>
             <IconButton>
               <CrossIcon />
             </IconButton>
-          </DialogPrimitive.DialogClose>
+          </DialogCloseIcon>
         </DialogContent>
       </DialogPrimitive.Root>
 
@@ -118,6 +118,8 @@ export const Wrapper = styled.div(({ theme }) => ({
   width: '100%',
   height: MENU_BAR_HEIGHT,
   padding: theme.space[4],
+  position: 'fixed',
+  zIndex: '1000',
 }))
 
 export const StyledDialogOverlay = styled(DialogPrimitive.Overlay)({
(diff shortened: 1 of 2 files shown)


apps/store/src/components/TopMenu/TopMenu.tsx
```diff
@@ -118,6 +118,8 @@ export const Wrapper = styled.div(({ theme }) => ({
   width: '100%',
   height: MENU_BAR_HEIGHT,
   padding: theme.space[4],
+  position: 'fixed',
+  zIndex: '1000',
```
KarnellSchultz, line 122: Wondering if we think we should put more thought into how we're using `z-index`. Looking around the codebase it looks like we're making it us as we go, with `zIndex: '1000'` being the highest I've seen.

Did some googling and found this: https://css-tricks.com/systems-for-z-index/

Let me know what you all think.

simonauner: I like the idea of the SASS map or something like it, so that it's easy to _in one place_ see which z-indices are being used for what, especially if they're named for that level they represent, like "modal", "menu" and so on. But then again things can be locally stacked as well, and then it could become confusing to specify local stacking contexts in a shared file, so in those cases maybe we could still have it in their component files.

simonauner (approved):
Not sure how it works with scrolling and so on, will we use a padding on top of every page where this is present? Perhaps it's already defined in LayoutWithMenu and then we're all good to go I guess 👍  nice

KarnellSchultz:
@simonauner Great point, I did not consider that. At the moment it's not account for, sounds like that's something we'd want to add directly into the layout when the header is present?

apps/store/src/components/TopMenu/TopMenu.tsx
alebedev: I've seen z-indexes added to theme.  May work for us

PR #451, opened 2022-08-29
Add Statistics and InsurableLimits components

simonauner:
## Describe your changes

* Add a `<Statistics>` component that can display `value` and `label`
* Add a `<InsurableLimits>` component that takes in an array of objects with `value` and `label` and displays a grid of `<Statistics>` components

Ticket https://hedvig.atlassian.net/browse/GRW-1361

## Jira issue(s): []

https://user-images.githubusercontent.com/1343979/187166483-0f19802b-a441-4a9a-a15b-54bb6cc28a95.mov

(the reason it looks like it's a bit glitchy when resizing the window is because I added a _very_ long value for the travel insurance to see how it behaves. In edge cases it will be like this:

image

Obviously the user wouldn't see that as they probably do not resize the window.

### DIFF
--- a/apps/store/src/components/InsurableLimits/InsurableLimits.tsx
+++ b/apps/store/src/components/InsurableLimits/InsurableLimits.tsx
@@ -1,0 +1,28 @@
+import styled from '@emotion/styled'
+import { Statistic } from '../Statistic/Statistic'
+
+const Wrapper = styled.div(({ theme }) => ({
+  display: 'grid',
+  gap: theme.space[2],
+  // fit as many columns as possible in the container without the width subceeding 200px
+  gridTemplateColumns: 'repeat(auto-fit, minmax(200px, 1fr))',
+}))
+
+type InsurableLimit = {
+  value: string
+  label: string
+}
+
+type InsurableLimitsProps = {
+  limits: InsurableLimit[]
+}
+
+export const InsurableLimits = ({ limits }: InsurableLimitsProps) => {
+  return (
+    <Wrapper>
+      {limits.map((limit) => (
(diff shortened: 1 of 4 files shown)


robinandeer (approved):
[haseverything](https://app.graphite.dev/macros/HedvigInsurance)

apps/store/src/components/InsurableLimits/InsurableLimits.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+const Wrapper = styled.div(({ theme }) => ({
+  display: 'grid',
+  gap: theme.space[2],
+  // fit as many columns as possible in the container without the width subceeding 200px
+  gridTemplateColumns: 'repeat(auto-fit, minmax(200px, 1fr))',
+}))
+
+type InsurableLimit = {
+  value: string | number
```
alebedev, line 12: Nit: do you think we'll ever have numbers here?  If this component expects formatted string, only string will be given

apps/store/src/components/Statistic/Statistic.tsx
```diff
@@ -0,0 +1,36 @@
+import styled from '@emotion/styled'
+
+export type StatisticProps = {
+  value: string | number
```
alebedev, line 4: same thing regarding value type

alebedev (approved):

apps/store/src/components/InsurableLimits/InsurableLimits.tsx
simonauner: Maybe not, I can narrow it down!

PR #449, opened 2022-08-29
Integrate update start date mutation on cart review page

robinandeer:
## Describe your changes

Remove API route to handle form submission.

Implement hook to handle form submission.

Pass loading state to CartReviewPage.

## Justify why they are needed

Just making the "cart review page" (we should rename it) work with the API.

We should also ask the BE team for a single endpoint to update multiple start dates.

## Checklist before requesting a review

- [ ] I have performed a self-review of my code

PR #448, opened 2022-08-29
Make CartCard a presentational component

robinandeer:
## Describe your changes

Avoid controlling Diaglog in CartCard.

Handle removing cart line item in CartPage component.

Pass cart ID to CartPage component.

Remove ShopSession hook.

## Justify why they are needed

So we don't have to setup one mutation/form handler for every item in the cart.

We also co-locate more of the "smart" things in the CartPage component.

## Checklist before requesting a review

- [ ] I have performed a self-review of my code

PR #447, opened 2022-08-29
Refactor to remove item from cart client side

robinandeer:
## Describe your changes

Remove server-side logic related to removing line item from cart.

Remove useForm hook.

Move transform from API to CartPage props client-side.

Add hook to get current shop session based on cookies.

## Justify why they are needed

Simplify the setup, being able to co-locate more.

I realizing I'm back-tracking a bit but I actually started to like this setup!
If we are using Apollo Client, let's lean into it.

We also get the benefit of using the apollo dev tools.

AND most importantly, this speeds up the UX since there's no serverless function involved.

## Checklist before requesting a review

- [ ] I have performed a self-review of my code

PR #445, opened 2022-08-26
Refactor price calculator API route

robinandeer:
## Describe your changes

Refactor price calculator API route

Decide whether to confirm a price intent in the API route.

Pass form template ID to the API route.

Make ID part of FormTemplate interface.

Remove intent form input.

## Justify why they are needed

It doesn't make much sense to leave up to the client to decide wether to generate a price or not since we can derive it based on the state.

## Checklist before requesting a review

- [ ] I have performed a self-review of my code

PR #453, opened 2022-08-29
Implement API for updating contact details

robinandeer:
## Describe your changes

Refactor CheckoutPage to separate based on payments flow.

Add new to handle form to update contact details.

Update ShopSession query to include checkout information.

Refactor getOrThrowFormValue to not use formidable dependency.

We need to update this implementation to also trigger the start of signing for countries that use the `AFTER_SIGN` payment flow.

### DIFF
--- a/apps/store/src/components/CheckoutPage/CheckoutSignPage.tsx
+++ b/apps/store/src/components/CheckoutPage/CheckoutSignPage.tsx
@@ -1,12 +1,56 @@
-import { Button } from 'ui'
+import styled from '@emotion/styled'
+import { Button, InputField, Space } from 'ui'
+import { CheckoutPageProps } from './CheckoutPage.types'
 import { CheckoutPageLayout } from './CheckoutPageLayout'
 
-export const CheckoutSignPage = () => {
+export const CheckoutSignPage = ({ prefilledData }: CheckoutPageProps) => {
   return (
-    <CheckoutPageLayout>
-      <Button type="submit" fullWidth>
-        Sign with BankID
-      </Button>
+    <CheckoutPageLayout
+      Footer={
+        <Space y={1.5}>
+          <MutedText>
+            By clicking &quot;Sign with BankID&quot; I confirm that I have read and understood the
+            terms and conditions, and that I approve that Hedvig handles my personal information.
+          </MutedText>
+          <Button type="submit" fullWidth>
+            Sign with BankID
+          </Button>
+        </Space>
+      }
+    >
+      <Space y={1}>
+        <InputField
+          label="Personal number"
+          name="personalNumber"
+          required
+          defaultValue={prefilledData.personalNumber ?? undefined}
+        />
+        <InputField
+          label="First name"
+          name="firstName"
+          required
+          defaultValue={prefilledData.firstName ?? undefined}
+        />
+        <InputField
+          label="Last name"
+          name="lastName"
+          required
+          defaultValue={prefilledData.lastName ?? undefined}
+        />
+        <InputField
+          label="Email"
+          name="email"
+          type="email"
+          required
+          defaultValue={prefilledData.email ?? undefined}
+        />
+      </Space>
     </CheckoutPageLayout>
   )
 }
+
+const MutedText = styled.p(({ theme }) => ({
+  fontSize: theme.fontSizes[0],
+  color: theme.colors.gray700,
+  textAlign: 'center',
+}))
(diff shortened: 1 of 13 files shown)


robinandeer:
* main
  * **PR #444** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/444" target="_blank">Graphite</a> 
    * **PR #445** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/445" target="_blank">Graphite</a> 
      * **PR #447** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/447" target="_blank">Graphite</a> 
        * **PR #448** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/448" target="_blank">Graphite</a> 
          * **PR #449** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/449" target="_blank">Graphite</a> 
            * **PR #453** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/453" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/453?utm_source=stack-comment).

apps/store/src/components/CheckoutPage/useHandleSubmitContactDetails.ts
```diff
@@ -0,0 +1,39 @@
...
+  flow: PaymentConnectionFlow
+  onSuccess: () => void
+}
+
+export const useHandleSubmitContactDetails = (params: Params) => {
+  const { checkoutId, flow, onSuccess } = params
+
+  const [updateContactDetails, result] = useContactDetailsUpdateMutation()
+
+  const handleSubmit: FormEventHandler<HTMLFormElement> = async (event) => {
```
alebedev, line 16: We return new `handleSubmit` on every render.  Should we make it stable like `useState -> setState` ?

robinandeer: Hmm I don't see how it will affect the current implementation but I could wrap it in "useCallback" just in case

I guess this is what "useEvent" will solve in the future!

apps/store/src/pages/checkout/index.tsx
```diff
@@ -1,8 +1,58 @@
...
+}
+
+const NextCheckoutPage: NextPage<NextPageProps> = (props) => {
+  const { flow, checkoutId, ...pageProps } = props
+
+  const router = useRouter()
+  const [handleSubmit] = useHandleSubmitContactDetails({
+    checkoutId,
+    onSuccess() {
+      router.push(PageLink.checkoutPaymentAdyen())
```
simonauner, line 24: Will it always be Adyen or is that just for now?

apps/store/src/components/CheckoutPage/CheckoutSignPage.tsx
```diff
@@ -1,12 +1,56 @@
...
 import { CheckoutPageLayout } from './CheckoutPageLayout'
 
-export const CheckoutSignPage = () => {
+export const CheckoutSignPage = ({ prefilledData }: CheckoutPageProps) => {
   return (
-    <CheckoutPageLayout>
-      <Button type="submit" fullWidth>
-        Sign with BankID
-      </Button>
+    <CheckoutPageLayout
```
simonauner, line 8: Not a huge deal, but we have used "Layout" before for `.getLayout`  on Pages. I don't see how to solve this nicely with different footers in `checkout/index.tsx` without more conditional renders, but just as a caution that we now have different ways of using the concept "layout" which might spread.

apps/store/src/pages/checkout/index.tsx
robinandeer: Aha no just for now -- otherwise it should be based on "PaymentConnectionFlow" I suppose.

apps/store/src/components/CheckoutPage/CheckoutSignPage.tsx
robinandeer: Ahaaa yeah I see. Good point.

I'm not trying to solve the same problem here so perhaps I should rename it something else.

simonauner: Well, just food for thought at least. Not a blocker to merge IMO

simonauner (approved):

PR #454, opened 2022-08-30
Add InsurableLimit(s) bloks

simonauner:
## Describe your changes

* Adds `InsurableLimitsBlock` and `InsurableLimitBlock` components
* Refactored `InsurableLimits` component since `InsurableLimitsBlock` loops out `InsurableLimitBlock`

**Question:** 
Should we for clarity's sake keep `InsurableLimits` component? It's pretty meaningless in itself and its styling could just as easily be in the `InsurableLimitsBlock` component. But if we want to use `InsurableLimits` outside of Storyblok's context it could still be useful... What's our stance on this?

## Jira issue(s): [GRW-1366]

image

image

[GRW-1366]: https://hedvig.atlassian.net/browse/GRW-1366?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/store/src/components/InsurableLimits/InsurableLimits.stories.tsx
+++ b/apps/store/src/components/InsurableLimits/InsurableLimits.stories.tsx
@@ -1,5 +1,6 @@
 import { INITIAL_VIEWPORTS } from '@storybook/addon-viewport'
 import { ComponentMeta, ComponentStory } from '@storybook/react'
+import { Statistic } from '@/components/Statistic/Statistic'
 import { InsurableLimits } from './InsurableLimits'
 
 export default {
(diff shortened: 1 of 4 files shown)


apps/store/src/components/InsurableLimits/InsurableLimits.stories.tsx
```diff
@@ -1,5 +1,6 @@
 import { INITIAL_VIEWPORTS } from '@storybook/addon-viewport'
 import { ComponentMeta, ComponentStory } from '@storybook/react'
+import { Statistic } from '../Statistic/Statistic'
```
robinandeer, line 3: nit: update import to `@/components/...`

robinandeer (approved):
I like the updates to `InsurableLimits`!

But I see your point that it doesn't do much. It could be a generic layout component. Perhaps when the design is better defined we can get rid of it but for now I think it's fine to keep.

gustaveen (approved):
Great work! 🙌

I hear you, it can feel a bit overkill. But in general I like the separation of concerns, I think the `AccordionBlock` is a perfect example on how we want to compose componets in blocks. But as stated we can remove it for this component if we see no use for it. :)

simonauner:
Thanks for your input @gustaveen and @robinandeer . Keeping it for now then 👍

PR #455, opened 2022-08-30
Upgrade @testing-library/react to support React 18

robinandeer:
## Describe your changes

Upgrade @testing-library/react to support React 18
https://github.com/testing-library/react-testing-library/releases/tag/v13.0.0

Add missing peer dependency @testing-library/dom.

Remove @testing-library/react-hooks (no longer needed).
Source: https://github.com/testing-library/react-hooks-testing-library

Update test that expects error.

Tests were using the deprecated React render API.

[GRW-1436]: https://hedvig.atlassian.net/browse/GRW-1436?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


robinandeer:
* main
  * **PR #455** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/455" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/455?utm_source=stack-comment).

simonauner (approved):
Nice. I like the new approach of asserting exceptions, that it throws instead 👍

packages/ui/src/lib/media-query.test.ts
```diff
@@ -28,13 +28,23 @@ describe('media-query', () => {
...
-      // Suppress log
-      jest.spyOn(global.console, 'error').mockImplementationOnce(() => {})
+      // prevent writing to stderr during this render
+      const err = console.error
+      console.error = jest.fn()
 
       // act
-      const { result } = renderHook(() => useBreakpoint(breakpoint as Level))
+      let exception: unknown
+      try {
```
simonauner, line 37: I guess one could also use


```suggestion
      expect(() =>  {
          renderHook(() => useBreakpoint(breakpoint as Level))
          }).toThrow(`Unknown breakpoint ${breakpoint}`)
```

https://jestjs.io/docs/expect#tothrowerror

packages/ui/src/lib/media-query.test.ts
```diff
@@ -28,13 +28,23 @@ describe('media-query', () => {
       // align
       const breakpoint = 'smurf'
 
-      // Suppress log
-      jest.spyOn(global.console, 'error').mockImplementationOnce(() => {})
+      // prevent writing to stderr during this render
+      const err = console.error
```
alebedev, line 32: I think original spyOn was better solution - less code and ensures that it rolls back after test (here we need to do it manually and I don't know what would happen if that line would not be reached because of exception)

packages/ui/src/lib/media-query.test.ts
```diff
@@ -28,13 +28,23 @@ describe('media-query', () => {
...
+      let exception: unknown
+      try {
+        renderHook(() => useBreakpoint(breakpoint as Level))
+      } catch (error) {
+        exception = error
+      }
+
       // assert
-      expect(result.error).toEqual(new Error(`Unknown breakpoint ${breakpoint}`))
+      expect(exception).toEqual(new Error(`Unknown breakpoint ${breakpoint}`))
```
alebedev, line 44: I know this code has not changed, but can we use expect.toThrow?

https://jestjs.io/docs/expect#tothrowerror

alebedev (approved):

packages/ui/src/lib/media-query.test.ts
robinandeer: yes, I thought so but it fails when I try to call it that way :/

packages/ui/src/lib/media-query.test.ts
robinandeer: Nope, I tried and jest doesn't catch the error for some reason :/

robinandeer (commented):
@alebedev agree with you, but that still printed the errors to the console -- not sure why so I updated to this solution

PR #456, opened 2022-08-30
Grw-1413 Chore inline blocks

KarnellSchultz:
## Describe your changes

Refactored storyblock components so they no longer use the generic `<StoryblokComponent />`.

This will make it a bit more difficult for components to be placed in unintended locations. For example if a `<HeaderBlock />` is only expected a `<h2 />` then it would no longer be possible to place another component inside of the `<HeaderBlock />`.


KarnellSchultz:
There are a few places where it makes sense to use `<StoryblokComponent />` like Page blocks or blocks that can take a wide variety of blocks.

apps/store/src/blocks/HeroBlock.tsx
```diff
@@ -1,24 +1,26 @@
...
 }>
 
 export const HeroBlock = ({ blok }: HeroBlockProps) => {
   return (
     <HeroSection {...storyblokEditable(blok)} bgImage={blok.background.filename}>
       <div>
-        {blok.content.map((nestedBlock) => (
-          <StoryblokComponent blok={nestedBlock} key={nestedBlock._uid} />
-        ))}
+        {blok.content.map((nestedBlock) => {
```
simonauner, line 17:
```suggestion
        {blok.content.map((nestedBlock) => <HeadingBlock blok={nestedBlock} key={nestedBlock._uid} />)}
```

simonauner (approved):
Nice refactoring 🙌

apps/store/src/blocks/HeroBlock.tsx
```diff
@@ -1,24 +1,26 @@
...
 }>
 
 export const HeroBlock = ({ blok }: HeroBlockProps) => {
   return (
     <HeroSection {...storyblokEditable(blok)} bgImage={blok.background.filename}>
       <div>
-        {blok.content.map((nestedBlock) => (
-          <StoryblokComponent blok={nestedBlock} key={nestedBlock._uid} />
-        ))}
+        {blok.content.map((nestedBlock) => <HeadingBlock blok={nestedBlock} key={nestedBlock._uid} />)}
```
alebedev, line 17: Let's also filter on client side as additional safeguard against incorrect Storyblok config

```
const contentBlocks = filterBlocksByType(blok.content, HeadingBlock.blockName)
```

alebedev:
From failed Vercel deployment you can see that yarn.lock needs updating.  Since you did not change any deps, let's investigate if our main branch is healthy

apps/store/src/blocks/HeroBlock.tsx
KarnellSchultz: totally forgot about that, good call 👍

apps/store/src/blocks/AccordionBlock.tsx
```diff
@@ -2,12 +2,12 @@ import styled from '@emotion/styled'
...
 import { AccordionItemBlock, AccordionItemBlockProps } from '@/blocks/AccordionItemBlock'
 import * as Accordion from '@/components/Accordion/Accordion'
-import { SbBaseBlockProps } from '@/services/storyblok/storyblok'
+import { ExpectedBlockType, SbBaseBlockProps } from '@/services/storyblok/storyblok'
 import { filterByBlockType } from '@/services/storyblok/Storyblok.helpers'
 
 type Props = SbBaseBlockProps<{
   title?: string
-  items: Array<AccordionItemBlockProps['blok']>
+  items: ExpectedBlockType<AccordionItemBlockProps>
```
alebedev, line 10: Nice!

apps/store/src/services/storyblok/storyblok.ts
```diff
@@ -27,6 +27,11 @@ export type SbBaseBlockProps<T> = {
   blok: SbBlokData & T
 }
 
+export type ExpectedBlockType<T> = T extends { blok: SbBlokData }
```
simonauner, line 30: Cool! TS Master Level 1000!

guilhermespopolin, line 30: Have never seen that _conditional types_ feature. You never stop learn :)

guilhermespopolin (approved):
Well done :)

apps/store/src/services/storyblok/storyblok.ts
KarnellSchultz: lolol - perhaps level 4 or 5. One day I'll get to level 1000 https://media.giphy.com/media/l0Exe5NhestV2TfC8/giphy.gif

KarnellSchultz: Yeah it's really one of the coolest things you can do.

apps/store/src/services/storyblok/Storyblok.helpers.ts
```diff
@@ -1,7 +1,7 @@
 import { SbBlokData } from '@storyblok/react'
 
 export const filterByBlockType = <BlockData extends SbBlokData>(
-  blocks: BlockData[],
+  blocks: BlockData[] = [],
```
KarnellSchultz, line 4: Added So this does not throw an error when the argument is optional.

PR #457, opened 2022-08-30
Add jsx-a11y ESLint plugin

simonauner:
## Describe your changes

* Add `jsx-a11y` plugin as was discussed on the tech forum last week
* Fix problems
* Filter out NextJS ESLint options from UI package ESLint config file, to omit this warning:

image

It seems that even though I remove `packages/**` from `rootDir` option the `'next/core-web-vitals'` rules will try to lint the UI package 🤷 

https://nextjs.org/docs/basic-features/eslint#rootdir

I think it's justified to remove NextJS rules from the UI package since it doesn't have a dependency on NextJS. Perhaps there's a prettier way to achieve it?

### DIFF
--- a/apps/onboarding/src/pages/_document.tsx
+++ b/apps/onboarding/src/pages/_document.tsx
@@ -35,6 +35,7 @@ export default class MyDocument extends Document {
               src={
                 process.env.NEXT_PUBLIC_APP_ENV === 'prod' ? gtmProdScript.body : gtmDevScript.body
               }
+              title="gtm"
               height="0"
               width="0"
               style={{ display: 'none', visibility: 'hidden' }}
--- a/apps/onboarding/src/pages/index.tsx
+++ b/apps/onboarding/src/pages/index.tsx
@@ -19,13 +19,9 @@ const HomePage = () => {
     <Wrapper>
       <Heading>Racoon</Heading>
 
-      <Link href={PageLink.forever({ code: '8batky' })}>
-        <a>Forever page</a>
-      </Link>
+      <Link href={PageLink.forever({ code: '8batky' })}>Forever page</Link>
 
-      <Link href={PageLink.landing()}>
-        <a>Landing page</a>
-      </Link>
+      <Link href={PageLink.landing()}>Landing page</Link>
     </Wrapper>
   )
 }
--- a/packages/scripts/.eslintrc.js
+++ b/packages/scripts/.eslintrc.js
@@ -5,8 +5,9 @@ module.exports = {
     'eslint:recommended',
     'plugin:@typescript-eslint/recommended',
     'prettier', // Uses eslint-config-prettier to turn off all rules that are unnecessary or might conflict with Prettier
+    'plugin:jsx-a11y/recommended',
   ],
-  plugins: ['import', 'testing-library', 'jest', '@typescript-eslint'],
+  plugins: ['import', 'testing-library', 'jest', '@typescript-eslint', 'jsx-a11y'],
   ignorePatterns: ['next.config.js'],
   settings: {
     'import/parsers': {
--- a/packages/ui/.eslintrc.js
+++ b/packages/ui/.eslintrc.js
@@ -1,1 +1,2 @@
+// eslint-disable-next-line @typescript-eslint/no-var-requires, no-undef
 module.exports = require('scripts/eslint-preset')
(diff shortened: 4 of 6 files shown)


apps/onboarding/src/pages/_document.tsx
```diff
@@ -35,6 +35,7 @@ export default class MyDocument extends Document {
               src={
                 process.env.NEXT_PUBLIC_APP_ENV === 'prod' ? gtmProdScript.body : gtmDevScript.body
               }
+              title="gtm"
```
simonauner, line 38: Must have titles on iframes

apps/onboarding/src/pages/index.tsx
```diff
@@ -19,13 +19,9 @@ const HomePage = () => {
     <Wrapper>
       <Heading>Racoon</Heading>
 
-      <Link href={PageLink.forever({ code: '8batky' })}>
-        <a>Forever page</a>
-      </Link>
+      <Link href={PageLink.forever({ code: '8batky' })}>Forever page</Link>
```
simonauner, line 22: Not allowed to have `a` without `href`. It doesn't make in sense in this case to have empty `a`:s anyway

packages/ui/.eslintrc.js
```diff
@@ -1 +1,10 @@
-module.exports = require('scripts/eslint-preset')
+// eslint-disable-next-line @typescript-eslint/no-var-requires, no-undef
+const preset = require('scripts/eslint-preset')
```
simonauner, line 2: For some reason `module.exports = require('scripts/eslint-preset')` is totally fine, but splitting them up generates a lot of errors 🙃

packages/ui/src/components/Card/Card.tsx
```diff
@@ -74,6 +74,5 @@ const Img = styled.img({
 })
 
 export const CardMedia = ({ image, height, alt }: CardMediaProps) => {
-  // eslint-disable-next-line @next/next/no-img-element
```
simonauner, line 77: Cannot reference NextJS rules in an app that doesn't have NextJS rules anymore

packages/scripts/.eslintrc.js
```diff
@@ -5,8 +5,9 @@ module.exports = {
     'eslint:recommended',
     'plugin:@typescript-eslint/recommended',
     'prettier', // Uses eslint-config-prettier to turn off all rules that are unnecessary or might conflict with Prettier
+    'plugin:jsx-a11y/recommended',
   ],
-  plugins: ['import', 'testing-library', 'jest', '@typescript-eslint'],
+  plugins: ['import', 'testing-library', 'jest', '@typescript-eslint', 'jsx-a11y'],
```
alebedev, line 10: What if instead of adding next-specific plugins here we create another config file that extends base config with next rules?  Something like `eslint-config.nextjs.js`

Adding extra rules is easier to understand that removing some rules in packages/ui

alebedev (commented):
Looks good, just need to find optimal structure for eslint config

packages/scripts/.eslintrc.js
simonauner: Yep, for sure! I can fix that 👍

simonauner: Opened a new PR instead of modifying this, it got really confusing after a while :D 

https://github.com/HedvigInsurance/racoon/pull/461

apps/onboarding/src/pages/index.tsx
robinandeer: Oh this is actually a next.js thing => `<Link>` doesn't render an `<a>` ref so you need to add it yourself

Super annoying and this will be fixed in an upcoming version but for now we need to add it manually

simonauner: Are you sure? It looks like it renders `<a>` for me when I render the page:
image

simonauner:
Rebased on top of my branch (merged to main) that split out the Next.js ESLint rules, so the UI package's ESLint config no longer filters away Next.js options as before.

I also had to dedupe the `yarn.lock` file since the [builds complained](https://github.com/HedvigInsurance/racoon/runs/8106947949?check_suite_focus=true) about conflicting versions of `jsx-a11y` 🤷 

I executed:

```
yarn dedupe
```
in the project root folder.

apps/onboarding/src/pages/index.tsx
robinandeer: Ahaaaaaa if the child is a string it will add the `<a>` element 🤷‍♂️ (https://stackoverflow.com/questions/67911391/nextjs-link-isnt-rendering-an-anchor-tag)

Otherwise it needs to be added manually (for now).

robinandeer (approved):

PR #458, opened 2022-08-30
Finalize i18n setup (client side)

robinandeer:
## Describe your changes

Finalize i18n setup (client side)

Add i18n-next dependency to store, upgrade for onboarding.

Add download-translations script to store.

Update `ButtonBlock` to use `NextLink` to support locales.

Update `[...slug]` page to generate pages for all locales (temp).

Fetch translations on all pages and return 404 for default locale.

Update types for `getStoryBySlug` since it also can return `undefined`.

I haven't used the translations on any page since I didn't find any use cases unless you get to the cart/checkout - and the API has updated so couldn't test it.

When we decide how to handle locales in CMS we will update how we handle that in `[...slug]`.


robinandeer:
* main
  * **PR #458** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/458" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/458?utm_source=stack-comment).

apps/store/src/pages/[...slug].tsx
```diff
@@ -27,20 +28,27 @@ const NextPage: NextPageWithLayout<StoryblokPageProps> = (props: StoryblokPagePr
...
 
-export const getStaticProps: GetStaticProps<StoryblokPageProps, StoryblokQueryParams> = async ({
-  params,
-  preview,
-}) => {
+export const getStaticProps: GetStaticProps<StoryblokPageProps, StoryblokQueryParams> = async (
+  context,
+) => {
+  const { params, preview, locale } = context
+  if (!locale || locale === 'default') return { notFound: true }
```
Comment (alebedev) [inline apps/store/src/pages/[...slug].tsx:35]: Would be nice to ensure we never get `'default'` and there's always meaningful fallback - would allow us to remove these checks from everywhere

alebedev (approved):

apps/store/src/pages/[...slug].tsx
Comment (robinandeer) [reply apps/store/src/pages/[...slug].tsx:35]: Hmm, we don't have a "default" locale right now

It's a feature we can't really turn off in Next.js tho. There are workarounds (https://nextjs.org/docs/advanced-features/i18n-routing#prefixing-the-default-locale) but I felt like it was safer to start like this and then we can discuss later!

Comment (alebedev) [reply apps/store/src/pages/[...slug].tsx:35]: Agree, let's address this in another MR.  Maybe we can handle default locale at middleware level and remove at that boilerplate?

alebedev (approved):
Looks good overall.  Two thing I think we should address in separate PRs
- localize install in download script
- boilerplate with empty / default locale in every page

PR #459, opened 2022-08-30
Fix print effect on forever page

robinandeer:
## Describe your changes

Update implementation.

With async rendering we can't be sure about the order anymore so we got unpredictable behavior with "setInterval".

I've tested this approach and seems to work fine.

### DIFF
--- a/apps/onboarding/src/components/forever/usePrintCodeEffect.ts
+++ b/apps/onboarding/src/components/forever/usePrintCodeEffect.ts
@@ -8,27 +8,20 @@ type PrintCodeEffectParams = {
 
 export const usePrintCodeEffect = ({ initialCode }: PrintCodeEffectParams) => {
   const [code, setCode] = useState('')
+  const [charIndex, setCharIndex] = useState(0)
 
   useEffect(() => {
-    let charIndex = 0
-    setCode('')
+    if (charIndex >= initialCode.length) return
 
-    const handle = window.setInterval(() => {
-      if (charIndex < initialCode.length) {
-        setCode((codePiece) => codePiece + initialCode[charIndex])
-        charIndex++
-      } else {
-        window.clearInterval(handle)
-      }
-    }, LOAD_CHAR_INTERVAL)
-
-    return () => {
-      window.clearInterval(handle)
+    const tick = () => {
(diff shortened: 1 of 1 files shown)


robinandeer:
* main
  * **PR #459** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/459" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/459?utm_source=stack-comment).

apps/onboarding/src/components/forever/usePrintCodeEffect.ts
```diff
@@ -8,27 +8,21 @@ type PrintCodeEffectParams = {
...
-    }, LOAD_CHAR_INTERVAL)
-
-    return () => {
-      window.clearInterval(handle)
+    if (charIndex >= initialCode.length) return
+
+    const tick = () => {
+      setCode(initialCode.slice(0, charIndex + 1))
+      setCharIndex(charIndex + 1)
+      console.log('ajd')
```
alebedev, line 19: debug log?

apps/onboarding/src/components/forever/usePrintCodeEffect.ts
```diff
@@ -8,27 +8,21 @@ type PrintCodeEffectParams = {
 
 export const usePrintCodeEffect = ({ initialCode }: PrintCodeEffectParams) => {
   const [code, setCode] = useState('')
```
alebedev, line 10: Do we need state for code?  We can calculate it as `initialCode.slice(0, charIndex + 1)` at every render

Less state we have, less opportunity for conflicts

alebedev (approved):

guilhermespopolin (approved):
Nice. Just remove the log as Alexander mentioned and we'd be good to go.

apps/onboarding/src/components/forever/usePrintCodeEffect.ts
robinandeer: @alebedev hmm but what triggers those renders? we need to update state to trigger the next render, right?

PR #462, opened 2022-08-30
GRW-1401 / Add hideMenu option to Page in storyblok

gustaveen:
Co-authored-by: karnell schultz <karnellschultz@gmail.com>

## Describe your changes

Add a Page config group to Page content type with a `hideMenu` checkbox.

https://user-images.githubusercontent.com/6661511/187471774-44ec1343-7a08-4aba-9114-d8e2eed199df.mov

Add possibility to hide menu/header on some pages.
However, haven't seen in design what pages this option is relevant for. So we might need to tweak the functionality. Do we want to hide the whole header (with Hedvig logo and cart icon) or just the menu in the header? 

### DIFF
--- a/apps/store/src/components/LayoutWithMenu/LayoutWithMenu.tsx
+++ b/apps/store/src/components/LayoutWithMenu/LayoutWithMenu.tsx
@@ -16,6 +16,7 @@ type LayoutWithMenuProps = {
 }
 
 export const LayoutWithMenu = ({ children }: LayoutWithMenuProps) => {
+  const story = children.props.story
   const globalStory = children.props.globalStory
   const router = useRouter()
   const handleChangeLocale: SiteFooterProps['onChangeLocale'] = (locale) => {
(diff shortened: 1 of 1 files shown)


KarnellSchultz (approved):
LGTM

simonauner:
I honestly thought that we would use a different Next.js layout for those pages where do not want a menu, or logo. But if it should be totally CMS-driven I think this is a good idea 👍 

In case of designs I can't say either, but I usually find that flows on other sites usually remove all distractions when it comes to checkout and payment so that users are only directed in one direction. Either way it's nice to have the option 👍

gustaveen:
@simonauner I agree, we should use different layouts for page templates that we know won't have a menu. But I suppose it will be usable for general pages on occasions. Just wanted to see so I understood the ticket correctly :) Do you have any more input @robinandeer

alebedev (approved):
Looks good as a study on ways to configure reusable layout elements.  Let's review if/when there's new information on use cases

PR #463, opened 2022-08-30
Fix shop session fetching server-side

robinandeer:
## Describe your changes

Fix shop session fetching server-side

  Pass shopSessionId to provider if we fetch it server side.
  Enable ssr for shop session query.

We get hydration error on pages that server render the header.

Not sure why Apollo require us to set `ssr`-option on the query level AND on the client.

Need to pass shop session ID to the provider since we can't read cookies inside React on the server.

### DIFF
--- a/apps/store/src/pages/_app.tsx
+++ b/apps/store/src/pages/_app.tsx
@@ -33,7 +33,7 @@ const MyApp = ({ Component, pageProps }: AppPropsWithLayout) => {
       <ApolloProvider client={apolloClient}>
         <ThemeProvider>
           <GlobalStyles />
-          <ShopSessionProvider>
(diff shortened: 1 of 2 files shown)


robinandeer:
* main
  * **PR #463** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/463" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/463?utm_source=stack-comment).

simonauner (approved):
I wouldn't say that I can follow along with all changes that are being made to the shop session recently, but I don't see any problems with the logic here 🙃 👍

robinandeer (commented):
@simonauner very good point, I'll try to document and share more now that we are coming to a better understanding how it will work!

PR #464, opened 2022-08-30
Update CartFragment to include cost

robinandeer:
## Describe your changes

Use cost from API on cart pages

We should continue discussing with BE team if amount should be string / number.

And should we decide cross out?

### DIFF
--- a/apps/store/src/graphql/CartFragment.graphql
+++ b/apps/store/src/graphql/CartFragment.graphql
@@ -1,5 +1,15 @@
 fragment CartFragment on Cart {
   id
+  cost {
+    subtotal {
+      amount
+      currencyCode
+    }
+    total {
+      amount
+      currencyCode
+    }
+  }
   lines {
     id
     startDate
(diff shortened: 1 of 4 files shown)


robinandeer:
* main
  * **PR #464** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/464" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/464?utm_source=stack-comment).

simonauner (approved):
Approved, but it hurts to have `parseInt` in our code - IMHO we should just show whatever BE sends us. If we want to show it formatted with certain amount of decimals we should receive it pre-formatted from BE. 

I guess BE knows the locale in the query for price? 
If so, they can format it accordingly.

* `1000` in `sv-SE`: `1 000`
* `1000` in `"en-SE"`: `1,000`

Further, I assume that BE also knows the currency for the market, and then they can also render it with currency:

* `1000` in `sv-SE`: `1 000 kr`
* `1000` in `"en-SE"`: `SEK 1,000`

robinandeer (commented):
Yeah that's very true -- I think that's the reason shopify designed the API to return "money amount" as strings. I'll ask them in Slack!

PR #469, opened 2022-09-01
GRW-1425 / feature / field-level translations

alebedev:
## Draft of field level translations support

See Accordion block title, that's localized in Storyblok UI and displayed in currently selected language
Screenshot 2022-09-01 at 17 19 48

- One step in localization support for CMS

[GRW-1425]: https://hedvig.atlassian.net/browse/GRW-1425?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/store/src/services/storyblok/storyblok.ts
```diff
@@ -128,26 +129,44 @@ export const initStoryblok = () => {
...
 
-export const getProductStory = async (slug: string, preview = false) => {
-  const story = await getStoryBySlug(`/products/${slug}`, preview)
+export const getProductStory = async (slug: string, options: StoryOptions) => {
+  const story = await getStoryBySlug(`/products/${slug}`, options)
   return story as ProductStory
 }
+
+// FIXME: Discuss locale format with team
+const localeToLanguage = (locale: string) => {
```
simonauner, line 165: So this is for `sv-se` to become `sv-SE`? Seems legit. Could also be useful to lowercase the first part just to make sure in case the function is called with `"SV-SE"`?

I guess the main problem for us is that we use `se-en` which is the reversed format of `{LANGUAGE-REGION}` that is the ISO standard. Maybe not in context for this PR though...

alebedev: Good point, locale format is a bit chaotic right now.  I think we should discuss it outside of this PR

simonauner (approved):

PR #471, opened 2022-09-02
Add query to fetch available payment methods

robinandeer:
## Describe your changes

Add query to fetch available payment methods

Fetch payment methods on adyen page.

Remove stuff related to Trustly for now -- we can focus on adding that when we build it for real :)

[GRW-1434]: https://hedvig.atlassian.net/browse/GRW-1434?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


robinandeer:
* main
  * **PR #471** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/471" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/471?utm_source=stack-comment).

apps/store/src/services/adyen/AdyenCheckout.tsx
```diff
@@ -5,20 +5,19 @@ import { useCurrentLocale } from '@/lib/l10n/useCurrentLocale'
...
+  paymentMethods: object
   onSuccess: (paymentConnection: unknown) => void
 }
 
-export const AdyenCheckout = ({ onSuccess }: Props) => {
-  const { locale } = useCurrentLocale()
+export const AdyenCheckout = ({ onSuccess, paymentMethods }: Props) => {
   const paymentContainer = useRef<HTMLDivElement>(null)
-  const paymentMethodConfiguration = useAdyenPaymentMethodConfiguration()
+  const configuration = useAdyenConfiguration()
```
alebedev, line 14: Are we planning to reuse that as a hook?  Otherwise we can have something like this

```
configuration = getAdyenConfiguration({locale, theme})
```

No useMemo overhead and testable without hooks

alebedev (approved):
Looks good

apps/store/src/services/adyen/AdyenCheckout.tsx
robinandeer: Hmm... so you mean we would call this new function inside the `createCheckout` function?

Otherwise, we still need `useMemo` :)

I like doing it this way since 1) there's not much to test here and 2) it hides some complexity from the caller.

alebedev: Yep, inside createCheckout.  If you're only using `configuration` once inside effect, you can construct it there and remove memo

alebedev (approved):

apps/store/src/services/adyen/AdyenCheckout.tsx
robinandeer: Okey, makes sense -- if we don't support switching locale client-side then that works

I'll look at it in a separate PR!

robinandeer:
**Graphite Merge Job**

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/471?utm_source=stack-comment).


Job Reference: D7PCCahe3PVdqM8XBcQY

PR #472, opened 2022-09-02
Fix semantic html structure

gustaveen:
## Describe your changes

- Render proper tags for `header` and `footer`
- Put header at start of document
- Remove nested `main` tags

The structure we should aim to follow:
```
<body>
  <header>
    <nav>…</nav>
  </header>

  <main>
  </main>

  <footer>
  </footer>
</body>
```

### Before
Screenshot 2022-09-02 at 11 18 04

### After
Screenshot 2022-09-02 at 11 28 51

It was a SEO concern from Peter if we would render the header and menu on the bottom of the page. Sorry for adding back `z-index` @robinandeer I feel your pain :D but I think it's nice to keep the low level logical order of the document as well. We have some options to colocate z-index throughout the app, I can make a proposal for it :) 

[GRW-1454]: https://hedvig.atlassian.net/browse/GRW-1454?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/store/src/components/LayoutWithMenu/LayoutWithMenu.tsx
+++ b/apps/store/src/components/LayoutWithMenu/LayoutWithMenu.tsx
@@ -6,7 +6,7 @@ import { FooterBlock } from '@/blocks/FooterBlock'
 import { HeaderBlock } from '@/blocks/TopMenuBlock'
 import { SiteFooterProps } from '@/components/SiteFooter/SiteFooter'
 
-const Wrapper = styled.main({
+const Wrapper = styled.div({
   minHeight: '100vh',
   display: 'flex',
   flexDirection: 'column',
@@ -28,8 +28,8 @@ export const LayoutWithMenu = ({ children }: LayoutWithMenuProps) => {
 
   return (
     <Wrapper>
-      {children}
       {(!story || !story.content.hideMenu) && <HeaderBlock blok={globalStory.content} />}
+      {children}
       {(!story || !story.content.hideMenu) && (
         <FooterBlock blok={globalStory.content} onChangeLocale={handleChangeLocale} />
       )}
--- a/apps/store/src/blocks/FooterBlock.tsx
+++ b/apps/store/src/blocks/FooterBlock.tsx
@@ -90,7 +90,7 @@ export const FooterBlock = ({ blok }: FooterBlockProps) => {
   const footerSections = filterByBlockType(blok.footerSections, FooterSection.blockName)
 
   return (
-    <Wrapper y={2}>
+    <Wrapper as="footer" y={2}>
       <Accordion.Root type="multiple">
         {footerSections.map((nestedBlok) => (
           <FooterSection key={nestedBlok._uid} blok={nestedBlok} />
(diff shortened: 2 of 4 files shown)


gustaveen:
* main
  * **PR #472** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/472" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/472?utm_source=stack-comment).

apps/store/src/components/LayoutWithMenu/LayoutWithMenu.tsx
```diff
@@ -6,7 +6,7 @@ import { FooterBlock } from '@/blocks/FooterBlock'
 import { HeaderBlock } from '@/blocks/TopMenuBlock'
 import { SiteFooterProps } from '@/components/SiteFooter/SiteFooter'
 
-const Wrapper = styled.main({
+const Wrapper = styled.div({
```
gustaveen, line 9: nit: I think we're good with removing the wrapper :)

robinandeer (approved):
Nice work @gustaveen!

I'm not convinced that web crawlers would be confused by us putting elements in this order.

However, I understand the case for accessibility and assistive devices where a logical order naturally is important!

[ditto](https://app.graphite.dev/macros/HedvigInsurance)

robinandeer (commented):
...oh and another comment -- we should move z-index values to the theme and avoid using explicit numbers but aliases like top, bottom, middle etc.

simonauner:
> ...oh and another comment -- we should move z-index values to the theme and avoid using explicit numbers but aliases like top, bottom, middle etc.

bottom
middle
top
topper
toppiest
new-top
new-topper

:trollface:

gustaveen:
> I'm not convinced that web crawlers would be confused by us putting elements in this order.
Me neither 😄 @robinandeer

gustaveen:
For sure @robinandeer @simonauner we should co-locate them! 🙌

And yeah, that logic can be iffy haha. As your example Simon it's hard to know ahead of time how many z-index use cases you need so you usually make some wiggle room in between numbers:

```
zindex-dropdown:       1000;
zindex-sticky:         1020;
zindex-fixed:          1030;
zindex-modal-backdrop: 1040;
zindex-modal:          1050;
zindex-popover:        1060;
zindex-tooltip:        1070;
```

But is that want you want to avoid Robin?

I thought this approach might be something even thought it feels a bit overkill (https://www.smashingmagazine.com/2021/02/css-z-index-large-projects/) :

```
// Utils
const base = 0;
const above = 1; // use this for all values above the base
const below = -1; // and this for all values below the base

// Page Layout
export const zLayoutNavigation = above + base;
export const zLayoutFooter = above + base;
export const zLayoutModal = above + zLayoutNavigation;
export const zLayoutPopUpAd = above + zLayoutModal;

// NavMenu
export const zNavMenuBackdrop = below + base;
export const zNavMenuPopover = above + base;
export const zNavMenuToggle = above + zNavMenuPopover;
```

simonauner:
I'm just kidding. Of course there are ways of organizing it. I like the idea of having them centrally somewhere. Obviously we can cram a lot of things in (10,000, right?), so let's just be conservative and observe when magic numbers appear. I don't think (hope) that we will use a lot of z-indexes, our designers seem to stay away from floating elements mostly 🙏

PR #473, opened 2022-09-02
GRW-1456 / Chore / Remove pages/index

alebedev:
## Remove pages/index, use catch-all route instead

Use https://nextjs.org/docs/routing/dynamic-routes#optional-catch-all-routes to handle index page as well as any slugged page

- Less code
- Required to implement per-market folders without code duplication

[GRW-1456]: https://hedvig.atlassian.net/browse/GRW-1456?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/store/src/pages/[[...slug]].tsx
```diff
@@ -51,7 +52,10 @@ export const getStaticProps: GetStaticProps<StoryblokPageProps, StoryblokQueryPa
 export const getStaticPaths: GetStaticPaths = async ({ locales = [] }) => {
   const links = await getAllLinks()
 
-  const paths: Path[] = []
+  const paths: Path[] = [
```
Comment (simonauner) [inline apps/store/src/pages/[[...slug]].tsx:55]: Perhaps link to the documentation in the PR description as it explains why we do this?

simonauner (approved):
Nice find 🕵️

apps/store/src/pages/[[...slug]].tsx
Comment (alebedev) [reply apps/store/src/pages/[[...slug]].tsx:53]: There's no direct link that I could use, but I'll explain why it's needed in comments

PR #475, opened 2022-09-05
Support language change from footer selector

alebedev:

Allows to test localization

### DIFF
--- a/apps/store/src/blocks/FooterBlock.tsx
+++ b/apps/store/src/blocks/FooterBlock.tsx
@@ -1,5 +1,6 @@
 import { storyblokEditable } from '@storyblok/react'
 import Link from 'next/link'
+import { useRouter } from 'next/router'
 import { ChangeEvent, useRef } from 'react'
 import { Space } from 'ui'
 import * as Accordion from '@/components/Accordion/Accordion'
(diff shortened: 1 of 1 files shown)


apps/store/src/blocks/FooterBlock.tsx
```diff
@@ -79,9 +80,10 @@ export const FooterBlock = ({ blok }: FooterBlockProps) => {
     formRef.current?.dispatchEvent(new Event('submit', { cancelable: true, bubbles: true }))
   }
 
-  // todo: connect to locale / market
-  const onChangeLocale = (value: Locale | undefined) => console.log('Location changed:', value)
-
+  const router = useRouter()
+  const onChangeLocale = (locale: Locale | undefined) => {
+    router.push(router.asPath, undefined, { locale })
```
simonauner, line 85: Should we forward any query parameters? They will be removed with this, I guess

alebedev: Good catch, I'll handle this

alebedev: Already works.  If I change language on `/sv-se/about/why-hedvig?q=1` I'm redirected to `sv-ev/about/why-hedvig?q=1` -- all query parameters are retained

simonauner: Magic 🧙

simonauner (approved):

PR #477, opened 2022-09-05
Extract market data from locales

alebedev:
- We need to operate with markets directly in many places
- Don't repeat common things like country code, currency, etc once per supported language

[GRW-1460]: https://hedvig.atlassian.net/browse/GRW-1460?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/store/src/blocks/FooterBlock.tsx
```diff
@@ -54,17 +56,21 @@ type FooterBlockProps = SbBaseBlockProps<{
...
     name: TEMP_TRANSLATIONS[`MARKET_LABEL_${market}`],
     value: market,
   }))
 
-  const languageList = MARKET_MAP[currentMarket].map((language) => ({
-    name: TEMP_TRANSLATIONS[`LANGUAGE_LABEL_${language}`],
-    value: language,
-  }))
+  const languageList = currentMarket.locales.map((locale) => {
+    const language = locale.split('-')[0]
```
robinandeer, line 68: Instead of manipulating string, why not create a complete data structure for the `locales`-array?

Like:

```
{ language: 'sv', countryCode: 'se', locale: 'sv-se' }
```

apps/store/src/lib/l10n/locales.ts
```diff
@@ -127,15 +104,11 @@ export const locales: Record<Locale, LocaleData> = {
       backendFormatExample: 'ÅÅÅÅ-MM-DD',
     },
     currencyLocale: Locale.DaDk,
-    countryCode: CountryCode.Dk,
   },
   'en-dk': {
     path: Locale.DaDk,
     locale: Locale.EnDk,
```
robinandeer, line 110: This looks incorrect 🤔

apps/store/src/lib/l10n/markets.ts
```diff
@@ -0,0 +1,60 @@
...
+    id: MarketLabel.DK,
+    apiMarket: Market.Denmark,
+    adtractionScriptSrc: 'https://cdn.adt387.com/jsTag?ap=1589794294',
+    countryCode: CountryCode.Dk,
+    defaultLocale: Locale.DaDk,
+    locales: [Locale.DaDk, Locale.EnDk],
+  },
+}
+
+export const getMarketByLocale = (locale: string | undefined | null): MarketData => {
```
robinandeer, line 42: Why is this not typed with `Locale` instead of `string`?

apps/store/src/lib/l10n/markets.ts
```diff
@@ -0,0 +1,60 @@
...
+  const marketId = (locale?.split('-')[1] ?? '').toUpperCase()
```
robinandeer, line 43: Again, string manipulation -- feel like we can come up with a data structure to link markets (or countries) to languages to avoid having to do this in multiple places.

apps/store/src/lib/l10n/markets.ts
```diff
@@ -0,0 +1,60 @@
...
+    throw new Error(`Failed to find market by locale=${locale}`)
+  }
+  return marketData
+}
+
+export const findMarketLocale = (market: string, language: string): Locale | undefined => {
+  const marketData = markets[market as MarketLabel]
+  if (!marketData) {
+    throw new Error(`Failed to find market id=${market}`)
+  }
```
robinandeer, line 55: Is it better perhaps to just return null?

apps/store/src/lib/l10n/useCurrentMarket.ts
```diff
@@ -0,0 +1,9 @@
+import { useRouter } from 'next/router'
+import { FALLBACK_LOCALE } from './locales'
+import { getMarketByLocale } from './markets'
+
+export const useCurrentMarket = () => {
+  const router = useRouter()
+  const locale = router.locale === 'default' ? FALLBACK_LOCALE : router.locale || FALLBACK_LOCALE
```
robinandeer, line 7: Let's discuss if we should continue with a "fallback locale" -- it might confuse us down the line - but OK for now!

robinandeer (approved):
Good struff!

I guess you plan to rename "market" to "country" as we discussed in another PR?

apps/store/src/blocks/FooterBlock.tsx
alebedev: Good catch, only one split remains in the code now - when we find country from locale

apps/store/src/lib/l10n/markets.ts
alebedev: This is programmer error, I'd rather blow up for ease of debugging

robinandeer: Yeah true, maybe we can rename the function to `getMarketLocale`

I would also type the input strictly as `market: MarketLabel` -- otherwise we loose type safety, right?

apps/store/src/lib/l10n/markets.ts
alebedev: This is old part.  From what I see, I assume we wanted safe defaults when using it, same as getLocale

apps/store/src/lib/l10n/markets.ts
alebedev: Fixed

apps/store/src/lib/l10n/locales.ts
alebedev: Fixed

apps/store/src/lib/l10n/markets.ts
alebedev: Typical usage looks like this
```
  const { locale, req, res, params: { product: slug } = {}, preview } = context

  if (typeof slug !== 'string') return { notFound: true }

  const { countryCode } = getMarketByLocale(locale)
```

Do you see a way to take Locale without moving runtime checks to call sites?  Since we're taking URL parts, we _may_ get empty or non-locale string due to misconfiguration - I think it makes sense to protect against it

apps/store/src/lib/l10n/useCurrentMarket.ts
alebedev: Agree.  This seems related to country selector logic

PR #480, opened 2022-09-06
Grw-1447  🌍  Country selector page 🌍

KarnellSchultz:
## Describe your changes

- ~Add redirect from `/` to the new `country-selector` page.~
- create `country-selector` page / component
- create `country-selector` storyblock page

The design is 'made up' since there is no actual design for this page. In the future we'd want to get input from designers on this page.

Screenshot 2022-09-05 at 22 07 10

👆`http://localhost:8040/country-selector`

At the moment all traffic to the root of the application gets a `404`. Laying the groundwork for having the ability to select a country when we do not know where a user is located.


apps/store/src/components/CountrySelectorPage/CountrySelectorPage.tsx
```diff
@@ -0,0 +1,58 @@
...
+const CountryOptionContainer = styled.div({
+  display: 'grid',
+  gap: 5,
+})
+
+const SupportedMarkets = {
+  Sweden: 'Sweden',
+  Norway: 'Norway',
+  Denmark: 'Denmark',
+} as const
```
robinandeer, line 27: Let's avoid defining these things in multiple places -- we should get this the same way as e.g. in footer block

See @alebedev PR :)

apps/store/src/components/CountrySelectorPage/CountrySelectorPage.tsx
```diff
@@ -0,0 +1,58 @@
...
+  textAlign: 'center',
+  background: theme.colors.gray100,
+}))
+
+const CountryOptionContainer = styled.div({
+  display: 'grid',
+  gap: 5,
+})
+
+const SupportedMarkets = {
```
robinandeer, line 23: Why not use an enum?

apps/store/src/pages/country-selector/index.tsx
```diff
@@ -0,0 +1,39 @@
...
+  const [story, globalStory] = await Promise.all([
+    getStoryBySlug(slug, { locale: getLocation as string, preview }),
+    getGlobalStory({ locale: getLocation as string, preview }),
+  ])
+
+  if (story === undefined) throw new Error('Unable to find page story')
+
+  return {
+    props: {
+      ...(await serverSideTranslations('sv-se')),
```
robinandeer, line 26: This page will never be in swedish tho 😅 the whole point is we don't know so better lock it to english, right?

apps/store/src/pages/country-selector/index.tsx
```diff
@@ -0,0 +1,39 @@
...
+export const getStaticProps: GetStaticProps<StoryblokPageProps> = async ({
+  preview,
+  locale = Locale.SvSe,
+}) => {
+  // Can be removed when we handle default locals
+  const getLocation = locale === 'default' ? Locale.SvSe : locale
+
+  const slug = 'home'
+  const [story, globalStory] = await Promise.all([
+    getStoryBySlug(slug, { locale: getLocation as string, preview }),
```
robinandeer, line 18: avoid type assertions ;) but okey for now

apps/store/src/components/CountrySelectorPage/CountrySelectorPage.tsx
KarnellSchultz: Sounds good. I'll wait for that PR to merge before going down that path. 👍

apps/store/src/components/CountrySelectorPage/CountrySelectorPage.tsx
```diff
@@ -0,0 +1,58 @@
...
+type SupportedMarketsKeys = keyof typeof SupportedMarkets
+
+const getLocaleFromMarket = (market: SupportedMarketsKeys) => MarketsMap.get(market)
+
+export const CountrySelectorPage = (props: CountrySelectorPageProps) => {
+  return (
+    <Container {...props}>
+      <h1>Select Country</h1>
+      <CountryOptionContainer>
+        {ObjectKeys(SupportedMarkets).map((market) => {
```
simonauner, line 44: Seems like a complicated way of doing just `Object.keys(SupportedMarkets)`? TS would give typing if the original object was typed, right?

simonauner (commented):
Nice to get a default page instead of just a 404 👍

apps/store/src/components/CountrySelectorPage/CountrySelectorPage.tsx
KarnellSchultz: It had something to do with how I was typing the functions below. But going to swap it out after using the new markets.

apps/store/src/components/CountrySelectorPage/CountrySelectorPage.tsx
KarnellSchultz: Perhaps there is a better way of getting that typing info from `Object.keys` but I'm not sure how else to do it. 

The keys here would only get typed as `string` when using `Object.keys`.

simonauner: But that's because the type of the keys in `SupportedMarkets` are strings?

KarnellSchultz: I think since a key can be a few different types TS has no way of knowing what's the type of the key. 

Put a little [TS playgroud](https://www.typescriptlang.org/play?#code/PTAEHUFMBsGMHsC2lQBd5oBYoCoE8AHSAZVgCcBLA1UABWgEM8BzM+AVwDsATAGiwoBnUENANQAd0gAjQRVSQAUCEmYKsTKGYUAbpGF4OY0BoadYKdJMoL+gzAzIoz3UNEiPOofEVKVqAHSKymAAmkYI7NCuqGqcANag8ABmIjQUXrFOKBJMggBcISGgoAC0oACCbvCwDKgU8JkY7p7ehCTkVDQS2E6gnPCxGcwmZqDSTgzxxWWVoASMFmgYkAAeRJTInN3ymj4d-jSCeNsMq-wuoPaOltigAKoASgAywhK7SbGQZIIz5VWCFzSeCrZagNYbChbHaxUDcCjJZLfSDbExIAgUdxkUBIursJzCFJtXydajBBCcQQ0MwAUVWDEQC0gADVHBQGNJ3KAALygABEAAkYNAMOB4GRonzFBTBPB3AERcwABS0+mM9ysygc9wASmCKhwzQ8ZC8iHFzmB7BoXzcZmY7AYzEg-Fg0HUiQ58D0Ii8fLpDKZgj5SWxfPADlQAHJhAA5SASPlBFQAeS+ZHegmdWkgR1QjgUrmkeFATjNOmGWH0KAQiGhwkuNok4uiIgMHGxCyYrA4PCC5MaVNAxHYBAI4oLAFlHPEc8JeQBvRQlErEKTcFH5UCR1eQdecSO8JfLmPi3J4TeRk-ppgHo8lAAiKMQ04vj84z7I8Vvy9AAEZN3y2DQCKUoAL7BMm0gAFaQLAqABDOeCCEqw6juOu5Tp+s46gEz4EEqSofjOqA6jyAB8oCEdOOY6nqkEwXBCGQEhKEjmOZCTtRqCCDheEEURNHkZRqHsZxWHcQA2gJqAALq0f2lI0PRsGoAA0sxc6gAAPPR4KrAoPDCI8sHitwWlUpQnDMPwXDxAMEicGRZFKvA0GbvROqbkqiFEh5EkyUJi4-k4qD4l4ymMYhyGuVBpEMMI3nMb50E6v5ijgTKNCIHgmHEepSE8qAEVqRprFoRxGFcTxuEMPh0lCdJeqZaA2W5Tm+WCBUggAJI8GshXFR1ZWiZV4nVXx9XchRInodwbWSdJclAA) together to explain what I'm saying a bit more. 


but perhaps it's overkill and not needed? Let me know what you think.

gustaveen:
GJ Karnell! 🙌
I'm not sure we should have redirect for this page, it should rather be on index (hedvig.com) itself. I know @alebedev removed the index page recently https://github.com/HedvigInsurance/racoon/pull/473 but I think it would be better for this use case :)

apps/store/src/components/CountrySelectorPage/CountrySelectorPage.tsx
simonauner: I see you are totally right! That's weird, haven't seen that behaviour before ⭐ TIL!

KarnellSchultz:
Changed the scope 🔭 of this ticket a bit. Removed the redirect from `/` to `/country-selector`. 

Sounds like we want to change how we handle the `index` page going forward. Perhaps we want to bring it back and remove the catch all routes [([[..slug]])](https://nextjs.org/docs/routing/dynamic-routes). 

Either way it seems like it'd be better to make those changes after a decision is made.

robinandeer (approved):

PR #486, opened 2022-09-07
Chore/upgrade build tools

alebedev:
## Update build-related tools

- No breaking changes for us in any of the major upgrades
- 2 false positives from typescript update, had to add type casting

- Maintenance day activity


packages/ui/src/components/Card/SelectableCardGroup.tsx
```diff
@@ -19,7 +19,7 @@ const recursiveMap = (children: React.ReactNode, fn: (child: React.ReactNode) =>
     }
 
     if (child.props.children) {
-      child = React.cloneElement(child, {
+      child = React.cloneElement(child as ReactElement, {
```
robinandeer, line 22: Is there a point in being defensive here and using the `React.isValidElement(child)` type guard instead of type casting?

Source: https://stackoverflow.com/a/42261933

robinandeer (approved):

packages/ui/src/components/Card/SelectableCardGroup.tsx
alebedev: This allows us to avoid cloning text nodes and fragments, I see no harm here

Overall, cloning and manipulating children is not very obvious way to do stuff.  I might've written SelectableCardGroup differently if we didn't have it already

robinandeer: @alebedev hmm no I don't mean that -- I mean instead of doing `as ReactElement`. Because we lose type safety with type casting, right?

alebedev: We have React.isValidElement above.  Looks like Typescript control flow analysis broke here.  I'll add a comment

PR #489, opened 2022-09-07
Improve workaround for child element type

alebedev:

Document why we had to change code in SelectableCardGroup.  Use ts-expect-error to make it self-cleaning


robinandeer (commented):
Hmm yeah this function makes me very confused, feel free to refactor or if you think you can make it more sensible 😅

Especially we seem to override the `child` variable many time over... perhaps we can use different variable names?

alebedev:
> Hmm yeah this function makes me very confused, feel free to refactor or if you think you can make it more sensible 😅
> 
> Especially we seem to override the `child` variable many time over... perhaps we can use different variable names?

I'd rather not touch it now to avoid making affected area larger.  Will keep it in mind for the future

simonauner:
This is my pet peeve, I like forwarding props to child components :D 

But this was a PoC component that I don't think is used anywhere? I created it mainly to harmonize all the different card components that were almost identical, but it's adapted by anyone.

guilhermespopolin (approved):

PR #496, opened 2022-09-07
GRW-1470 / Feature / Unify locale format to xx-XX

alebedev:
## GRW-1470 / Feature / Unify locale format to xx-XX

Remove non-standard locale variants

Less fields, less normalization, easier to use external APIs

[GRW-1470]: https://hedvig.atlassian.net/browse/GRW-1470?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/store/src/lib/l10n/countries.ts
+++ b/apps/store/src/lib/l10n/countries.ts
@@ -1,4 +1,4 @@
-import { getLocale, normalizeLocale } from './locales'
+import { getLocaleOrFallback, normalizeLocale } from './locales'
 import { CountryCode, Language, Locale, CountryLabel } from './types'
 
 export type CountryData = {
@@ -53,7 +53,7 @@ export const getCountryLocale = (country: CountryLabel, language: Language): Loc
     throw new Error(`Failed to find country id=${country}`)
   }
   return (
-    countryData.locales.find((locale) => getLocale(locale).language === language) ||
+    countryData.locales.find((locale) => getLocaleOrFallback(locale).language === language) ||
     countryData.defaultLocale
   )
 }
--- a/apps/store/src/lib/l10n/countries.ts
+++ b/apps/store/src/lib/l10n/countries.ts
@@ -1,4 +1,4 @@
-import { getLocale } from './locales'
+import { getLocale, normalizeLocale } from './locales'
 import { CountryCode, Language, Locale, CountryLabel } from './types'
 
 export type CountryData = {
@@ -40,7 +40,7 @@ const localeCountries = Object.fromEntries(
 ) as Record<Locale, CountryLabel>
 
 export const getCountryByLocale = (locale: string): CountryData => {
-  const countryData = countries[localeCountries[locale as Locale]]
+  const countryData = countries[localeCountries[normalizeLocale(locale) as Locale]]
   if (!countryData) {
     throw new Error(`Failed to find country by locale=${locale}`)
   }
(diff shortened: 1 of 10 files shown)


apps/store/src/lib/l10n/locales.ts
```diff
@@ -51,6 +51,14 @@ export enum LocaleField {
   Language = 'language',
 }
 
+export const normalizeLocale = (locale: string | undefined): string | undefined => {
```
guilhermespopolin, line 54: Initially I thought about changing `normalizeLocale` so it returns `Locale`. With that we could replace `locales[normalizeLocale(locale) as Locale]` to `locales[normalizeLocale(locale)]`. But that change would trigger so many others I guess it's not worth it 😅

guilhermespopolin (approved):

apps/store/src/lib/l10n/locales.ts
```diff
@@ -65,6 +51,14 @@ export enum LocaleField {
...
+  if (!locale?.includes('-')) {
+    return locale
```
gustaveen, line 56: When would this case be true? 🤔

alebedev: 'default'

apps/store/src/lib/l10n/countries.ts
```diff
@@ -40,7 +40,7 @@ const localeCountries = Object.fromEntries(
 ) as Record<Locale, CountryLabel>
 
 export const getCountryByLocale = (locale: string): CountryData => {
-  const countryData = countries[localeCountries[locale as Locale]]
+  const countryData = countries[localeCountries[normalizeLocale(locale) as Locale]]
```
simonauner, line 43: Should this be 
```suggestion
  const countryData = countries[localeCountries[getLocale(locale)]]
```
then we don't need to cast it here?

apps/store/src/lib/l10n/types.ts
```diff
@@ -2,12 +2,12 @@ export { CountryCode } from '@/services/apollo/generated'
 
 /** An enum representing explicitly endorsed Locales supported by our system. */
 export enum Locale {
-  SvSe = 'sv-se',
-  EnSe = 'en-se',
-  NbNo = 'nb-no',
-  EnNo = 'en-no',
-  DaDk = 'da-dk',
-  EnDk = 'en-dk',
+  SvSe = 'sv-SE',
```
simonauner, line 5: gif

apps/store/src/lib/l10n/useCurrentLocale.ts
```diff
@@ -1,11 +1,11 @@
...
+import { FALLBACK_LOCALE, locales, normalizeLocale } from './locales'
 import { Locale } from './types'
 
 export const useCurrentLocale = () => {
   const router = useRouter()
 
   const locale = router.locale === 'default' ? FALLBACK_LOCALE : router.locale || FALLBACK_LOCALE
 
-  return locales[locale as Locale]
+  return locales[normalizeLocale(locale) as Locale]
```
simonauner, line 10: `getLocale(locale)` with the same motivation as above?

apps/store/src/lib/l10n/countries.ts
alebedev: This may fall back to default locale which is something I'd prefer to avoid here

simonauner: Then I would prefer that we have `getLocaleWithFallback`/`getLocaleOrFallback` and `getLocale` 🤔

apps/store/src/lib/l10n/locales.ts
alebedev: In regular code you just call `getLocale(locale)`, direct locales lookup is only needed if you want special logic when you miss which is few places

PR #507, opened 2022-09-08
Chore/eslint ternary rules

alebedev:
Disallow some forms of ternary abuse and make code easier to read

### DIFF
--- a/apps/onboarding/src/components/LandingPage/BaseCard.tsx
+++ b/apps/onboarding/src/components/LandingPage/BaseCard.tsx
@@ -54,26 +54,20 @@ const Wrapper = styled.button(({ theme }) => ({
 }))
 
 export const SelectableCardWrapper = styled(Wrapper)<SelectableCardWrapperProps>(
-  ({ theme, isCheckable, selected, enableHover }) => {
-    let boxShadow = 'initial'
-    if (selected) {
-      boxShadow = `0 0 0 1px ${theme.colors.black}, 0px 2px 2px rgba(0, 0, 0, 0.1)`
-    }
-    let hoverBoxShadow = 'initial'
-    if (enableHover) {
-      hoverBoxShadow = `0 0 0 1px ${theme.colors.gray600}, 0px 2px 2px rgba(0, 0, 0, 0.1)`
-    }
-    return {
-      cursor: isCheckable ? 'pointer' : 'initial',
-      boxShadow,
-      '&:hover:not([disabled]), &:focus-visible': {
-        boxShadow: hoverBoxShadow,
-      },
-      '&:disabled': {
-        opacity: 0.5,
-      },
-    }
-  },
+  ({ theme, isCheckable, selected, enableHover }) => ({
+    cursor: isCheckable ? 'pointer' : 'initial',
+    boxShadow: selected
+      ? `0 0 0 1px ${theme.colors.black}, 0px 2px 2px rgba(0, 0, 0, 0.1)`
+      : 'initial',
+    '&:hover:not([disabled]), &:focus-visible': {
+      boxShadow: enableHover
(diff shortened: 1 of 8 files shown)


packages/scripts/.eslintrc.js
```diff
@@ -18,6 +18,9 @@ module.exports = {
     },
   },
   rules: {
+    'no-unneeded-ternary': 'error',
+    'no-nested-ternary': 'error',
+    'multiline-ternary': ['error', 'never'],
```
simonauner, line 23: I approve of `'no-nested-ternary'` since it can be really hard to understand what's going on with nested ternaries but the third I don't think makes the code easier to read 🤔 (which I assume is what we want to achieve)

Now we instead get `if` statements and `let`s that really do not increase readability.

I get that this is a matter of _taste_ which is hard to argue over, but I would anyway _argue_ that the following lines aren't hard to understand even though they're on three lines. 

```
 IS_HOUSE_INSURANCE_ENABLED
                    ? 'LANDING_PAGE_MULTI_MAIN_COVERAGE_SUBHEADING'
                    : 'LANDING_PAGE_SUBHEADING'
```
This reads in my mind:

> translate to, If `IS_HOUSE_INSURANCE_ENABLED` is truthy, use first value, otherwise second. 


It's even easier to read when split on three lines (something that Prettier does for us).
Old code:

The new code reads (in my mind)

> Declare `subheading` as `t('LANDING_PAGE_SUBHEADING')`, oh wait, there's an `if` statement here. Oh, it changes `subheading` if `IS_HOUSE_INSURANCE_ENABLED` is truthy. Will `subheading` be changed somewhere else in the code since it's a `let` 🤔 

```
  let subheading = t('LANDING_PAGE_SUBHEADING')
  if (IS_HOUSE_INSURANCE_ENABLED) {
    subheading = t('LANDING_PAGE_MULTI_MAIN_COVERAGE_SUBHEADING')
  }
```

I'm also not a fan of the React-craze that everything needs to be `const`s and ternary expressions to the death, but sometimes they make the code _more_ readable for me. 

My 2 cents 🙃 Take it or leave it, hehe

gustaveen: I agree with Simon, I do think that piece of code is easier to read before 😄

apps/onboarding/src/components/LandingPage/BaseCard.tsx
```diff
@@ -54,20 +54,26 @@ const Wrapper = styled.button(({ theme }) => ({
...
+    if (selected) {
+      boxShadow = `0 0 0 1px ${theme.colors.black}, 0px 2px 2px rgba(0, 0, 0, 0.1)`
+    }
+    let hoverBoxShadow = 'initial'
+    if (enableHover) {
+      hoverBoxShadow = `0 0 0 1px ${theme.colors.gray600}, 0px 2px 2px rgba(0, 0, 0, 0.1)`
+    }
+    return {
+      cursor: isCheckable ? 'pointer' : 'initial',
+      boxShadow,
```
gustaveen, line 68: I get the point of this, the ternary is long. But for me it's quicker to comprehend if don't have to look up what the value of `boxShadow` is above, compared to reading it inline attached to the css rule itself :) as it was before:
```js
boxShadow: selected
      ? `0 0 0 1px ${theme.colors.black}, 0px 2px 2px rgba(0, 0, 0, 0.1)`
      : 'initial',
```

packages/scripts/.eslintrc.js
alebedev: I agree, let's allow multiline ternaries.  Some of the changes I'll revert, ones I think are to the better I will keep and we can have another round of review after this

simonauner (approved):
I think seeing `let`s is just something to get used to, still hurts my eyes but it's better than before 😄 👍

PR #511, opened 2022-09-08
Create a product page context

robinandeer:
## Describe your changes

Create a product page context
Remove `getBlockContext` functionality.
Subscribe to block context in product page blocks.

This will be much easier to maintain than the previous solution.

This way we can keep everything to do with a block in one place.


robinandeer:
* main
  * **PR #511** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/511" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/511?utm_source=stack-comment).

apps/store/src/components/ProductPage/ProductPageContext.tsx
```diff
@@ -0,0 +1,37 @@
+import { createContext, PropsWithChildren, useContext, useState } from 'react'
+import { ProductVariant } from '@/services/apollo/generated'
+import { ProductPageProps } from './ProductPage.types'
+
+type ProductPageContextData = ProductPageProps & {
+  selectedVariant: null | ProductVariant
+  selectedVariantUpdate: (variant: ProductVariant) => void
```
alebedev, line 7: onVariantSelect?

apps/store/src/components/ProductPage/ProductPageContext.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+type Props = PropsWithChildren<Omit<ProductPageContextData, 'selectedVariantUpdate'>>
+
+export const ProductPageContextProvider = (props: Props) => {
+  const { children, selectedVariant: initialSelectedVariant, ...rest } = props
+  const [selectedVariant, setSelectedVariant] = useState(initialSelectedVariant)
+
+  return (
+    <ProductPageContext.Provider
+      value={{
```
alebedev, line 20: Can this provider re-render with new value if we change something at higher level?  If so, we may need to use memo for value

apps/store/src/components/ProductPage/ProductPageContext.tsx
```diff
@@ -0,0 +1,37 @@
...
+  selectedVariant: null | ProductVariant
+  selectedVariantUpdate: (variant: ProductVariant) => void
+}
+
+const ProductPageContext = createContext<ProductPageContextData | null>(null)
+
+type Props = PropsWithChildren<Omit<ProductPageContextData, 'selectedVariantUpdate'>>
+
+export const ProductPageContextProvider = (props: Props) => {
+  const { children, selectedVariant: initialSelectedVariant, ...rest } = props
```
simonauner, line 15: Could be destructured in the function call
```
export const ProductPageContextProvider = ({ children, selectedVariant: initialSelectedVariant, ...rest }: Props) => {
```

apps/store/src/components/ProductPage/ProductPage.tsx
```diff
@@ -1,37 +1,13 @@
...
-            priceIntent,
-          }),
-        })),
-      },
-    }
-  }, [storyblokStory, priceIntent, priceFormTemplate, shopSession])
-
-  return <StoryblokComponent blok={story.content} />
+  return (
+    <ProductPageContextProvider {...props} story={story} selectedVariant={null}>
```
simonauner, line 9: I like the idea of using `initialSelectedVariant` in `ProductPageContextProvider`, perhaps it would be even more clear if that was the name of the attribute already here? Or should it be controllable from the outside? If it can only set the initial value it might be misinterpreted if you think you can always change it from here. But if it's called `initialSelectedVariant` there's less risk of misinterpretation.

simonauner (approved):
I like it! Just some comments, nothing major. But I share Alex's question regarding the re-render - because I don't know 🙃

apps/store/src/components/ProductPage/ProductPageContext.tsx
robinandeer: hmm, no I think that would be incorrect -- this is not a callback, it's the method you use to update the selected product variant :)

apps/store/src/components/ProductPage/ProductPageContext.tsx
robinandeer: ahh you mean since we create a new object here

that would technically only be a problem if someone puts the whole provider context in a `useEffect`, right? Unlikely but I can add `useMemo`!

apps/store/src/components/ProductPage/ProductPage.tsx
robinandeer: Ahh very good point - I will rename this prop

apps/store/src/components/ProductPage/ProductPageContext.tsx
robinandeer: hehe I was just trying to avoid too long lines

what do we think of that?

apps/store/src/components/ProductPage/ProductPageContext.tsx
alebedev: Let me try to reproduce the problem first, I feed we need more concrete data here

apps/store/src/components/ProductPage/ProductPageContext.tsx
alebedev: I see no problem with long argument destructurings.  If lines feel too long perhaps we should tweak prettier printWidth (https://prettier.io/docs/en/rationale.html#print-width)

apps/store/src/components/ProductPage/ProductPageContext.tsx
alebedev: What about `setSelectedVariant(value)`?

robinandeer:
**Graphite Merge Job**

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/511?utm_source=stack-comment).


Job Reference: sSnyAHeEh6GY05lQjjkm

PR #512, opened 2022-09-09
GRW-1471 / Feature / Map country default locale to Storyblok default language

alebedev:
## Describe your changes

Map Storyblok default locale to country-specific default locale
Assume Storyblok uses 2-char language codes for other languages

Implement default language logic for localization experiment

[GRW-1471]: https://hedvig.atlassian.net/browse/GRW-1471?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/store/src/pages/api/preview.ts
+++ b/apps/store/src/pages/api/preview.ts
@@ -20,12 +20,11 @@ const preview = async (req: NextApiRequest, res: NextApiResponse) => {
 
   const [countryLabel, ...slugFragments] = story.full_slug.split('/')
 
-  // Convert default => English if applicable
-  let locale = req.query._storyblok_lang
-  if (locale === 'default') {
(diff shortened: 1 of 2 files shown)


robinandeer (approved):
Hmm yeah let's try this out but I want to raise some concerns here:

Relying on implicit logic like this risks getting us in strange situations that are hard to predict. And harder to understand code that needs additional comments.

alebedev:
> Hmm yeah let's try this out but I want to raise some concerns here:
> 
> Relying on implicit logic like this risks getting us in strange situations that are hard to predict. And harder to understand code that needs additional comments.

I agree, let's improve this some time between figuring out how we localize and going live

PR #514, opened 2022-09-09
Fix endless loop on console.error

alebedev:
## Describe your changes

Don't mirror datadog events to console unless caused by explicit logging

Fix endless loop caused by console.error

We probably don't want to log network errors as well, browser does this for us


robinandeer (approved):
Thanks for fixing this!

[clean](https://app.graphite.dev/macros/HedvigInsurance)

guilhermespopolin (approved):

PR #515, opened 2022-09-09
GRW-1471 / Fix / Locale-specific URLs

alebedev:
## GRW-1481 / Fix / Locale-specific URLs

Fix link generation logic
- Use resolve_links=url to ensure we get fresh links from storyblok, cached_url is unsafe
- Handle $lang slug prefix for non-default languages

Ensure we always get correct URLs in menu

[GRW-1471]: https://hedvig.atlassian.net/browse/GRW-1471?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/store/src/blocks/ButtonBlock.tsx
+++ b/apps/store/src/blocks/ButtonBlock.tsx
@@ -3,6 +3,7 @@ import { storyblokEditable } from '@storyblok/react'
 import NextLink from 'next/link'
 import { ButtonVariant, LinkButton } from 'ui'
 import { LinkField, SbBaseBlockProps } from '@/services/storyblok/storyblok'
+import { useStroryblokLinkURL } from '@/utils/useStroryblokLinkURL'
 
 export type ButtonBlockProps = SbBaseBlockProps<{
   text: string
@@ -11,9 +12,10 @@ export type ButtonBlockProps = SbBaseBlockProps<{
 }>
 
 export const ButtonBlock = ({ blok }: ButtonBlockProps) => {
+  const href = useStroryblokLinkURL(blok.link)
   return (
     <Wrapper {...storyblokEditable(blok)}>
-      <NextLink href={blok.link.cached_url} passHref>
+      <NextLink href={href} passHref>
         <LinkButton variant={blok.variant} color="dark" size="lg">
           {blok.text}
         </LinkButton>
--- a/apps/store/src/services/storyblok/Storyblok.helpers.ts
+++ b/apps/store/src/services/storyblok/Storyblok.helpers.ts
@@ -1,4 +1,6 @@
 import { SbBlokData } from '@storyblok/react'
+import { getCountryByLocale } from '@/lib/l10n/countries'
+import { LocaleData } from '@/lib/l10n/locales'
 import { LinkField } from './storyblok'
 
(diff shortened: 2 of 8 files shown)


robinandeer (commented):
Hmmm, maybe I don't understand but Next.js has built-in support for locales in the Router/link so we don't need to specify it for client-side navigation.

Next.js will make sure that the user stays in the same locale when linking.

alebedev:
> Hmmm, maybe I don't understand but Next.js has built-in support for locales in the Router/link so we don't need to specify it for client-side navigation.
> 
> Next.js will make sure that the user stays in the same locale when linking.

Right, we don't need to append locale explicitly.  Removed this

We still need to transform Storyblok slugs into valid URLs though

apps/store/src/services/storyblok/Storyblok.helpers.ts
```diff
@@ -27,6 +29,23 @@ export const checkBlockType = <BlockData extends SbBlokData>(
   else return null
 }
 
-export const getLinkFieldURL = (link: LinkField) => {
-  return link.cached_url === 'home' ? '/' : `/${link.cached_url}`
+export const getLinkFieldURL = (link: LinkField, locale: LocaleData) => {
```
robinandeer, line 32: passing in "locale" here every time seems tedious -- why not make it a hook and call `useCurrentLocale` inside?

apps/store/src/services/storyblok/Storyblok.helpers.ts
```diff
@@ -27,6 +29,23 @@ export const checkBlockType = <BlockData extends SbBlokData>(
...
+    fragments.shift()
+  }
+
+  // SE/page => page
+  fragments.shift()
+
+  // home => '', products/home => unchanged
+  if (fragments.length === 2 && link.story.slug === 'home') {
+    fragments.pop()
+  }
```
robinandeer, line 48: Why not treat it as an edge case?

```
if (link.story.slug === 'home') return '/'
```

robinandeer (commented):
Still not sold on the benefits of using a default locale as many different languages 😅 I think it makes this code hard to understands

have a look at my comments, but otherwise I guess we can test this out :)

apps/store/src/services/storyblok/Storyblok.helpers.ts
alebedev: No longer relevant with Storyblok-side handling of folder root

apps/store/src/services/storyblok/Storyblok.helpers.ts
alebedev: Fixed

apps/store/src/blocks/ButtonBlock.tsx
```diff
@@ -11,9 +12,10 @@ export type ButtonBlockProps = SbBaseBlockProps<{
 }>
 
 export const ButtonBlock = ({ blok }: ButtonBlockProps) => {
+  const href = useStroryblokLinkURL(blok.link)
```
robinandeer, line 15: Nice!

robinandeer (approved):
Great stuff :D

PR #516, opened 2022-09-09
Feature / Integrate why-did-you-render

alebedev:
## Describe your changes

Integrate https://github.com/welldone-software/why-did-you-render for detecting inefficient rendering

Let's try it out then fine-tune based on impressions

Easier to detect and fix unneeded renders

### DIFF
--- a/apps/store/next.config.js
+++ b/apps/store/next.config.js
@@ -1,3 +1,4 @@
+const path = require('path')
 const { i18n } = require('./next-i18next.config')
 
 /** @type {import('next').NextConfig} */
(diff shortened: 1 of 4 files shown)


apps/store/src/utils/why-did-you-render.ts
```diff
@@ -0,0 +1,11 @@
+import React from 'react'
```
robinandeer, line 1: nit: rename this file to use camelCase like our conventions say

robinandeer (approved):
Sure, I guess we can try this

Always worried when it comes to customizing webpack since that's something we wanted to move away from :/

alebedev:
I didn't manage to make it work consistently, so closing for now.  I'll probably experiment more at some point in the future

PR #521, opened 2022-09-13
Fix / Store static paths 

alebedev:
## Describe your changes
- Extract locale from slug, let nextjs handle it
- Return /home => / normalization, lost it accidentaly in recent routing changes

Fix 404 for all top-level routes

### DIFF
--- a/apps/store/src/pages/[[...slug]].tsx
+++ b/apps/store/src/pages/[[...slug]].tsx
@@ -13,10 +13,11 @@ import {
   StoryblokQueryParams,
 } from '@/services/storyblok/storyblok'
 
-type Path = {
+type RoutingPath = {
   params: {
     slug: string[]
   }
+  locale?: string
 }
 
 const NextPage: NextPageWithLayout<StoryblokPageProps> = (props: StoryblokPageProps) => {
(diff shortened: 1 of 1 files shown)


apps/store/src/pages/[[...slug]].tsx
```diff
@@ -54,18 +55,23 @@ export const getStaticProps: GetStaticProps<StoryblokPageProps, StoryblokQueryPa
 
 export const getStaticPaths: GetStaticPaths = async () => {
   const links = await getAllLinks()
-  const paths: Path[] = []
+  const paths: RoutingPath[] = []
   Object.values(links)
     .filter((link) => !link.is_folder)
     .filter((link) => !link.slug.includes('/products/') && !link.slug.endsWith('/global'))
```
Comment (robinandeer) [inline apps/store/src/pages/[[...slug]].tsx:61]: Wonder if the storyblok API has some filters that we can use to exclude these links?

I also think we should aim to move these things into the Storyblok service layer :)

robinandeer (approved):
Let's start with this and iterate :)

apps/store/src/pages/[[...slug]].tsx
Comment (alebedev) [reply apps/store/src/pages/[[...slug]].tsx:61]: Good points, I'll explore them in separate PR(s)

PR #524, opened 2022-09-14
Add Storybook Composition

gustaveen:
## Describe your changes
Took another stab at co-locating our Storybooks. The problem has been importing stories across packages like this:
```js
stories: ['../src/**/*.stories.@(ts|tsx)', '../../../apps/store/src/**/*.stories.@(ts|tsx)'],
```
Since we use path aliases in `tsconfig` that wouldn't resolve. Have tried all of the following approaches but can't get it to work since it needs to resolve aliases from multiple packages 🤠 
https://ittutorialpoint.com/how-to-resolve-aliases-in-storybook/

So I resorted to using [Storybook Composition](https://storybook.js.org/docs/react/sharing/storybook-composition#gatsby-focus-wrapper) where can reference local or published Storybooks. So you still need to have both Storybooks running but can view them together.

So let's see how that can plays along with Chromatic. Since Chromatic only allows you to have one Storybook per repository we can't publish two storybooks from the same repository. https://www.chromatic.com/docs/monorepos

Ideal solution IMO would still be to do like Chromatic suggest:
```
module.exports = {
  stories: ["../project-1/**/*.stories.js", "../project-2/**/*.stories.js"],
}
```
But let's see how this would work :)

- View all stories in one Storybook
- Publish a single Storybook with `ui` and `store` stories

https://user-images.githubusercontent.com/6661511/190161995-76de906e-ac34-4a01-88a5-c076375e3a6c.mov


gustaveen:
* main
  * **PR #524** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/524" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/524?utm_source=stack-comment).

alebedev:
Oddly, when I go to Chromatic deployment I only see Storybook from store.  Does it need to be updated?

alebedev (approved):
Looks good, works locally.  Up to you if deployment should be updated now or in separate PR

gustaveen:
> Looks good, works locally. Up to you if deployment should be updated now or in separate PR

Thanks! Yes I'll dabble with that in a new PR 🙌

PR #536, opened 2022-09-16
GRW-1492 / Fix / Refactor page link collection

alebedev:
## Describe your changes

Move page type filtering to service layer.  We may need to reuse it for promise catalog in the future

Separate Storyblok mapping from nextjs


gustaveen (approved):
👍

PR #542, opened 2022-09-16
Fix / rollback msw update

alebedev:
## Describe your changes

Rollback MSW, was breaking local development

Prevent renovate from updating MSW, we should update manually in the future

Replace import path in mocks/index to relative, safer to use


KarnellSchultz (approved):
🔙

PR #550, opened 2022-09-19
chore / Add robot.txt file > noindex forever pages

KarnellSchultz:
## Describe your changes

- Add robots.txt file

We don't want the forever codes to appear in google search results 🚫

[GRW-824]: https://hedvig.atlassian.net/browse/GRW-824?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/store/public/robots.txt
+++ b/apps/store/public/robots.txt
@@ -1,0 +1,7 @@
+# Block all crawlers for /forever
+User-agent: *
+Disallow: /forever/
+
+# Allow all crawlers
+User-agent: *
+Allow: /


alebedev (approved):

apps/store/public/robots.txt
```diff
@@ -0,0 +1,7 @@
+# Block all crawlers for /forever
+User-agent: *
+Disallow: /forever/
```
gustaveen, line 3: I asked Peter if we could solve it with `disallow` but that would equal `noindex, nofollow` but we want to have `follow` :) And `Noindex` is not allowed in `robots.txt` anymore. So we would need to solve this with a meta tag on that page instead:

```html
<meta name="robots" content="noindex, follow">
```

PR #552, opened 2022-09-19
Revert "chore / Add robot.txt file > noindex forever pages"

KarnellSchultz:
Reverts HedvigInsurance/racoon#550

we need to address this issue via a `meta` tag.


alebedev:
Do you plan to add separate PR for meta?  Maybe using one for both would be easier

PR #559, opened 2022-09-20
GRW-1466 / Setup GTM coverage store

amandaflod:
## Describe your changes

Setup GTM in the store project. 

Have added the new env variable in vercel.

[GRW-1466]: https://hedvig.atlassian.net/browse/GRW-1466?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/store/src/lib/gtm.tsx
+++ b/apps/store/src/lib/gtm.tsx
@@ -1,0 +1,86 @@
+/**
+ * Based on https://github.com/vercel/next.js/tree/canary/examples/with-google-tag-manager
+ */
+
+import { useRouter } from 'next/router'
+import Script from 'next/script'
+import { useEffect } from 'react'
+
+export const GTM_ID = process.env.NEXT_PUBLIC_GOOGLE_TAG_MANAGER_ID
+
+type GTMUserProperties = {
+  market: string
+  environment: 'development' | 'production' | 'test'
+}
+
+type GTMPageData = {
+  page: string
+  title: string
+}
+
+type DataLayerObject = {
+  event?: string
+  userProperties?: GTMUserProperties
+  offerData?: Record<string, string>
+  eventData?: Record<string, string>
+  pageData?: GTMPageData
+}
+
+declare global {
+  interface Window {
+    dataLayer: DataLayerObject[]
+  }
+}
+
+export const pushToGTMDataLayer = (obj: DataLayerObject) => {
+  if (!window.dataLayer) window.dataLayer = []
+  window.dataLayer.push(obj)
+}
+
+export const pageview = (url: string) => {
+  pushToGTMDataLayer({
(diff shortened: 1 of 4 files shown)


simonauner (approved):
Nice 🙌

apps/store/src/lib/gtm.tsx
```diff
@@ -0,0 +1,86 @@
...
+  offerData?: Record<string, string>
+  eventData?: Record<string, string>
+  pageData?: GTMPageData
+}
+
+declare global {
+  interface Window {
+    dataLayer: DataLayerObject[]
+  }
+}
```
robinandeer, line 33: Not sure where the best place to put this but maybe centralize it in `src/index.d.ts`?

apps/store/src/lib/gtm.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+export const pushToGTMDataLayer = (obj: DataLayerObject) => {
+  if (!window.dataLayer) window.dataLayer = []
```
robinandeer, line 36: Why do we need to do this? 🧐

Does it need to be exported from this module?

robinandeer (commented):
Great stuff! Although let's move this to the `services` folder since this is a third party service :)  

@gustaveen did we have a need to support multiple environments in our setup?

gustaveen:
Awesome! 🎉

apps/store/src/lib/gtm.tsx
amandaflod: I'm following the setup from onboarding. See this comment for the rationale https://github.com/HedvigInsurance/racoon/commit/38dbb73d552a590f652bbbe537d4d8ed4b0399f8

The function is exported since we'll be using it for tracking later on, but we can remove the export meanwhile if you prefer :)

apps/store/src/lib/gtm.tsx
amandaflod: Thank you! Fixing in a separate PR.

apps/store/src/lib/gtm.tsx
simonauner: To add a bit more context: we had a but where we fired an event when a React component is rendered, and that occurred before GTM was loaded, and then we lost that event. Can't recall which one it was exactly, but this is standard way of handling events for GTM if it hasn't been loaded when we start pushing events to the data layer.

gustaveen:
> @gustaveen did we have a need to support multiple environments in our setup?

Yes, we should add that as well. 👍 But was thinking we can do it in a separate PR, let's add a ticket for it

PR #561, opened 2022-09-20
Fix gtm setup comments

amandaflod:
## Describe your changes

Fixing comments from https://github.com/HedvigInsurance/racoon/pull/559


apps/store/src/lib/gtm.tsx
```diff
@@ -26,13 +26,7 @@ type DataLayerObject = {
...
 
-declare global {
-  interface Window {
-    dataLayer: DataLayerObject[]
-  }
-}
-
-export const pushToGTMDataLayer = (obj: DataLayerObject) => {
+const pushToGTMDataLayer = (obj: DataLayerObject) => {
   if (!window.dataLayer) window.dataLayer = []
```
robinandeer, line 30: nit: add a comment that we need this in case we send event before GTM is loaded

robinandeer (approved):
Looks good!

alebedev (approved):

apps/store/src/lib/gtm.tsx
amandaflod: will do!

PR #565, opened 2022-09-20
GRW-1509/feat/ Middleware geolocation redirects 🌐

KarnellSchultz:
## Describe your changes

Created middleware to redirect users based on geolocation.

### DIFF
--- a/apps/store/src/middleware.ts
+++ b/apps/store/src/middleware.ts
@@ -1,0 +1,27 @@
+import { NextResponse } from 'next/server'
+import type { NextRequest } from 'next/server'
+import { countries } from '@/lib/l10n/countries'
+
+export const config = {
+  matcher: '/',
+}
+
+export function middleware(req: NextRequest) {
+  const country = (req.geo && req.geo.country) || 'SE'
+
+  switch (country) {
+    case countries.NO.id:
+      req.nextUrl.pathname = countries.NO.defaultLocale
+      break
+    case countries.DK.id:
+      req.nextUrl.pathname = countries.DK.defaultLocale
+      break
+    case countries.SE.id:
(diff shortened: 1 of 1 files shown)


apps/store/src/middleware.ts
```diff
@@ -0,0 +1,27 @@
+import { NextResponse } from 'next/server'
+import type { NextRequest } from 'next/server'
+import { countries } from '@/lib/l10n/countries'
+
+export const config = {
+  matcher: '/',
+}
+
+export function middleware(req: NextRequest) {
+  const country = (req.geo && req.geo.country) || 'SE'
```
alebedev, line 10: Will never reach default since we provide 'SE' here.  Debug code?

alebedev (approved):

apps/store/src/middleware.ts
KarnellSchultz: I had this in there in an attempt to be extra safe about making sure we get country information from `geo`. 

But I see that we handled that in the switch statement as well so it's not needed. Will remove 👍

apps/store/src/middleware.ts
```diff
@@ -0,0 +1,27 @@
+import { NextResponse } from 'next/server'
+import type { NextRequest } from 'next/server'
+import { countries } from '@/lib/l10n/countries'
+
+export const config = {
+  matcher: '/',
+}
+
+export function middleware(req: NextRequest) {
+  const country = req.geo && req.geo.country
```
gustaveen, line 10: Neat that with get geo from Next 🙏

KarnellSchultz: They really make it easy for you :-)

PR #575, opened 2022-09-22
GRW-1510 / improvements to middleware routing

KarnellSchultz:
## Describe your changes

- Add logging on the root middleware (this is for our testing now, _could_ be removed later)
- Changed the routing to use lowercase routing for pretty urls 🍒
    - the url preview url is going from `.../sv-SE/` to `.../sv-se/`
    - Screenshot 2022-09-22 at 10 03 04

- Tighten up the types being used by for localisation
    - Screenshot 2022-09-22 at 10 02 26

- Creates nicer urls which are lowercase which is inline with how we'd like to show our urls. 
- The types changes prevent users of some of the localisation functions so that users do not use the funcs in a way that would be unexpected. 

[GRW-1510]: https://hedvig.atlassian.net/browse/GRW-1510?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/store/src/components/CountrySelectorPage/CountrySelectorPage.tsx
+++ b/apps/store/src/components/CountrySelectorPage/CountrySelectorPage.tsx
@@ -4,8 +4,7 @@ import { LinkButton } from 'ui'
 import { countries } from '@/lib/l10n/countries'
 import { routingLocale, TEMP_TRANSLATIONS } from '@/lib/l10n/locales'
 import { PageLink } from '@/lib/PageLink'
-
-type CountrySelectorPageProps = any
+import { StoryblokPageProps } from '@/services/storyblok/storyblok'
 
 const Container = styled.div({
   height: 400,
@@ -19,7 +18,7 @@ const CountryOptionContainer = styled.div({
   gap: 5,
 })
 
-export const CountrySelectorPage = (props: CountrySelectorPageProps) => {
+export const CountrySelectorPage = (props: StoryblokPageProps) => {
   return (
     <Container {...props}>
       <h1>Select Country</h1>
--- a/apps/store/src/components/CountrySelectorPage/CountrySelectorPage.tsx
+++ b/apps/store/src/components/CountrySelectorPage/CountrySelectorPage.tsx
@@ -2,7 +2,7 @@ import styled from '@emotion/styled'
 import Link from 'next/link'
 import { LinkButton } from 'ui'
 import { countries } from '@/lib/l10n/countries'
-import { TEMP_TRANSLATIONS } from '@/lib/l10n/locales'
+import { routingLocale, TEMP_TRANSLATIONS } from '@/lib/l10n/locales'
 import { PageLink } from '@/lib/PageLink'
 
 type CountrySelectorPageProps = any
@@ -28,7 +28,7 @@ export const CountrySelectorPage = (props: CountrySelectorPageProps) => {
           <Link
             key={country}
             href={PageLink.store({
-              locale: countries[country as keyof typeof countries].defaultLocale,
+              locale: routingLocale(countries[country as keyof typeof countries].defaultLocale),
             })}
             passHref
           >
--- a/apps/store/src/lib/l10n/locales.ts
+++ b/apps/store/src/lib/l10n/locales.ts
@@ -1,4 +1,4 @@
-import { Language, Locale } from './types'
+import { Language, Locale, LocaleValues, RoutingLocale } from './types'
 
 export const FALLBACK_LOCALE: Locale = Locale.EnSe
 
@@ -7,7 +7,7 @@ export type LocaleData = {
   language: Language
 }
 
-export const locales: Record<Locale, LocaleData> = {
+export const locales: Record<LocaleValues, LocaleData> = {
   [Locale.SvSe]: {
     locale: Locale.SvSe,
     language: Language.Sv,
@@ -60,7 +60,7 @@ export const normalizeLocale = (locale: string | undefined): string | undefined 
 }
 
 // We use en-SE ISO format for settings but downcase it for routing to get nicer URLs
-export const routingLocale = (locale: string) => locale.toLowerCase()
+export const routingLocale = (locale: LocaleValues) => locale.toLowerCase() as RoutingLocale
 
 // TODO: Make fallback market-specific
 export const getLocaleOrFallback = (locale: Locale | string | undefined): LocaleData => {
--- a/apps/store/src/lib/l10n/locales.ts
+++ b/apps/store/src/lib/l10n/locales.ts
@@ -1,13 +1,13 @@
 import { Language, Locale, LocaleValue, RoutingLocale } from './types'
 
-export const FALLBACK_LOCALE: Locale = Locale.EnSe
+export const FALLBACK_LOCALE = Locale.EnSe
 
 export type LocaleData = {
-  locale: Locale
+  locale: LocaleValue
   language: Language
 }
 
-export const locales: Record<LocaleValue, LocaleData> = {
+export const locales = {
   [Locale.SvSe]: {
     locale: Locale.SvSe,
     language: Language.Sv,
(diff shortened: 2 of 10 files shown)


apps/store/src/lib/l10n/types.ts
```diff
@@ -10,6 +10,18 @@ export enum Locale {
...
+  SvSe: 'sv-SE',
+  EnSe: 'en-SE',
+  NbNo: 'nb-NO',
+  EnNo: 'en-NO',
+  DaDk: 'da-DK',
+  EnDk: 'en-DK',
+} as const
+
+export type LocaleValues = typeof LocaleObject[keyof typeof LocaleObject]
+export type RoutingLocale = Lowercase<Locale>
```
simonauner, line 23: 😍

apps/store/src/lib/l10n/locales.ts
```diff
@@ -60,7 +60,7 @@ export const normalizeLocale = (locale: string | undefined): string | undefined
 }
 
 // We use en-SE ISO format for settings but downcase it for routing to get nicer URLs
-export const routingLocale = (locale: string) => locale.toLowerCase()
+export const routingLocale = (locale: LocaleValues) => locale.toLowerCase() as RoutingLocale
```
simonauner, line 63: I'm no TS pro, but could this be or will TS complain?

```suggestion
export const routingLocale = (locale: LocaleValues):RoutingLocale => locale.toLowerCase()
```

simonauner (approved):
gif

Yazzzz! More types! Type it harder!

apps/store/src/lib/l10n/types.ts
```diff
@@ -10,6 +10,18 @@ export enum Locale {
...
+export const LocaleObject = {
+  SvSe: 'sv-SE',
+  EnSe: 'en-SE',
+  NbNo: 'nb-NO',
+  EnNo: 'en-NO',
+  DaDk: 'da-DK',
+  EnDk: 'en-DK',
+} as const
+
+export type LocaleValues = typeof LocaleObject[keyof typeof LocaleObject]
```
robinandeer, line 22: nit: rename to `LocaleValue` since it represents a single locale at a time

apps/store/src/components/CountrySelectorPage/CountrySelectorPage.tsx
```diff
@@ -28,7 +28,7 @@ export const CountrySelectorPage = (props: CountrySelectorPageProps) => {
           <Link
             key={country}
             href={PageLink.store({
-              locale: countries[country as keyof typeof countries].defaultLocale,
+              locale: routingLocale(countries[country as keyof typeof countries].defaultLocale),
```
robinandeer, line 31: nit: do we have to assert the type here?

Isn't `countries` already of the correct type?

robinandeer (approved):
Looks good! Something to think about for the future is that we might want to redirect any URL that doesn't contain a locale but let's start like this :)

apps/store/src/lib/l10n/locales.ts
KarnellSchultz: Yeahhh feels a lil dirty to be casting that type but TS complains without it since `toLowerCase()` only returns a `string`. So going in we have out clean and specific type. but coming out we only have a string :feelsbad:

apps/store/src/lib/l10n/types.ts
KarnellSchultz: Will do!

apps/store/src/lib/l10n/types.ts
```diff
@@ -10,6 +10,18 @@ export enum Locale {
   EnDk = 'en-DK',
 }
 
+export const LocaleObject = {
```
guilhermespopolin, line 13: Maybe we could replace `Locale` enum with this object format so we don't end up with two different types that represents the same thing. Based on the [docs](https://www.typescriptlang.org/docs/handbook/enums.html#objects-vs-enums:~:text=The%20biggest%20argument%20in%20favour%20of%20this%20format%20over%20TypeScript%E2%80%99s%20enum%20is%20that%20it%20keeps%20your%20codebase%20aligned%20with%20the%20state%20of%20JavaScript%2C%20and%20when/if%20enums%20are%20added%20to%20JavaScript%20then%20you%20can%20move%20to%20the%20additional%20syntax.), it seems a good approach. Or at least it would be nice to be able to build a type of the union of `enum Locale` values, but after a quick search that doesn't seem feasible 😢

guilhermespopolin (approved):

apps/store/src/components/CountrySelectorPage/CountrySelectorPage.tsx
```diff
@@ -2,7 +2,7 @@ import styled from '@emotion/styled'
 import Link from 'next/link'
 import { LinkButton } from 'ui'
 import { countries } from '@/lib/l10n/countries'
-import { TEMP_TRANSLATIONS } from '@/lib/l10n/locales'
+import { routingLocale, TEMP_TRANSLATIONS } from '@/lib/l10n/locales'
 import { PageLink } from '@/lib/PageLink'
 
 type CountrySelectorPageProps = any
```
alebedev, line 8: Nit: Since we're touching it now, can we get rid of any here?

apps/store/src/components/CountrySelectorPage/CountrySelectorPage.tsx
KarnellSchultz: Unfortunately, I think we do need to either cast it or type it to use it as in index here. 

This came up on a previous [PR](https://github.com/HedvigInsurance/racoon/pull/480#discussion_r963313281) where @simonauner  and I chatted about this same thing.

Here is a [TS Playground example](https://www.typescriptlang.org/play?ssl=1&ssc=1&pln=3&pc=1#code/FAYw9gdgzgLgBAZQK4AcVgE4wKYBMCyAhhgNbYxRwC8cA3sHI4gO57YQBccA5Aq7u24AaBkwBymZoQCeXbhIxTpw0YwAi7ALbEScjRG2kVTOAEYuAIgAW2ADa2wF4AF9gwAPIAjAFbYQMADoyaSgACmQ0TBwCHXIoAEoA7RRQ0MMyGHjqAD44NNjM+I8fP0DgsIj0LDwiUjjE5NT08iyqXPDUKujajKgAbWaYAF14otBIWDgvX38AaWwQ6jgAHmm4bAAPHAhcSgAlP0xcZdgMAEsIAHMhOCQIEggwZghs7NCwHy5p+K5Q4LAAGZTHzxPpDHJ0VRwDDkJAYCDAmZlBZhD7eLKESh-BaAxGgoYucbQeCaaQ9cjzRY0aalSkVTpRGoFBJJQgpQYQwZFcDEuCk8kwOkAQSgAEkdpsljS5iiOpFqjE6hQGmz8kqIZVGYregMCiMgA) of how the object keys could be typed if we wanted to avoid having to cast those keys.

Perhaps there is a better way, I'm totally open to suggestions?

apps/store/src/lib/l10n/locales.ts
alebedev: Can we worker around with conversion function that return proper type
`
asRoutingLocale = (locale: LocaleValues): RoutingLocale => {
  return locale.toLowerCase() as RoutingLocale
}
`

But only ways without assertion that I know are
- to have explicit mapping or explicit switch statement
- to check toLowerCase result agains all possible keys in RoutingLocale

alebedev (approved):
Nice!  Some nit comments

apps/store/src/lib/l10n/types.ts
KarnellSchultz: Yeah I prefer the Objects to Enums all day in these situations. 

I like enums for those _magic strings/numbers_ but that's about it for me.

I can replace the enums if we're cool with that?

PR #576, opened 2022-09-22
GRW-1494 / feat / Add HeroVideoBlock via S3

simonauner:
## Describe your changes

This PR is one example of how to add videos in Storyblok for the Coverage store. 

It uses external storage for the video, in my example the video is uploaded to S3.

The user then configures the blok as follows:

image

(note that the user can configure where the heading element should reside since different videos may have different optimal vertical point to put text on)

(@gustaveen suggested that we could pick up the format from the file extension, so that's a future improvement we could make after we decide on how to move forward 👍 )

For actually rendering the video I used the HTML5 Video element with correct props to make it [autoplay on iOS devices](https://webkit.org/blog/6784/new-video-policies-for-ios/).

All in all it looks like this:

https://user-images.githubusercontent.com/1343979/191693543-50a3b263-5158-4ac3-ab9f-6f89b37f96ee.mov

[GRW-1494]: https://hedvig.atlassian.net/browse/GRW-1494?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/store/README.md
+++ b/apps/store/README.md
@@ -20,3 +20,17 @@ Requirements:
 Requirements:
 
 - Add `APOLLO_KEY` to `.env.local` (see Vercel)
+
+## Using the Storyblok editor
+
+To preview your Storyblok blocks that you have developed locally, use
+
+```
+<server url>/editor.html
+```
+
+by default:
+
+```
+http://localhost:8040/editor.html
+```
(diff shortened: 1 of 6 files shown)


simonauner:
We will go ahead with using Storyblok and/or Vimeo for now. Closing this one.

PR #579, opened 2022-09-22
GRW-1494 / feat / Add HeroVideoBlock via Storyblok

simonauner:
## Describe your changes

Basically the same PR as https://github.com/HedvigInsurance/racoon/pull/576 but with uploading video assets to Storyblok instead of S3.

image

## Checklist before requesting a review

- [ ] I have performed a self-review of my code

[GRW-1493]: https://hedvig.atlassian.net/browse/GRW-1493?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #580, opened 2022-09-22
GRW-1494 / feat / Add HeroVideoBlock via Vimeo

simonauner:
## Describe your changes

More or less like https://github.com/HedvigInsurance/racoon/pull/576 and https://github.com/HedvigInsurance/racoon/pull/579 , but with Vimeo as content platform.

Here we launch an iframe and emulate `object-fit: cover` by forcing `width` to 1.8 times height (16/9 rounded).

NOTE that only certain videos will allow `background=1`.. Maybe only our own videos which we allow to be so? Haven't researched. Read more here: https://vimeo.zendesk.com/hc/en-us/articles/115011183028-Embed-background-and-chromeless-videos

This video works: `76979871` (a demo video from Vimeo), but this one doesn't: `22439234` (most viewed, the one I use in the screenshot). When videos don't allow it the controls, logo, title etc will be overlayed.

Screenshot 2022-09-22 at 14 58 57

[GRW-1494]: https://hedvig.atlassian.net/browse/GRW-1494?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


alebedev (approved):
Looks good

PR #581, opened 2022-09-22
GRW-1514 / Add Global Reference block

gustaveen:
## Update

- Renamed `globalReferenceBlock` to `reusableBlockReference`
- Created a folder Reusable Blocks and added a content type `reusableBlock`

Next step: will add a visual preview for `reusableBlock`

https://user-images.githubusercontent.com/6661511/193525196-e0775a17-6e2f-4b57-abaf-bd6550a0d867.mov

## Describe your changes
In order to reuse a block in multiple pages (the Contact support block in this example), instead of having to create in on all pages it is used, we can make use of Storybloks `resolve_relations`.

So I added a new block called `globalReference` that has a field called `reference` of type `single-option`.  Under `Source` you pick `Stories` and filter stories to the current locales `/global` folder.

Screenshot 2022-09-22 at 16 53 49

Now you can add a new story under global folder with the a `Contact Support` block as body:

https://user-images.githubusercontent.com/6661511/191781586-9d15ee83-9c37-4214-84f2-53421c7bcb17.mov

Then on any page you can add a `Global Reference` block, find the Contact Support story and add it to the page:

https://user-images.githubusercontent.com/6661511/191781679-32bd0799-cfbe-4c46-92a6-1a0840cc297b.mov

Again, I haven't updated the prod space yet, so let me know if you want to view it in my space.

**Note:** To make it easier for editors we use the `global` content type for this, the same content type as for header and footer. The benefit is that we can restrict what content type you can create under `/global` folder. 

The only gotcha I can think of is that the `header` and `footer` field is still visible on all stories. When we only will look for header and footer from the "home" story with slug `/global` 

To make it easier to reuse repeated content on the site.

For more info on the topic, check out this article: https://www.storyblok.com/tp/using-relationship-resolving-to-include-other-content-entries


simonauner (commented):
I'm sure you know what you're doing, but I don't quite understand it 🤔, but I would really like to understand!

I guess what I fail to understand is why it would be beneficial to create this global block like you did in the video, instead of just using the ContactSupport block directly.

And if we want to have it on all (or most) pages, wouldn't it make sense to place it in the footer instead? 

Maybe we can have a chat and you can explain it to me like I'm five 🧒

alebedev (approved):
Great stuff!

gustaveen:
> I'm sure you know what you're doing, but I don't quite understand it 🤔, but I would really like to understand!
> 
> I guess what I fail to understand is why it would be beneficial to create this global block like you did in the video, instead of just using the ContactSupport block directly.
> 
> And if we want to have it on all (or most) pages, wouldn't it make sense to place it in the footer instead?
> 
> Maybe we can have a chat and you can explain it to me like I'm five 🧒

For sure @simonauner ! I realise this is might not be so straight forward :D Let's do a run down of this feature after standup 🙌 And see if we feel like it brings value.

guilhermespopolin (approved):

PR #582, opened 2022-09-22
GRW-1518 / design / add country selector page design

KarnellSchultz:
## Describe your changes
- add design to country-selector page
- [Figma doc](https://www.figma.com/file/XFYNg7y7quhTZGC9cHzEt3/Hedvig.com---MVP-Wireframes?node-id=534%3A87051&viewport=-1843%2C7064%2C0.55)

### Figma Screenie

Screenshot 2022-09-22 at 17 09 07

### Dev Screenie 🤳

Screenshot 2022-09-22 at 16 59 06

- 🧽

[GRW-1518]: https://hedvig.atlassian.net/browse/GRW-1518?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/store/src/components/CountrySelectorPage/CountrySelectorPage.tsx
+++ b/apps/store/src/components/CountrySelectorPage/CountrySelectorPage.tsx
@@ -19,7 +19,7 @@ const CountryOptionsContainer = styled.div({
   gap: 8,
 })
 
-export const CountrySelectorPage = (props: CountrySelectorPageProps) => {
+export const CountrySelectorPage = (props: StoryblokPageProps) => {
   return (
     <Container {...props}>
       <Space y={3}>
@@ -31,7 +31,7 @@ export const CountrySelectorPage = (props: CountrySelectorPageProps) => {
             <Link
               key={country}
               href={PageLink.store({
-                locale: countries[country as keyof typeof countries].defaultLocale,
+                locale: routingLocale(countries[country as keyof typeof countries].defaultLocale),
               })}
               passHref
             >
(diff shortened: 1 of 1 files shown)


gustaveen (approved):
[nicesteve](https://app.graphite.dev/macros/HedvigInsurance)
Nice! I wonder if we should have other navigation like header and footer on that page since it wont know what market/locale to fetch it from. But let's check Peter says

alebedev (approved):

apps/store/src/components/CountrySelectorPage/CountrySelectorPage.tsx
```diff
@@ -12,30 +12,39 @@ const Container = styled.div({
...
 
-const CountryOptionContainer = styled.div({
+const CountryOptionsContainer = styled.div({
   display: 'grid',
-  gap: 5,
+  gap: 8,
 })
 
+const Button = styled(LinkButton)(({ theme }) => ({
+  borderRadius: theme.space[1],
```
simonauner, line 24: nit: Why should we have different border radius on this page in particular? Shouldn't we mostly stay with the defined button that we have 🤓 ? And if we want to have border-radius, shouldn't it be coded into `LinkButton`? 

The reason why I'm commenting on this is that styling our UI components with Emotion is an anti-pattern to using a Design System where buttons look and behave the same across the product 🤓

KarnellSchultz: Yeah, totally agree. Thanks for that comment. I was trying to guess at what the border-radius was and reached for that styled component.

Going to go with the default and remove the styled component 👍

simonauner (approved):
🚀

PR #587, opened 2022-09-26
GRW-1511 / GTM user properties

amandaflod:
## Describe your changes

Added user properties (market, environment) to gtm push. 

[GRW-1511]: https://hedvig.atlassian.net/browse/GRW-1511?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/store/src/lib/gtm.tsx
```diff
@@ -70,12 +63,27 @@ export const GTMBodyScript = () => (
...
   useEffect(() => {
+    const pageview = (url: string) => {
+      pushToGTMDataLayer({
+        event: 'virtual_page_view',
+        pageData: {
+          page: url,
+          title: document.title,
+        },
+        userProperties: {
+          market: countryLabel,
```
robinandeer, line 77: Should we ask @sonnyandersson1 if we should switch from calling this market in analytics also?

apps/store/src/lib/gtm.tsx
```diff
@@ -70,12 +63,27 @@ export const GTMBodyScript = () => (
...
+          environment: process.env.NODE_ENV,
```
robinandeer, line 78: Pretty sure this will be "production" all the time unless local dev where we shouldn't send any analytics

Let's use a custom environment variable for this :)

gustaveen, line 78: As I understood from Sonny, this shouldn't be pushed on route changes but rather one time when a user lands on the page before any other pushes are sent. And then it should listen on `environment` and `country` changes. :) So I think we should split up that in to two datalayer pushes

Example from web-onboarding:
```js
  useEffect(() => {
    pushToGTMDataLayer({
      userProperties: {
        environment,
        market,
      },
    })
  }, [environment, market])
```

So it might be better to rename `useGTMRouteEvents` since not all events will be setup to listen to route changes.

robinandeer (commented):
Nice changes, just think we can take it even further :D

If you like, do also consider moving this module from `lib` -> `services` since it's a 3rd party service.

apps/store/src/lib/gtm.tsx
```diff
@@ -5,12 +5,15 @@
 import { useRouter } from 'next/router'
 import Script from 'next/script'
 import { useEffect } from 'react'
+import { useCurrentLocale } from './l10n/useCurrentLocale'
 
 export const GTM_ID = process.env.NEXT_PUBLIC_GOOGLE_TAG_MANAGER_ID
 
+type AppEnvironment = 'development' | 'production' | 'test'
```
gustaveen, line 12: We've used env variables for this previously to distinguish between staging and development. And I think `NODE_ENV` will be production in staging :) 

export type AppEnvironment = 'development' | 'staging' | 'production'

apps/store/src/lib/gtm.tsx
gustaveen: Let's check with him, I guess it depends on if he want to be able to compare properties with the current site

apps/store/src/lib/gtm.tsx
```diff
@@ -70,12 +63,27 @@ export const GTMBodyScript = () => (
 
 export const useGTMRouteEvents = () => {
   const router = useRouter()
+  const { countryLabel } = useCurrentLocale()
```
alebedev, line 66: There's `useCurrentCountry()`, does it do what you need without manually extending locale data?

alebedev (commented):
Looks good overall, let's just figure out right way to get country

apps/store/src/lib/l10n/locales.ts
```diff
@@ -1,36 +1,43 @@
-import { Language, Locale, LocaleValue, RoutingLocale } from './types'
+import { CountryCode, Language, Locale, LocaleValue, RoutingLocale } from './types'
 
 export const FALLBACK_LOCALE = Locale.EnSe
 
 export type LocaleData = {
   locale: LocaleValue
   language: Language
+  country: CountryCode
```
alebedev, line 8: I think we can remove this now and use `getCountryByLocale()` as a source of truth

alebedev (approved):

apps/store/src/services/gtm.tsx
```diff
@@ -5,12 +5,14 @@
...
+import { CountryCode } from '@/lib/l10n/types'
+import { useCurrentCountry } from '@/lib/l10n/useCurrentCountry'
 
 export const GTM_ID = process.env.NEXT_PUBLIC_GOOGLE_TAG_MANAGER_ID
 
 type GTMUserProperties = {
-  market: string
-  environment: 'development' | 'production' | 'test'
+  country: CountryCode
+  environment?: string
```
robinandeer, line 15: any particular reason why we want to make this optional and remove the type?

robinandeer (approved):
Looks really good!

I think I agree with @alebedev on using the utility function to get country by locale.

apps/store/src/services/gtm.tsx
gustaveen: Would be nice if we could set standardised env name as @simonauner suggested https://hedviginsurance.slack.com/archives/G01MARCFUUW/p1664259960731899
But we can add that once we come to a decision. 🙌

apps/store/src/services/gtm.tsx
```diff
@@ -70,8 +62,28 @@ export const GTMBodyScript = () => (
 
 export const useGTMRouteEvents = () => {
```
gustaveen, line 63: nit: maybe we should call it something more generic since user properties isn't triggered by by route change :)

apps/store/src/lib/gtm.tsx
amandaflod: I checked and he was fine with renaming it to country! :)

apps/store/src/services/gtm.tsx
amandaflod: I adopted the suggestion `<local|staging|prod>` and updated the type! :) Hope that works for now and let's try to make a decision on it soon.

PR #594, opened 2022-09-26
Redirect users to locale instead of rewriting.

robinandeer:
## Describe your changes

Redirect users instead of rewriting the URL in middleware.

Keep rewriting URL when we want to show country selector page.

We want the URL to always match the locale.

### DIFF
--- a/apps/store/src/middleware.ts
+++ b/apps/store/src/middleware.ts
@@ -10,21 +10,24 @@ export const config = {
 export function middleware(req: NextRequest) {
   const country = req.geo && req.geo.country
 
+  const nextURL = req.nextUrl.clone()
+
   switch (country) {
     case countries.NO.id:
-      req.nextUrl.pathname = routingLocale(countries.NO.defaultLocale)
+      nextURL.pathname = routingLocale(countries.NO.defaultLocale)
       break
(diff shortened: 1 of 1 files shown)


apps/store/src/middleware.ts
```diff
@@ -10,21 +10,24 @@ export const config = {
...
-      req.nextUrl.pathname = routingLocale(countries.DK.defaultLocale)
+      nextURL.pathname = routingLocale(countries.DK.defaultLocale)
       break
     case countries.SE.id:
-      req.nextUrl.pathname = routingLocale(countries.SE.defaultLocale)
+      nextURL.pathname = routingLocale(countries.SE.defaultLocale)
       break
     default:
-      req.nextUrl.pathname = '/country-selector'
+      console.info(`Routing visitor from ${country} to country selector`)
```
alebedev, line 26: Why not `debug`?  We probably don't want to see these messages in Datadog

alebedev (approved):
Looks good.  I'll add my changes that ignores requests that already have known locale after this is merged

apps/store/src/middleware.ts
robinandeer: good point

PR #595, opened 2022-09-27
Implement logger for Datadog on server side

simonauner:
## Describe your changes

* Replace all `console.log|err|warn` with `logger` calls on server-side code
* Use both DD transport and terminal output for local development

See screenshot where all variables are shown:
image

You can even filter on your own machine:
image

I still haven't figured out how to get the Vercel URL into the Datadog logs, neither `hostname`, `hostName` or `host` seems to set it 🤔 

We need to be able to troubleshoot server-side code in production and it's not possible by just using `console` statements.


apps/store/src/services/logger/client.ts
```diff
@@ -1,39 +1,30 @@
...
 export const initDatadog = () => {
   if (APPLICATION_ID && CLIENT_TOKEN) {
     datadogRum.init({
-      ...COMMON_CONFIG,
+      ...CLIENT_CONFIG,
       clientToken: CLIENT_TOKEN,
       applicationId: APPLICATION_ID,
       trackInteractions: true,
       defaultPrivacyLevel: 'mask-user-input',
+      silentMultipleInit: true,
```
robinandeer, line 16: oh thanks 🙏

robinandeer (approved):
Awesome!

Added via Giphy

PR #607, opened 2022-09-28
Use template instead of product name for price intent key

robinandeer:
## Describe your changes

Use template instead of product name for price intent key

Pass the whole price template when fetching/creating price intent.

Add unique name property to Template.

This is needed to support using the same product on multiple product pages.

We don't want to mix up the price intents in these cases tho so we need a different identifier than shop session + product name.

Since the template will be different for different product pages I decided to use that combo instead.

### DIFF
--- a/apps/store/src/blocks/PriceCalculatorBlock.tsx
+++ b/apps/store/src/blocks/PriceCalculatorBlock.tsx
@@ -44,7 +44,7 @@ export const PriceCalculatorBlock = ({ blok }: Props) => {
       gradient: product.gradient,
     })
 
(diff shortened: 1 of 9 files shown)


robinandeer:
* main
  * **PR #607** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/607" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/607?utm_source=stack-comment).

simonauner (approved):
Nice! Good find 🙌

PR #608, opened 2022-09-28
GRW-1564 / Feature / Type-safe handling of different locales

alebedev:
## Describe your changes

- Renamed some locale-related types to express their intended use
- Make most locale-specific code agnostic to type of input locale

Our locale-handling code is a mess, hard to understand and easy to use incorrectly. See discussion in #hedvig-dot-com for details

[GRW-1564]: https://hedvig.atlassian.net/browse/GRW-1564?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/store/src/lib/l10n/locales.ts
+++ b/apps/store/src/lib/l10n/locales.ts
@@ -52,17 +52,17 @@ export enum LocaleField {
   Language = 'language',
 }
 
-export const isoLocale = (locale: UiLocale): IsoLocale => {
+export const toIsoLocale = (locale: UiLocale): IsoLocale => {
   const parts = locale.split('-', 2)
   return `${parts[0].toLowerCase()}-${parts[1].toUpperCase()}` as IsoLocale
 }
 
-export const routingLocale = (locale: UiLocale) => locale.toLowerCase() as RoutingLocale
+export const toRoutingLocale = (locale: UiLocale) => locale.toLowerCase() as RoutingLocale
 
 // TODO: Make fallback market-specific
 export const getLocaleOrFallback = (locale: UiLocale | string | undefined): LocaleData => {
   if (!isSupportedLocale(locale)) {
     return locales[FALLBACK_LOCALE]
   }
-  return locales[isoLocale(locale)]
+  return locales[toIsoLocale(locale)]
 }
--- a/apps/store/src/lib/l10n/locales.ts
+++ b/apps/store/src/lib/l10n/locales.ts
@@ -1,13 +1,14 @@
-import { Language, Locale, LocaleValue, RoutingLocale } from './types'
+import { isSupportedLocale } from '@/utils/isSupportedLocale'
+import { IsoLocale, Language, Locale, RoutingLocale, UiLocale } from './types'
 
 export const FALLBACK_LOCALE = Locale.EnSe
 
 export type LocaleData = {
-  locale: LocaleValue
+  locale: IsoLocale
   language: Language
 }
 
-export const locales = {
+export const locales: Record<IsoLocale, LocaleData> = {
   [Locale.SvSe]: {
     locale: Locale.SvSe,
     language: Language.Sv,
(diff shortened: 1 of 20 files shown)


apps/store/src/lib/l10n/types.ts
```diff
@@ -10,9 +10,12 @@ export const Locale = {
   EnDk: 'en-DK',
 } as const
 
-export type LocaleType = typeof Locale
-export type LocaleValue = typeof Locale[keyof typeof Locale]
-export type RoutingLocale = Lowercase<LocaleValue>
+export type IsoLocale = typeof Locale[keyof typeof Locale]
```
robinandeer, line 13: Very nice name change :)

apps/store/src/lib/l10n/locales.ts
```diff
@@ -51,18 +52,17 @@ export enum LocaleField {
   Language = 'language',
 }
 
-export const normalizeLocale = (locale: string | undefined): string | undefined => {
-  if (!locale?.includes('-')) {
-    return locale
-  }
+export const isoLocale = (locale: UiLocale): IsoLocale => {
```
robinandeer, line 55: I would rename this function... it's a little hard to understand what it's supposed to do

Perhaps `formatIsoLocale` or `convertToIsoLocale`?

apps/store/src/utils/isSupportedLocale.ts
```diff
@@ -0,0 +1,9 @@
+import { routingLocale } from '@/lib/l10n/locales'
+import { Locale, UiLocale } from '@/lib/l10n/types'
+
+export const isSupportedLocale = (locale: unknown): locale is UiLocale => {
+  return (
+    typeof locale === 'string' &&
+    Object.values(Locale).some((x) => x === locale || routingLocale(x) === locale)
```
robinandeer, line 7: nit: create a constant from the value of `Object.values(Locale)` to gain a little performance

apps/store/src/middleware.ts
```diff
@@ -1,15 +1,17 @@
...
+import { NextLocale } from '@/lib/l10n/types'
+import { isSupportedLocale } from '@/utils/isSupportedLocale'
 
 export const config = {
   matcher: '/',
 }
 
 export function middleware(req: NextRequest) {
-  if (isLocale(normalizeLocale(req.nextUrl.locale))) {
+  const locale = req.nextUrl.locale as NextLocale
```
robinandeer, line 13: Do we need to assert the type here? Seems like `isSupportedLocale` can accept any value :)

apps/store/src/lib/l10n/types.ts
```diff
@@ -10,9 +10,12 @@ export const Locale = {
...
+// Would be different in the future - 'se' for Sweden+Swedish, for example
+export type RoutingLocale = Lowercase<IsoLocale>
+export type NextLocale = RoutingLocale | 'default'
+// TODO: Make all locale-taking functions accept this:
+export type UiLocale = IsoLocale | RoutingLocale
```
robinandeer, line 18: You know better than me if this is possible...

However, would it not be even better if we only passed around `IsoLocale` in our code and converted it to `RoutingLocale` on-demand?

robinandeer (approved):
Great work! Sorry about all the comments 😅

apps/store/src/utils/isSupportedLocale.ts
alebedev: If we want to really optimize here, we should replace linear array lookup with Set.has and precompute possible values

Tried doing this, then rolled back because it does not seem that adding complexity is justified.   This is like memoizing some function without profiling first

alebedev: Looked roughly like this
```
const supportedLocaleVariants = new Set<string>(
  Object.values(Locale).flatMap((locale) => [locale, toRoutingLocale(locale)]),
)

export const isSupportedLocale = (locale: unknown): locale is UiLocale => {
  return typeof locale === 'string' && supportedLocaleVariants.has(locale)
}
```

apps/store/src/lib/l10n/types.ts
alebedev: This is what we tried before - it proved difficult to track where locale comes from and required explicit type casting when interacting with next / Storyblok

I'll try one more time when implementing `/se` like URLs for default languages

PR #609, opened 2022-09-28
Auto-focus first input when accordion is expanded

simonauner:
Consider this a proposal to solve an a11y issues by using a non-recommended (in terms of a11y) solution

## Describe your changes

Set `auto-focus` on a PriceForm input when
1. It's any section after the first
2. It's the first input in the section

Otherwise, the user has to tab through the entire page every time they submit the form.

The use of `auto-focus` is [discouraged](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/docs/rules/no-autofocus.md) since it can cause usability and a11y issues, but as it solves the current issue, which is also a11y, I think it's worth the trade-off.

The user interaction would be (for https://racoon-store.vercel.app/sv-se/products/home):
1. Tab to the form
2. Fill in SSN
3. Enter to submit section
4. Tab to next input (Ownership type)
5. Fill in section
6. Enter to submit section
7. Tab to next input (Ownership type)
5. Fill in section
6. Enter to submit section and calculate price

After this PR:
1. Tab to the form
2. Fill in SSN
3. Enter to submit section
5. Fill in section
6. Enter to submit section
5. Fill in section
6. Enter to submit section and calculate price

#### Video before

https://user-images.githubusercontent.com/1343979/192765337-4573d940-180d-4092-8c3a-39078ddadc4b.mov

#### Video after

https://user-images.githubusercontent.com/1343979/192765388-41473e34-db84-41d0-aa1d-c94284833e56.mov

### DIFF
--- a/apps/store/src/components/PriceForm/PriceForm.tsx
+++ b/apps/store/src/components/PriceForm/PriceForm.tsx
@@ -34,11 +34,20 @@ export const PriceForm = ({ form, priceIntent, onSuccess, loading }: Props) => {
 
   return (
     <PriceFormAccordion form={form}>
-      {(section) => (
+      {(section, sectionIndex) => (
         <PriceFormSection section={section} onSubmit={handleSubmit} loading={isLoading}>
           <FormGrid items={section.items}>
-            {(field) => (
-              <AutomaticField field={field} onSubmit={handleSubmit} loading={isLoading} />
+            {(field, index) => (
+              <AutomaticField
+                field={field}
+                onSubmit={handleSubmit}
+                loading={isLoading}
+                // We don't want to mess up focusing for the user by setting autoFocus on the
+                // first item in the form, since that would make it unintuitive to navigate our
+                // site. But when the user is in the form editing, even having submitted the first
+                // section, we want to set autoFocus for the next section. Hence sectionIndex > 0
+                autoFocus={sectionIndex > 0 && index === 0}
+              />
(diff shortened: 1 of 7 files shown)


apps/store/src/components/PriceForm/PriceForm.tsx
```diff
@@ -34,11 +34,20 @@ export const PriceForm = ({ form, priceIntent, onSuccess, loading }: Props) => {
...
-              <AutomaticField field={field} onSubmit={handleSubmit} loading={isLoading} />
+            {(field, index) => (
+              <AutomaticField
+                field={field}
+                onSubmit={handleSubmit}
+                loading={isLoading}
+                // We don't want to mess up focusing for the user by setting autoFocus on the
+                // first item in the form, since that would make it unintuitive to navigate our
+                // site. But when the user is in the form editing, even having submitted the first
+                // section, we want to set autoFocus for the next section. Hence sectionIndex > 0
```
robinandeer, line 48: Should this be an HTML comment? 🤔

robinandeer (approved):
Nice improvement :D

apps/store/src/components/PriceForm/PriceForm.tsx
simonauner: Nope, comments in JSX are JS comments.. At least according to VSCode 🙃

PR #612, opened 2022-09-28
feat / Introduce custom field for Swedish SSN

simonauner:
## Describe your changes

Introduce custom field that formats Swedish SSN to a format that is acceptable by BE. The `personnummer` lib handles validation and formatting in an easy way.

We should allow the user to type in SSN in all different variations and always send it in the correct format to BE. The format that they accept is "long" according to [personnummer docs](https://personnummer.dev/) -> `yyyymmddNNNN`.

Maybe @robinandeer knows some good way to get rid of the `useState`? 🙃 

### Video

https://user-images.githubusercontent.com/1343979/192812438-fff5d157-ba17-4063-bb64-f76ef7188509.mov

### DIFF
--- a/apps/store/src/components/PriceForm/SsnSeField.tsx
+++ b/apps/store/src/components/PriceForm/SsnSeField.tsx
@@ -1,0 +1,46 @@
+import Personnummer from 'personnummer'
+import { ChangeEventHandler, useState } from 'react'
+import { InputField } from 'ui'
+import { SsnSeField as SsnSeFieldType } from '@/services/PriceForm/Field.types'
+import { useTranslateTextLabel } from './useTranslateTextLabel'
+
+type SsnSeFieldProps = {
+  field: SsnSeFieldType
+}
+
+export const SsnSeField = ({ field }: SsnSeFieldProps) => {
+  const translateLabel = useTranslateTextLabel({ data: {} })
+  const [value, setValue] = useState(field.defaultValue)
+
+  const handleOnChange: ChangeEventHandler<HTMLInputElement> = (event) => {
+    let value = event.target.value
+
+    if (typeof value === 'string' && Personnummer.valid(value)) {
+      value = Personnummer.parse(value).format(true)
+    }
+    setValue(value)
+  }
+
+  return (
+    <>
+      <input
+        type="text"
+        name={field.name}
+        required={field.required}
+        defaultValue={field.value ?? field.defaultValue}
+        value={value}
+        hidden
+      />
+      <InputField
+        type="text"
+        name={`${field.name}-visible-input`}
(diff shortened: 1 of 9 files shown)


apps/store/src/components/PriceForm/SsnSeField.tsx
```diff
@@ -0,0 +1,37 @@
...
+  const translateLabel = useTranslateTextLabel({ data: {} })
+  const [value, setValue] = useState(field.defaultValue)
+
+  const handleOnChange: ChangeEventHandler<HTMLInputElement> = async (event) => {
+    let value = event.target.value
+
+    if (typeof value === 'string' && Personnummer.valid(value)) {
+      value = Personnummer.parse(value).format(true)
+    }
+    setValue(value)
```
alebedev, line 21: I think useState is fine here.  Other option is using uncontrolled element and changing value with DOM API but this does seem like a good mix with React `onChange` handler.  Either full-DOM or full-React (like you use) options seem more predictable

alebedev (approved):
Nice!

apps/store/src/components/PriceForm/SsnSeField.tsx
robinandeer: I don't think using `useState` here is bad :)

From outside this component, it's basically an uncontrolled input

apps/store/src/components/PriceForm/SsnSeField.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+type SsnSeFieldProps = {
+  field: SsnSeFieldType
+}
+
+export const SsnSeField = ({ field }: SsnSeFieldProps) => {
+  const translateLabel = useTranslateTextLabel({ data: {} })
+  const [value, setValue] = useState(field.defaultValue)
+
+  const handleOnChange: ChangeEventHandler<HTMLInputElement> = async (event) => {
```
guilhermespopolin, line 15: Why is this function `async`? 🤔

simonauner: That's because I use the famous programming method of Copy Paste 😂 
Thanks! 🕵️ 🏅

robinandeer:
Nice stuff!

It might be a little odd for users the input is changed on the fly, no?

Another option would be to use this as the displayed input but use another hidden input with the transformed value which is actually submitted with the form 😉

simonauner:
> Nice stuff!
> 
> It might be a little odd for users the input is changed on the fly, no?
> 
> Another option would be to use this as the displayed input but use another hidden input with the transformed value which is actually submitted with the form 😉

Ah, yes, I like it! I'll try it out 👍

PR #620, opened 2022-09-29
Update to fix layout on checkout contact details page

robinandeer:
## Describe your changes

Fix link back to review/start date screen on Checkout Contact Details page.

Tweak layout of same page to match the review/start date screen.

## Checklist before requesting a review

- [ ] I have performed a self-review of my code

[GRW-1569]: https://hedvig.atlassian.net/browse/GRW-1569?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #613, opened 2022-09-28
Integrate checkout signing API

robinandeer:
## Describe your changes

Integrate Checkout signing API on checkout / contact details page.

Move form logic inside checkout details pages.

Keep track of started "checkout signing" in cookie (key = checkoutId).

Flow: update contact details -> start sign -> poll for updates to checkout signing.

Still to be done:

- store the access token in a cookie after it's generated
- update cookie settings to be secure

## Justify why they are needed

Needed to support the "after sign" payment flow.

We stored the checkout signing ID in a cookie in case the user reloads the page while a signing is ongoing.

## Checklist before requesting a review

- [ ] I have performed a self-review of my code

[GRW-1562]: https://hedvig.atlassian.net/browse/GRW-1562?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #610, opened 2022-09-28
Switch to using root-level priceIntent query

robinandeer:
## Describe your changes

see title.

## Justify why they are needed

The old query is deprecated.

## Checklist before requesting a review

- [ ] I have performed a self-review of my code

PR #622, opened 2022-09-29
Store access token in cookie

robinandeer:
## Describe your changes

Store access token in encrypted cookie after successful signing.

Add `cookies-next` as dependency to store app. We already used it.

Pass req + res when initializing apollo to get access token from cookies server side.

Use "auth link" from official example to pick up access token from cookies client side.

We need the authorization header to make authenticated calls to the API.

We want to access the token both client and server-side so best way to store it is in a cookie.

We should probably replace the `js-cookie` dependency and just use `cookies-next`.

[GRW-1564]: https://hedvig.atlassian.net/browse/GRW-1564?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/store/src/components/CheckoutContactDetailsPage/CheckoutSignPage.tsx
+++ b/apps/store/src/components/CheckoutContactDetailsPage/CheckoutSignPage.tsx
@@ -1,8 +1,12 @@
 import styled from '@emotion/styled'
+import { setCookie } from 'cookies-next'
 import { Button, InputField, Space } from 'ui'
+import { AUTH_COOKIE_KEY } from '@/services/apollo/client'
 import { CheckoutContactDetailsPageProps } from './CheckoutContactDetails.types'
 import { CheckoutContactDetailsPageLayout } from './CheckoutContactDetailsPageLayout'
 import { useHandleSubmitContactDetailsAndSign } from './useHandleSubmitContactDetailsAndSign'
+
+const MAX_AGE = 60 * 60 * 24 // 24 hours
 
 export const CheckoutSignPage = ({
   checkoutId,
@@ -13,8 +17,14 @@ export const CheckoutSignPage = ({
   const [handleSubmit, loading] = useHandleSubmitContactDetailsAndSign({
     checkoutId,
     checkoutSigningId,
-    onSuccess() {
-      // @TODO: add access token to session
+    onSuccess(accessToken) {
+      setCookie(AUTH_COOKIE_KEY, accessToken, {
+        maxAge: MAX_AGE,
+        expires: new Date(Date.now() + MAX_AGE * 1000),
+        sameSite: 'strict',
+        secure: process.env.NODE_ENV === 'production',
+        path: '/',
+      })
       onSignSuccess()
     },
   })
--- a/apps/store/src/pages/products/[product].tsx
+++ b/apps/store/src/pages/products/[product].tsx
@@ -37,7 +37,7 @@ export const getServerSideProps: GetServerSideProps<NextPageProps> = async (cont
   const { countryCode } = getCountryByLocale(locale)
 
   try {
-    const apolloClient = initializeApollo()
+    const apolloClient = initializeApollo(undefined, req, res)
 
     const [shopSession, story, globalStory] = await Promise.all([
       getShopSessionServerSide({ req, res, apolloClient, countryCode }),
--- a/apps/store/src/services/apollo/client.ts
+++ b/apps/store/src/services/apollo/client.ts
@@ -1,9 +1,13 @@
 import { ApolloClient, from, HttpLink, InMemoryCache, NormalizedCacheObject } from '@apollo/client'
+import { setContext } from '@apollo/client/link/context'
 import { onError } from '@apollo/client/link/error'
 import { mergeDeep } from '@apollo/client/utilities'
+import { getCookie } from 'cookies-next'
+import { GetServerSidePropsContext } from 'next'
 import { useMemo } from 'react'
 
 export const APOLLO_STATE_PROP_NAME = '__APOLLO_STATE__'
+export const AUTH_COOKIE_KEY = 'HEDVIG_ACCESS_TOKEN'
 
 let apolloClient: ApolloClient<NormalizedCacheObject>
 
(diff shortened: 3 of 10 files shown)


robinandeer:
* main
  * **PR #610** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/610" target="_blank">Graphite</a> 
    * **PR #613** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/613" target="_blank">Graphite</a> 
      * **PR #620** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/620" target="_blank">Graphite</a> 
        * **PR #622** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/622" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/622?utm_source=stack-comment).

apps/store/src/components/CheckoutContactDetailsPage/CheckoutSignPage.tsx
```diff
@@ -13,8 +17,14 @@ export const CheckoutSignPage = ({
   const [handleSubmit, loading] = useHandleSubmitContactDetailsAndSign({
     checkoutId,
     checkoutSigningId,
-    onSuccess() {
-      // @TODO: add access token to session
+    onSuccess(accessToken) {
+      setCookie(AUTH_COOKIE_KEY, accessToken, {
+        maxAge: MAX_AGE,
```
simonauner, line 22: We should probably sync this with BE so that we don't store the cookie longer than what their expiry is for an auth token

apps/store/src/services/apollo/client.ts
```diff
@@ -15,22 +19,42 @@ const errorLink = onError(({ graphQLErrors, networkError }) => {
...
+  const accessToken = getCookie(AUTH_COOKIE_KEY, { req, res })
+
   return new ApolloClient({
     name: 'Web:Racoon:Store',
     ssrMode: typeof window === 'undefined',
-    link: from([errorLink, httpLink]),
+    link: from([errorLink, authLink, httpLink]),
     cache: new InMemoryCache(),
+    headers: {
+      authorization: typeof accessToken === 'string' ? accessToken : '',
```
simonauner, line 47: If the other possible type for `accessToken` is `undefined` this could be `accessToken ?? ''`

apps/store/src/services/apollo/client.ts
```diff
@@ -15,22 +19,42 @@ const errorLink = onError(({ graphQLErrors, networkError }) => {
...
 })
 
-const createApolloClient = () => {
+const httpLink = new HttpLink({ uri: process.env.NEXT_PUBLIC_GRAPHQL_ENDPOINT })
+
+const createApolloClient = (
+  req?: GetServerSidePropsContext['req'],
+  res?: GetServerSidePropsContext['res'],
+) => {
+  const accessToken = getCookie(AUTH_COOKIE_KEY, { req, res })
```
simonauner, line 39: Do we need `req` and `res` all the way down here for `authLink`? Otherwise we could just pass the accessToken I guess instead of making everything aware of Next.js here in Apollo client.. But perhaps it already is? :D

simonauner (approved):
Nice! I don't understand how Apollo sets it up, but I think I get the rest :D 

gif

apps/store/src/components/CheckoutContactDetailsPage/CheckoutSignPage.tsx
robinandeer: Yep, that information should be available in the API :)

In this case I think it basically never expires 🫣

robinandeer: We should also abstract this away in some service in the future

apps/store/src/services/apollo/client.ts
robinandeer: Unfortunately, it can also be `boolean | null` so we have to be more specific in this check... not sure why

apps/store/src/services/apollo/client.ts
robinandeer: The `authLink` is only relevant client-side since it reads cookies from `document`.

Req + res is used to set a default header when initializing apollo server-side. We could perhaps just pass in the `accessToken` to this function tho...

alebedev (approved):

apps/store/src/pages/products/[product].tsx
```diff
@@ -37,7 +37,7 @@ export const getServerSideProps: GetServerSideProps<NextPageProps> = async (cont
   const { countryCode } = getCountryByLocale(locale)
 
   try {
-    const apolloClient = initializeApollo()
+    const apolloClient = initializeApollo(undefined, req, res)
```
Comment (guilhermespopolin) [inline apps/store/src/pages/products/[product].tsx:40]: How do you feel about changing `initializApollo` function signature to accept an object instead of a list of arguments like the following:

```suggestion
    const apolloClient = initializeApollo({ req, res })
```

I understand that TS will help us to understand what is that first argument that we're passing `undefined` to but by using an object it would be even clear and parameter order wouldn't matter.

apps/store/src/components/CheckoutContactDetailsPage/CheckoutSignPage.tsx
```diff
@@ -1,9 +1,13 @@
 import styled from '@emotion/styled'
+import { setCookie } from 'cookies-next'
 import { Button, InputField, Space } from 'ui'
+import { AUTH_COOKIE_KEY } from '@/services/apollo/client'
 import { CheckoutContactDetailsPageProps } from './CheckoutContactDetails.types'
 import { CheckoutContactDetailsPageLayout } from './CheckoutContactDetailsPageLayout'
 import { useHandleSubmitContactDetailsAndSign } from './useHandleSubmitContactDetailsAndSign'
 
+const MAX_AGE = 60 * 60 * 24 // 24 hours
```
guilhermespopolin, line 9:
```suggestion
const AUTH_COOKIE_MAX_AGE = 60 * 60 * 24 // 24 hours
```

guilhermespopolin (approved):
Well done. Just added a couple of comments.

apps/store/src/pages/products/[product].tsx
Comment (robinandeer) [reply apps/store/src/pages/products/[product].tsx:40]: yep I think that's a good idea -- will add it to the backlog!

apps/store/src/components/CheckoutContactDetailsPage/CheckoutSignPage.tsx
robinandeer: Will change in the later PR that moves this constant

robinandeer:
**Graphite Merge Job**

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/622?utm_source=stack-comment).


Job Reference: otDkpy8tB4rmLMp8y6We

PR #625, opened 2022-09-30
Discussion: mocks refactoring

guilhermespopolin:
## Describe your changes

I've recently had a problem with running _Hedvig.com_ locally and it turns out I had _msw_ mocks enabled with mocks hat were outdated. My initial thought was to update the mocks but then I've realised that the maintenance work of having mocks definition up to date might not worth for existing API's like _PriceIntent_ and _Shopsession_.

For APIs that don't exist yet, I can see the benefit of defining mocks as we could work in parallel with BE while they implement the actual API. But as soon we have those APIs exposed on staging, the work of keeping mocks definition up to date seems waste since we end up not using them.

Despite of that, I still believe it might be worth to keep mock setup for [testing edge cases](https://kentcdodds.com/blog/stop-mocking-fetch#colocation-and-erroredge-case-testing). What do you guys think?


guilhermespopolin:
* main
  * **PR #625** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/625" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/625?utm_source=stack-comment).

robinandeer (approved):
I like it! I think this will make the mocks more usable to us in this moment

Let's see what we should add back once the UI is stable enough to write tests against :)

PR #629, opened 2022-09-30
Implement signing on checkout payment page

robinandeer:
## Describe your changes

Implement signing on payment page in Checkout flow.

Add new "Auth" service to collect related logic.

Add new "Checkout" service to collect related logic.

- Add new `useHandleSignCheckout` hook to deal with only signing a checkout

This is how the "payment before sign" flow should work.

However, I think we should break up the payment page more in the future before we add "after sign" and Trustly support.

[GRW-1565]: https://hedvig.atlassian.net/browse/GRW-1565?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


robinandeer:
* main
  * **PR #610** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/610" target="_blank">Graphite</a> 
    * **PR #613** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/613" target="_blank">Graphite</a> 
      * **PR #620** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/620" target="_blank">Graphite</a> 
        * **PR #622** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/622" target="_blank">Graphite</a> 
          * **PR #629** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/629" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/629?utm_source=stack-comment).

apps/store/src/pages/checkout/payment.tsx
```diff
@@ -36,12 +40,23 @@ export const getServerSideProps: GetServerSideProps<CheckoutPaymentPageAdyenProp
       countryCode: shopSession.countryCode,
     })
 
+    const checkoutSigning = await fetchCurrentCheckoutSigning({
+      req,
+      apolloClient,
+      checkoutId: shopSession.checkout.id,
+    })
+
+    if (isPaymentBeforeSign && checkoutSigning?.completion) {
```
alebedev, line 49: We already check `isPaymentBeforeSign` above, do we need to do it here as well?

apps/store/src/services/Auth/Auth.ts
```diff
@@ -0,0 +1,23 @@
+import { getCookie, setCookie } from 'cookies-next'
+import { GetServerSidePropsContext } from 'next'
+
+const AUTH_COOKIE_KEY = 'HEDVIG_ACCESS_TOKEN'
+const MAX_AGE = 60 * 60 * 24 // 24 hours
+
+export const save = (accessToken: string) => {
+  setCookie(AUTH_COOKIE_KEY, accessToken, {
+    maxAge: MAX_AGE,
+    expires: new Date(Date.now() + MAX_AGE * 1000),
```
simonauner, line 10: We don't need both `maxAge` and `expires`, right?

apps/store/src/services/Auth/Auth.ts
```diff
@@ -0,0 +1,23 @@
+import { getCookie, setCookie } from 'cookies-next'
```
simonauner, line 1: This isn't a module installed in `package.json`, or is it? Should we really rely on this without installing it?

alebedev (approved):
Nice, cleaner separation of layers than we had before

apps/store/src/services/Checkout/useHandleSignCheckout.ts
```diff
@@ -0,0 +1,40 @@
...
+        deleteCookie(checkoutId)
+      }
+    },
+  })
+
+  const [startSign, { loading: loadingStartSign }] = useCheckoutStartSignMutation({
+    variables: { checkoutId },
+    onCompleted(data) {
+      setCheckoutSigningId(data.checkoutStartSign.signing.id)
+      setCookie(checkoutId, data.checkoutStartSign.signing.id)
```
simonauner, line 34: What do we use this cookie for? I can't see it being used other than being deleted on row 25

apps/store/src/services/Auth/Auth.ts
```diff
@@ -0,0 +1,23 @@
...
+    path: '/',
+  })
+}
+
+export const getAccessToken = (
+  req?: GetServerSidePropsContext['req'],
+  res?: GetServerSidePropsContext['res'],
+) => {
+  const value = getCookie(AUTH_COOKIE_KEY, { req, res })
+  return typeof value === 'string' ? value : undefined
```
simonauner, line 22: Isn't cookie values always of type string or undefined? If so, this doesn't add any value?

apps/store/src/services/Auth/Auth.ts
robinandeer: Oh I thought it was for cross-browser compat but seems only IE was lagging behind -- will keep maxAge only then

apps/store/src/services/Auth/Auth.ts
robinandeer: I think we installed it perhaps in the previous PR...

apps/store/src/services/Auth/Auth.ts
robinandeer: haha I think it's the lib we use that transforms it maybe -- it claims it can be null and boolean also :P

so just staying on the safe side and trying to make TS happy

apps/store/src/services/Checkout/useHandleSignCheckout.ts
robinandeer: We use it to resume an ongoing signing if the user reloads the page for example

Perhaps it would be clearer if it was part of a "service"

apps/store/src/pages/checkout/payment.tsx
robinandeer: true good catch

robinandeer:
**Graphite Merge Job**

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/629?utm_source=stack-comment).


Job Reference: otDkpy8tB4rmLMp8y6We

apps/store/src/services/Auth/Auth.ts
simonauner: aaah, ok. Difficult to keep track of 6 stacked PR:s :D

PR #647, opened 2022-10-03
Configure Datadog service name for RUM (client)

robinandeer:
## Describe your changes

Add `NEXT_PUBLIC_DATADOG_SERVICE_NAME` env var that will be exposed to the JS client to set Datadog service name.

This is required for Datadog tracing.

I decided to keep `DD_SERVICE` since it's the official env var for this setting.
Perhaps Datadog on the server will read this as well...


robinandeer:
* main
  * **PR #647** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/647" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/647?utm_source=stack-comment).

simonauner (approved):
Nice fix, but weird that this stopped working, I'm quite sure it worked when I was implementing the server part 🤔 .

The BE logging will not read the env variable per se, but the `service` prop in the config is used.

alebedev (approved):
Looks good

robinandeer (commented):
@simonauner yeah exactly, I guess the env var would only be a backup... perhaps we can remove it and just rely on the config :)

Don't think anything stopped working -- `DD_SERVICE` wasn't used before in the client I guess.

PR #648, opened 2022-10-03
Store users language preference in cookie

simonauner:
Use it in middleware to redirect to instead of country picker

## Describe your changes

Store users' preference of language in cookie.

If the user has picked a language preference they shouldn't need to see the Country selector again. And if they picked English in Sweden the country selector page would still route to `/se`, not `/en-se`.

## Jira issue(s): []

Don't mind the error messages. They appear when clearing cookies.

https://user-images.githubusercontent.com/1343979/193598594-23414fa9-605a-4cbb-9c30-b665758c28af.mov

### DIFF
--- a/apps/store/src/blocks/FooterBlock.tsx
+++ b/apps/store/src/blocks/FooterBlock.tsx
@@ -9,7 +9,12 @@ import { InputSelect } from '@/components/InputSelect/InputSelect'
 import { SpaceFlex } from '@/components/SpaceFlex/SpaceFlex'
 import { countries } from '@/lib/l10n/countries'
 import { getCountryLocale } from '@/lib/l10n/countryUtils'
-import { LocaleField, TEMP_TRANSLATIONS } from '@/lib/l10n/locales'
+import {
+  LocaleField,
+  LOCALE_COOKIE_EXPIRY,
+  LOCALE_COOKIE_KEY,
+  TEMP_TRANSLATIONS,
+} from '@/lib/l10n/locales'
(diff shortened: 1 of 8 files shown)


alebedev (approved):
Looks good!

robinandeer (commented):
Not sure but perhaps we should use the Next.js locale cookie for this as well?

https://nextjs.org/docs/advanced-features/i18n-routing#leveraging-the-next_locale-cookie

simonauner:
> Not sure but perhaps we should use the Next.js locale cookie for this as well?
> 
> https://nextjs.org/docs/advanced-features/i18n-routing#leveraging-the-next_locale-cookie

Yeah, that looks like a good idea. However, when setting `NEXT_LOCALE` to `se` nothing really happens, it's the exact same behaviour as with setting `HEDVIG_LOCALE` to `se` 🤔 

Some googling leads me to believe that it doesn't work, there are some SO threads about it and a GH issue: https://github.com/vercel/next.js/issues/22375

I have however set an expiry and changed it to `NEXT_LOCALE` now, maybe it works on the Edge? But since we have our middleware in place it's difficult to know if it's the middleware that does its job or Vercel's Edge logic that redirects... If anyone wants to experiment I guess it's just to remove `middleware.ts` temporarily and push on to a Preview deployment. Removing it locally only gives me a 404, so Next.js doesn't seem to use `NEXT_LOCALE` for any internal redirect.

simonauner:
Tried the above, doesn't seem to work ☝️ 

https://github.com/HedvigInsurance/racoon/pull/650

PR #651, opened 2022-10-04
Explore implementation of dynamic values

robinandeer:
## Describe your changes

Example of how we could support simple dynamic values in the price calculator forms.

One use case would be to support a "household size" field that submits to the backend as "numberCoInsured" after being transformed.

The reason to solve this in the form template solution is to give us more freedom to help out the editors to customize the form.

Otherwise it would be up to backend to add this support even if the request comes from a content editor.

### DIFF
--- a/apps/store/src/components/PriceCalculator/PriceCalculatorSection.tsx
+++ b/apps/store/src/components/PriceCalculator/PriceCalculatorSection.tsx
@@ -1,6 +1,6 @@
 import { FormEventHandler, ReactNode } from 'react'
 import { Button, Space } from 'ui'
-import { deserializeField } from '@/services/PriceCalculator/PriceCalculator.helpers'
+import { deserializeField } from '@/services/PriceCalculator/Field.helpers'
 import { FormSection, JSONData } from '@/services/PriceCalculator/PriceCalculator.types'
 import { useTranslateTextLabel } from './useTranslateTextLabel'
 
--- a/apps/store/src/services/PriceCalculator/Field.helpers.ts
+++ b/apps/store/src/services/PriceCalculator/Field.helpers.ts
@@ -1,0 +1,11 @@
+import { InputField, NumberField } from './Field.types'
+
+export const getIntegerField = (field: Omit<NumberField, 'type' | 'deserialize'>): NumberField => ({
+  ...field,
+  type: 'number',
+  deserialize: (value) => parseInt(value, 10),
+})
+
+export const deserializeField = (field: InputField, value: string) => {
+  return field.deserialize?.(value) ?? value
+}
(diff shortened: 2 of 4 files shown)


robinandeer:
* main
  * **PR #651** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/651" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/651?utm_source=stack-comment).

simonauner (approved):
Nice. This could be used to transform SSN fields as well, instead of having the more complex field that's being used now 👍

apps/store/src/services/PriceForm/Field.types.ts
```diff
@@ -6,6 +6,7 @@ type BaseField<ValueType> = {
   required?: boolean
   value?: ValueType
   defaultValue?: ValueType
+  transform?: (value: ValueType) => ValueType
```
guilhermespopolin, line 9: I thought in the exact same thing while doing some tests for [houseHold size input investigation](https://hedvig.atlassian.net/browse/GRW-1568) but my mistake was that I was trying to execute the transformation  inside form's handle submission function, which was not scalable.

guilhermespopolin (approved):

robinandeer (commented):
@guilhermespopolin yeah I wanna keep everything possible co-located

If that's very hard to achieve I think we might turn to backend tho :)
So we are thinking the same!

apps/store/src/services/PriceForm/PriceForm.helpers.ts
```diff
@@ -89,16 +89,13 @@ const calculateSectionState = (section: FormSection): FormSection => {
 
 export const deserializeField = (field: InputField, value: string) => {
   switch (field.type) {
-    case 'text':
-    case 'radio':
-    case 'select':
-    case 'date':
-      return value
-
     case 'number':
```
alebedev, line 92: We're checking transform at least twice.  Maybe we can do it before switch instead?

alebedev (approved):
Nice.  Simple and would get the job done

apps/store/src/services/PriceForm/Field.types.ts
```diff
@@ -6,6 +6,8 @@ type BaseField<ValueType> = {
   required?: boolean
   value?: ValueType
   defaultValue?: ValueType
+  deserialize?: (value: string) => ValueType
+  transform?: (value: ValueType) => ValueType
```
simonauner, line 10: `transform` isn't used in this new version, right?

guilhermespopolin: +1

guilhermespopolin (approved):

PR #652, opened 2022-10-04
GRW-1549 / Form template no house

amandaflod:
## Describe your changes

Add NO house template and extrabuildings options. Also added missing fields from SE House. 

[GRW-1549]: https://hedvig.atlassian.net/browse/GRW-1549?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/store/src/services/PriceForm/data/NO_HOUSE.ts
```diff
@@ -0,0 +1,161 @@
...
+            ],
+          },
+          layout: { columnSpan: 3 },
+        },
+        {
+          field: {
+            type: 'extra-buildings',
+            name: 'extraBuildings',
+            label: { key: 'Extra buildings' },
+            buildingOptions: [
```
alebedev, line 119: Discussion topic - do we need to keep options and their values in sync with the backend?  Are they the same or different per promise, per country?

Not a change for this PR, but let's plan the best way to handle this in the future and leave some notes

apps/store/src/components/PriceForm/AutomaticField.tsx
```diff
@@ -95,6 +95,16 @@ export const AutomaticField = ({ field, onSubmit, loading, autoFocus }: Props) =
       return <SsnSeField field={field} />
 
     case 'extra-buildings':
-      return <ExtraBuildingsField field={field} onSubmit={onSubmit} loading={loading} />
+      return (
+        <ExtraBuildingsField
+          field={field}
+          buildingOptions={field.buildingOptions.map((buildingOption) => ({
+            ...buildingOption,
+            name: translateLabel(buildingOption.label),
```
alebedev, line 103: Should we translate here?  Can we do it inside component to keep this case block shorter and ignorant of field types' innards?

simonauner, line 103: nit: Could this perhaps overwrite the `label` prop? Then we wouldn't need a new type for `buildingOptions` in ExtraBuildingsField.tsx perhaps, and could just reuse `FieldOption`?

alebedev (approved):
Well done!  Some nit comments inside

simonauner (approved):
Nice 🙌 

Got a bit confused by `label`/`name` conversion and the introduction of a new type. Maybe we can reuse `FieldOption`?

apps/store/src/components/PriceForm/AutomaticField.tsx
guilhermespopolin: +1 on translating `buildingOptions` in `ExtrBuildingsField`. Building options are included in `field` and `ExtrBuildingsField` component already make usage of `translateLabel` so it feels like a small effort with some gains

guilhermespopolin (approved):

apps/store/src/services/PriceForm/data/NO_HOUSE.ts
amandaflod: I think that is the plan! They will be different depending on promises and countries. Anything to add @robinandeer ?

apps/store/src/components/PriceForm/AutomaticField.tsx
amandaflod: That makes sense, I will fix that :)

amandaflod: Ok! I will fix that, will also add that change to the "select" type :) Thanks!

PR #654, opened 2022-10-04
Fix / Support published version preview

alebedev:
## Describe your changes

There's "Open published" button in storyblok that works through preview API.  Support it by detecting what version we're looking at from query params

- Fix crash in Storyblok
- Improve editor experience, no need to delete cookies to preview published version

### DIFF
--- a/apps/store/src/services/storyblok/storyblok.ts
+++ b/apps/store/src/services/storyblok/storyblok.ts
@@ -54,8 +54,10 @@ export type StoryblokPageProps = {
   globalStory: GlobalStory
 }
 
+export type StoryblokVersion = 'draft' | 'published'
+
 export type StoryblokPreviewData = {
-  version?: 'draft' | 'published'
+  version?: StoryblokVersion
 }
 
 export type StoryblokImage = {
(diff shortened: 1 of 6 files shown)


apps/store/src/services/storyblok/storyblok.ts
```diff
@@ -173,12 +177,12 @@ export const initStoryblok = () => {
 
 type StoryOptions = {
   locale: string
-  preview?: boolean
+  version?: 'draft' | 'published'
```
guilhermespopolin, line 180: It seems those 2 types share `version` field definition in a sense that if we change it in one we'd probably change in the other. With that I _nitly_ suggest explicitly tight them up.

```suggestion
  type StoryOptions = { locale: string } & Pick<StoryblokPreviewData, 'version'> // or StoryblokPreviewData['version'] I guess
```

guilhermespopolin (approved):

apps/store/src/services/storyblok/storyblok.ts
alebedev: Good point.  I extracted StoryblokVersion type to make it explicit

PR #655, opened 2022-10-04
styles(Accordion): added focus ring

guilhermespopolin:
## Describe your changes

It adds focus ring for `Accordion` trigger

Screenshot 2022-10-04 at 14 05 59

### DIFF
--- a/apps/store/src/components/PriceForm/Accordion.tsx
+++ b/apps/store/src/components/PriceForm/Accordion.tsx
@@ -22,6 +22,13 @@ export const Content = styled(Accordion.Content)(({ theme }) => ({
   paddingTop: theme.space[0],
 }))
 
+const StyledTrigger = styled(Accordion.Trigger)(({ theme }) => ({
+  '&:focus-visible': {
+    outline: `2px solid ${theme.colors.gray900}`,
+    outlineOffset: 2,
+  },
+}))
+
 export const Root = Accordion.Root
 export const Item = Accordion.Item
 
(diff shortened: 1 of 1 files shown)


gustaveen:
Nice! 🙌 I wonder how we could style this to harmonise with the design. It almost looks like a checkbox with a square around it :) Because this would show up on click as well right?

guilhermespopolin:
> Nice! 🙌 I wonder how we could style this to harmonise with the design. It almost looks like a checkbox with a square around it :) Because this would show up on click as well right?
That's true, it does look like a checkbox 😅
It would only appear by focusing via keyboard tho

apps/store/src/components/PriceForm/Accordion.tsx
```diff
@@ -22,6 +22,13 @@ export const Content = styled(Accordion.Content)(({ theme }) => ({
   paddingTop: theme.space[0],
 }))
 
+const StyledTrigger = styled(Accordion.Trigger)(({ theme }) => ({
+  '&:focus-visible': {
+    outline: `2px solid ${theme.colors.gray900}`,
```
alebedev, line 27: Let's ask design team, maybe we want to use some highlight color here

guilhermespopolin: So it seems any designs in that direction is [gonna take some time to get done](https://hedviginsurance.slack.com/archives/C041SD67G82/p1664895741834449). Should we close this PR them? 🤔

simonauner: I would argue the counterpart @guilhermespopolin . Merge first and then we can tweak later. Accessibility is a feature 🚀

PR #661, opened 2022-10-05
Fix / Allow require() calls when linting NodeJs files

alebedev:
## Describe your changes

Overwrite eslint config to recognise JS files as Node environment
Stop complaining about require() calls there


simonauner (approved):
gif

PR #663, opened 2022-10-06
GRW-1587 - refact(initializeApollo): updated functions signature to accept a configuration object

guilhermespopolin:
## Describe your changes

Updated `initializeApollo` function signature to accept a configuration object instead of a list of arguments

It simplifies its usage

## Jira issue(s): [GRW-1587](https://hedvig.atlassian.net/browse/GRW-1587)

### DIFF
--- a/apps/store/src/services/apollo/client.ts
+++ b/apps/store/src/services/apollo/client.ts
@@ -49,9 +49,11 @@ type InitializeApolloParams = {
   res?: GetServerSidePropsContext['res']
 }
 
-export const initializeApollo = (
-  { initialState, req, res }: InitializeApolloParams = { initialState: null },
-) => {
+export const initializeApollo = ({
+  initialState = null,
+  req,
+  res,
+}: InitializeApolloParams = {}) => {
   const _apolloClient = apolloClient ?? createApolloClient(Auth.getAccessToken(req, res))
 
   if (initialState) {
--- a/apps/store/src/services/apollo/client.ts
+++ b/apps/store/src/services/apollo/client.ts
@@ -43,10 +43,14 @@ const createApolloClient = (accessToken?: string) => {
   })
 }
 
+type InitializeApolloParams = {
+  initialState?: unknown
+  req?: GetServerSidePropsContext['req']
+  res?: GetServerSidePropsContext['res']
+}
+
 export const initializeApollo = (
-  initialState: unknown = null,
-  req?: GetServerSidePropsContext['req'],
-  res?: GetServerSidePropsContext['res'],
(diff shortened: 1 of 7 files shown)


apps/store/src/services/apollo/client.ts
```diff
@@ -43,10 +43,14 @@ const createApolloClient = (accessToken?: string) => {
...
+  initialState?: unknown
+  req?: GetServerSidePropsContext['req']
+  res?: GetServerSidePropsContext['res']
+}
+
 export const initializeApollo = (
-  initialState: unknown = null,
-  req?: GetServerSidePropsContext['req'],
-  res?: GetServerSidePropsContext['res'],
+  { initialState, req, res }: InitializeApolloParams = { initialState: null },
```
robinandeer, line 53: nit: this should do it, no?

```suggestion
  { initialState = null, req, res }: InitializeApolloParams,
```

robinandeer (approved):
Great stuff!

apps/store/src/services/apollo/client.ts
guilhermespopolin: Right?! I also thought the same but it fails. Trying to figure it out why..

guilhermespopolin: Now I understand. [`useApollo`](https://github.com/HedvigInsurance/racoon/blob/main/apps/store/src/services/apollo/client.ts#L70) call `initializeApollo` passing in something that can be undefined (in fact in will be `undefined` at some point that's why the app breaks). That means `initializeApollo` should have its signature changed again since it should be able to be called without passing in any argument. With that said I can see 2 options:

1. `export const initializeApollo = ({ intialState = null, req, res }: InitializeApolloParams = {}) => {...}`
2.  `export const initializeApollo = ({ initialState, req, res }: InitializeApolloParams = { initialState: null }) => {..}`

In both cases we'd need to define a default value for `intializeApollo` because that would be the away to [tell TS that that argument is optional.](https://bobbyhadz.com/blog/typescript-parameter-cannot-have-question-mark-and-initializer#:~:text=in%20TS%20%23-,To%20solve%20the%20%22Parameter%20cannot%20have%20question%20mark%20and%20initializer%22%20error%20in%20TypeScript%2C%20don%27t%20use%20a%20question%20mark%20to%20set%20parameters%20to%20optional%20when%20using%20an%20equal%20sign%20to%20provide%20a%20default%20value.%20Parameters%20with%20default%20values%20are%20implied%20to%20be%20optional.,-Here%20is%20an)
I guess I prefer option 1. Do you have any preference?

PR #665, opened 2022-10-06
Add multi-root VSCode workspace

robinandeer:
## Describe your changes

Add multi-root VSCode workspace: https://code.visualstudio.com/docs/editor/multi-root-workspaces

Add setup to make vscode-jest work as expected!

Move settings inside workspace file.

To make vscode-test work.

### DIFF
--- a/.vscode/racoon.code-workspace
+++ b/.vscode/racoon.code-workspace
@@ -1,0 +1,33 @@
+{
+  "folders": [
+    {
+      "name": "ROOT",
+      "path": "../"
+    },
+    {
+      "name": "apps/onboarding",
+      "path": "../apps/onboarding"
+    },
+    {
+      "name": "apps/store",
+      "path": "../apps/store"
+    },
+    {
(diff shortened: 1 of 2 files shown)


gustaveen (commented):
Just looking at the documentation, it sounds resonable for a monorepo https://code.visualstudio.com/docs/editor/workspaces 

What cons could it bring? Couldn't figure out that bit from the docs :)

robinandeer (commented):
@gustaveen I think it more has to do with committing editor setup to git

Perhaps someone wants to setup their editor differently and then this would become a problem

gustaveen:
> @gustaveen I think it more has to do with committing editor setup to git
> 
> Perhaps someone wants to setup their editor differently and then this would become a problem

Gotcha! I have to little knowledge on the topic, but it seems that if you add your own `.vscode/settings.json` file you can still override the workspace settings if you like :)

gustaveen (approved):
Let's try it! 🙌

PR #669, opened 2022-10-06
GRW-1566 / Integrate Intercom

amandaflod:
## Describe your changes 
Integrate Intercom chat in Customer Support block, by using react-use-intercom (https://github.com/devrnt/react-use-intercom). 

[GRW-1566]: https://hedvig.atlassian.net/browse/GRW-1566?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/store/src/components/ContactSupport/ContactSupport.tsx
```diff
@@ -8,11 +9,9 @@ export type ContactSupportProps = {
   availabilityText?: string
 }
```
alebedev, line 11: Can you reorder components and filters in some logically consistent way?  I don't remember if we have a convention for ordering

apps/store/src/components/ContactSupport/ContactSupport.tsx
```diff
@@ -39,6 +38,32 @@ export const ContactSupport = ({
...
+  if (appId) return appId
+  throw new Error('Expected env variable INTERCOM_APP_ID to be defined')
+}
+
+export const ContactSupport = ({
+  title,
+  showCallButton,
+  availabilityText,
+}: ContactSupportProps) => {
+  return (
```
alebedev, line 52: If there a reason to separate provider and usage here?  Instead of having `ContactSupport` + `ContactSupportInner` you can have `ContactSupport` and `IntercomChatButton` - does it look cleaner?

package.json
```diff
@@ -39,5 +39,8 @@
     "lint-staged": "13.0.3",
     "prettier": "2.7.1",
     "turbo": "1.5.5"
+  },
+  "dependencies": {
+    "react-use-intercom": "2.0.0"
   }
```
robinandeer, line 45: I think you accidentally added this as a global repo dependency ;)

use `yarn workspace store add react-use-intercom@2.0.0`

apps/store/src/components/ContactSupport/ContactSupport.tsx
amandaflod: I tidied it up with the `IntercomChatButton` fix.

alebedev (approved):
Good job!

apps/store/src/components/ContactSupport/ContactSupport.tsx
```diff
@@ -1,7 +1,24 @@
 import styled from '@emotion/styled'
+import { IntercomProvider, useIntercom } from 'react-use-intercom'
 import { Heading, Button, Space } from 'ui'
 import { SpaceFlex } from '@/components/SpaceFlex/SpaceFlex'
 
+const getIntercomAppId = (): string => {
+  const appId = process.env.NEXT_PUBLIC_INTERCOM_APP_ID
+  if (appId) return appId
+  throw new Error('Expected env variable INTERCOM_APP_ID to be defined')
```
robinandeer, line 9: Hmmm I'm a little worried about throwing this during runtime

Either we should check early (outside React) and make it a required ENV VAR.

Or we should handle it gracefully without throwing an exception.

amandaflod: I've refactored the code to handle the error gracefully by disabling the button if the env variable is not defined.

Since env variables can be defined when the application starts I think our only option is to handle the error gracefully. The alternative seems to be to throw a runtime error when the application starts. What do you think? :)

alebedev: It would be nice to disable integration if something breaks on Intercom side.  In that case disabled/hidden button is exactly what we need.

Having empty APP_ID would do it, even if not 100% explicitly

apps/store/src/components/ContactSupport/IntercomChatButton.tsx
```diff
@@ -0,0 +1,41 @@
...
+}
+
+export const IntercomChatButton = () => {
+  const appId = process.env.NEXT_PUBLIC_INTERCOM_APP_ID
+
+  useEffect(() => {
+    if (!appId) {
+      datadogLogs.logger.warn('Expected env variable INTERCOM_APP_ID to be defined')
+    }
+  }, [appId])
```
robinandeer, line 28: Let's move all this outside React :)

We should perhaps just log this during build time... 🤔 we don't need to log it everytime someone loads the app.

robinandeer (approved):
Looks good

I would only suggest we move the app ID checkout outside React -- we only need to read and log once. But feel free to refactor in a new PR :)

apps/store/src/components/ContactSupport/IntercomChatButton.tsx
amandaflod: Ok let's circle back to this later. When we do so, I think we should lift out all env variable checks :)

PR #672, opened 2022-10-07
Implement device id header

robinandeer:
## Describe your changes

Implement the device ID setup that we have on Web Onboarding.

Bascially we create a unique UUID that we store in the browser.
Then we send this as a header with all requests to the backend.

I decided to link this to the Apollo Client setup to ensure that the UUID is created before we send any API requests.

I talked with Aleks about collecting things like this in a `LocalContext` -- let's iterate on exactly how it should work!

The purpose with the header is to link a `ShopSession` with a `PaymentConnection` on the backend.

[GRW-1609]: https://hedvig.atlassian.net/browse/GRW-1609?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/store/package.json
+++ b/apps/store/package.json
@@ -31,6 +31,7 @@
     "@radix-ui/react-switch": "1.0.0",
     "@radix-ui/react-tabs": "1.0.0",
     "@storyblok/react": "1.3.2",
+    "@types/uuid": "8.3.4",
     "ajv": "8.11.0",
     "cookies-next": "2.1.1",
     "dd-trace": "3.3.1",
@@ -45,7 +46,8 @@
     "pino-pretty": "9.1.0",
     "react": "18.2.0",
     "react-dom": "18.2.0",
-    "ui": "workspace:"
+    "ui": "workspace:",
+    "uuid": "9.0.0"
   },
   "devDependencies": {
     "@babel/core": "7.19.3",
--- a/apps/store/src/services/Auth/Auth.ts
+++ b/apps/store/src/services/Auth/Auth.ts
@@ -1,5 +1,5 @@
 import { getCookie, setCookie } from 'cookies-next'
-import { GetServerSidePropsContext } from 'next'
+import { OptionsType } from 'cookies-next/lib/types'
 
 const AUTH_COOKIE_KEY = 'HEDVIG_ACCESS_TOKEN'
 const MAX_AGE = 60 * 60 * 24 // 24 hours
@@ -13,10 +13,15 @@ export const save = (accessToken: string) => {
   })
 }
 
-export const getAccessToken = (
-  req?: GetServerSidePropsContext['req'],
-  res?: GetServerSidePropsContext['res'],
-) => {
+type CookieParams = Pick<OptionsType, 'req' | 'res'>
+
+const getAccessToken = ({ req, res }: CookieParams) => {
   const value = getCookie(AUTH_COOKIE_KEY, { req, res })
   return typeof value === 'string' ? value : undefined
+}
+
+export const getAuthHeader = (params: CookieParams = {}): Record<string, string> => {
+  const accessToken = getAccessToken(params)
+  if (accessToken) return { authorization: accessToken }
+  return {}
 }
--- a/apps/store/src/services/LocalContext/LocalContext.helpers.ts
+++ b/apps/store/src/services/LocalContext/LocalContext.helpers.ts
@@ -1,0 +1,28 @@
+import { getCookie, setCookie } from 'cookies-next'
+import { OptionsType } from 'cookies-next/lib/types'
+import { v4 as uuidv4 } from 'uuid'
+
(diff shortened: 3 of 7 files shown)


robinandeer:
* main
  * **PR #665** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/665" target="_blank">Graphite</a> 
    * **PR #672** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/672" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/672?utm_source=stack-comment).

apps/store/package.json
```diff
@@ -45,7 +46,8 @@
     "pino-pretty": "9.1.0",
     "react": "18.2.0",
     "react-dom": "18.2.0",
-    "ui": "workspace:"
+    "ui": "workspace:",
+    "uuid": "9.0.0"
```
robinandeer, line 50: I tried using the `crypto` module but it's not supported by >90% or browsers so had to add this dependency.

That would not be too bad but it will get bundled client-side which is not great but perhaps we should optimize later.

alebedev: It's probably tiny module anyway

apps/store/src/services/Auth/Auth.ts
```diff
@@ -13,10 +13,16 @@ export const save = (accessToken: string) => {
   })
 }
 
-export const getAccessToken = (
-  req?: GetServerSidePropsContext['req'],
-  res?: GetServerSidePropsContext['res'],
-) => {
+const getAccessToken = (req?: OptionsType['req'], res?: OptionsType['res']) => {
   const value = getCookie(AUTH_COOKIE_KEY, { req, res })
   return typeof value === 'string' ? value : undefined
```
alebedev, line 18: Is this line doing anything?  I'd expect value to be undefined already when it's not a string

apps/store/src/services/LocalContext/LocalContext.helpers.ts
```diff
@@ -0,0 +1,25 @@
+import { getCookie, setCookie } from 'cookies-next'
+import { OptionsType } from 'cookies-next/lib/types'
+import { v4 as uuidv4 } from 'uuid'
+
+const DEVICE_ID_COOKIE_KEY = '_hv_device_id'
+
+const getOrCreateDeviceId = (req?: OptionsType['req'], res?: OptionsType['res']) => {
```
alebedev, line 7: Would it make sense to use same object argument as in Auth module?

apps/store/src/services/LocalContext/LocalContext.helpers.ts
```diff
@@ -0,0 +1,25 @@
...
+  const currentDeviceId = getDeviceId(req, res)
+  if (currentDeviceId) return currentDeviceId
+
+  const newDeviceId = uuidv4()
+  setCookie(DEVICE_ID_COOKIE_KEY, newDeviceId, { req, res })
```
alebedev, line 12: Do we want long lifetime for that cookie?  It's a session cookie by default, AFAIK

alebedev (approved):
Looks good

apps/store/src/services/Auth/Auth.ts
robinandeer: haha yes this is very confusing but seems like the cookie-lib does some parsing so the value is possibly one of multiple types but we only care about strings :)

apps/store/src/services/LocalContext/LocalContext.helpers.ts
robinandeer: ahh true that, I was maybe lazy

apps/store/src/services/LocalContext/LocalContext.helpers.ts
robinandeer: oh damn, yes we sure want it to be -- I'll update that!

guilhermespopolin (approved):
> The purpose with the header is to link a ShopSession with a PaymentConnection on the backend.

It might be a silly question but I'm gonna ask anyway 😅 : Why do we want to be able to link a ShopSession with a PaymentConnection in the BE?

robinandeer (commented):
@guilhermespopolin we want to link them so that when the shop session is signed we can connect the new member to the correct payment details :)

We don't want to link the payment directly to the shop session to be able to use it for other use cases -- for example to directly link payment to a member!

robinandeer:
**Graphite Merge Job**

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/672?utm_source=stack-comment).


Job Reference: sKYQaJwroXIqUEDaDs7Z

PR #680, opened 2022-10-10
Add callback to web-onboarding payment URL

robinandeer:
## Describe your changes

Add callback URL as query parameter `redirect_url` that is passed to Web Onboarding to connect payments.

We need a way for Web Onboarding to communicate back to Racoon Store after the payment connection flow is complete.

We don't have designs for the fail case where the payment was not connected - will ask for designs for this :)

### DIFF
--- a/apps/store/src/pages/checkout/payment/[status].tsx
+++ b/apps/store/src/pages/checkout/payment/[status].tsx
@@ -1,0 +1,64 @@
+import type { GetServerSideProps, NextPage } from 'next'
+import Link from 'next/link'
+import { isRoutingLocale } from '@/lib/l10n/localeUtils'
+import { PageLink } from '@/lib/PageLink'
+import { initializeApollo } from '@/services/apollo/client'
+import { PaymentConnectionFlow } from '@/services/apollo/generated'
+import logger from '@/services/logger/server'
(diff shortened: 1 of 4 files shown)


apps/store/src/pages/checkout/payment/[status].tsx
```diff
@@ -0,0 +1,64 @@
...
+import { PageLink } from '@/lib/PageLink'
+import { initializeApollo } from '@/services/apollo/client'
+import { PaymentConnectionFlow } from '@/services/apollo/generated'
+import logger from '@/services/logger/server'
+import { getCurrentShopSessionServerSide } from '@/services/shopSession/ShopSession.helpers'
+
+const PaymentRedirectPage: NextPage = () => {
+  return (
+    <div>
+      Something went wrong. Please <Link href={PageLink.checkoutPayment()}>try again</Link>.
```
Comment (alebedev) [inline apps/store/src/pages/checkout/payment/[status].tsx:13]: Nit: how would we localize inline links in the future?  Could make it a separate button to avoid the issue

alebedev (approved):
Looks good

apps/store/src/pages/checkout/payment/[status].tsx
Comment (robinandeer) [reply apps/store/src/pages/checkout/payment/[status].tsx:13]: Yeah that would probably be a good idea

The checkout shouldn't be too customizable :)

PR #754, opened 2022-10-26
GRW-1693 / Feature / Canonical link and alternates

alebedev:
## Describe your changes

Add CMS-defined page metadata
- link rel=canonical for A/B testing
- link rel=alternate for localized variants

SEO, make Google happy

[GRW-1693]: https://hedvig.atlassian.net/browse/GRW-1693?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/store/src/components/HeadSeoInfo/HeadSeoInfo.tsx
+++ b/apps/store/src/components/HeadSeoInfo/HeadSeoInfo.tsx
@@ -1,6 +1,7 @@
 import { StoryData } from '@storyblok/react'
 import Head from 'next/head'
-import { isRoutingLocale } from '@/utils/l10n/localeUtils'
+import { isRoutingLocale, toIsoLocale } from '@/utils/l10n/localeUtils'
+import { ORIGIN_URL } from '@/utils/PageLink'
 
 export const HeadSeoInfo = ({ story }: { story: StoryData }) => {
   // AB testing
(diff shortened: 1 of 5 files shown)


apps/store/src/components/HeadSeoInfo/HeadSeoInfo.tsx
```diff
@@ -0,0 +1,35 @@
...
+      {alternates.map((alternate) => (
+        <AlternateLink key={alternate.id} fullSlug={alternate.full_slug} />
+      ))}
+    </>
+  )
+}
+
+const AlternateLink = ({ fullSlug }: { fullSlug: string }) => {
+  return (
+    <Head>
```
robinandeer, line 26: Why this separate head component? We could just place inside the head above, no?

robinandeer (approved):
Looks good! Just had a minor question :)

apps/store/src/components/HeadSeoInfo/HeadSeoInfo.tsx
alebedev: Any links has to be direct children of next/head or they are ignored.  So, when we want to structure our code nicely, we end up with many Head wrappers.  This is intended usage pattern

apps/store/src/components/HeadSeoInfo/HeadSeoInfo.tsx
```diff
@@ -0,0 +1,35 @@
...
+      <link rel="alternate" href={`/${fullSlug}`} hrefLang={getHrefLang(fullSlug)} />
+    </Head>
+  )
+}
+
+const getHrefLang = (fullSlug: string) => {
+  const slugLocale = fullSlug.split('/')[0]
+  return isRoutingLocale(slugLocale) ? slugLocale : 'x-default'
```
gustaveen, line 34: if I understood Peter,`x-default` should only be rendered on `hedvig.com` the market selection page right? Any risk it would be outputted in more places?

gustaveen (approved):
Looks about right! 🙌
Just a question on `x-default`

apps/store/src/components/HeadSeoInfo/HeadSeoInfo.tsx
alebedev: This and possible CMS misconfiguration.  I don't see any other cases

PR #760, opened 2022-10-27
GRW-1695 / Feature / GTM metadata fixes

alebedev:
## Describe your changes

- Fix inital virtual_page_view tracking
- Change the way we provide metadata so Google Tag Manager can process it correctly
- Add siteVersion context for future A/B testing data

### DIFF
--- a/apps/store/src/pages/_app.tsx
+++ b/apps/store/src/pages/_app.tsx
@@ -4,7 +4,7 @@ import type { AppPropsWithLayout } from 'next/app'
 import Head from 'next/head'
 import { ThemeProvider } from 'ui'
 import { useApollo } from '@/services/apollo/client'
-import { GTMAppScript, useGTMEvents } from '@/services/gtm'
+import { GTMAppScript, trackPageView, useGTMEvents } from '@/services/gtm'
 import * as Datadog from '@/services/logger/client'
 import { SHOP_SESSION_PROP_NAME } from '@/services/shopSession/ShopSession.constants'
 import { ShopSessionProvider } from '@/services/shopSession/ShopSessionContext'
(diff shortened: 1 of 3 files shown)


apps/store/src/services/gtm.tsx
```diff
@@ -120,3 +121,14 @@ export const useGTMEvents = () => {
     }
   }, [router.events])
 }
+
+const trackPageView = (url: string) => {
+  console.debug('pageview', url)
```
robinandeer, line 126: Do we want to use console here or always datadog logs?

apps/store/src/services/gtm.tsx
```diff
@@ -95,22 +96,22 @@ export const useGTMEvents = () => {
...
+        siteVersion: 'racoon',
         country: countryCode,
+        environment: getGtmEnvironment(),
       },
     })
   }, [countryCode])
 
+  useEffect(() => {
+    trackPageView(window.location.pathname)
+  }, [])
```
robinandeer, line 110: Shouldn't this be run outside React?

```
if (typeof window === 'undefined') { ... }
```

apps/store/src/services/gtm.tsx
alebedev: Intentionally not sending to Datadog here, since we'll be sending that data to Analytics

alebedev: I think it's benign and can stay without harming us.  For example, JIRA outputs lots of debug logs in production and that doesn't seem to be a problem

But let's figure this out closer to launch

apps/store/src/services/gtm.tsx
alebedev: Just enabling it didn't work, created a ticket to investigate and fix in the future

apps/store/src/pages/_app.tsx
```diff
@@ -17,6 +17,8 @@ if (process.env.NEXT_PUBLIC_API_MOCKING === 'enabled') {
 
 if (typeof window === 'undefined') {
   Datadog.initDatadog()
```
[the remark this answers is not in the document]
alebedev: Won't work, created backlog task to fix this

PR #768, opened 2022-10-28
GRW-1695 / Feature / Track AB redirects

alebedev:
## Describe your changes

Send experiment_impression events so that AB redirects are visible in Google Optimize and Analytics

Our first ever redirect data in Optimize:
Screenshot 2022-10-28 at 09 33 19

First step to conversion tracking between old and new sites

### DIFF
--- a/apps/store/src/utils/abRedirects.ts
+++ b/apps/store/src/utils/abRedirects.ts
@@ -1,0 +1,31 @@
+import { getCookies } from 'cookies-next'
+
+type AbRedirect = {
+  // Use in env variables
+  id: string
+  // Take this from Google Optimize
+  optimizeExperimentId: string
+  sourcePath: string
+  redirectPath: string
+}
+
+// Needs to be in sync with market-web, redirects start there
+export const AB_REDIRECTS: AbRedirect[] = [
+  {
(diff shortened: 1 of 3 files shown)


apps/store/src/utils/abRedirects.ts
```diff
@@ -0,0 +1,31 @@
...
+  redirectPath: string
+}
+
+// Needs to be in sync with market-web, redirects start there
+export const AB_REDIRECTS: AbRedirect[] = [
+  {
+    id: 'SE_INDEX',
+    optimizeExperimentId: 'L2us-_SfSueXuS-p8uyV3A',
+    sourcePath: '/se',
+    redirectPath: '/se',
```
gustaveen, line 18: So we're configuring here it's a redirect from hedvig.com/se -> to [new-hedvig.com]/se? @alebedev

gustaveen (approved):
LGTM! Awesome you get that data flowing! 🙌

apps/store/src/utils/abRedirects.ts
alebedev: Yes, but we don't need to know either domain at runtime, therefore only paths are here

PR #789, opened 2022-10-31
GRW-1707 / Feature / Track offer_created event

alebedev:
## Describe your changes

Quick and dirty support for offer_created event tracking
- Send event via GTM
- Include very basic metadata
- Some TODOs for future improvement of tracking

Screenshot of event from GTM test page:
Screenshot 2022-10-31 at 12 57 29

We want to track conversion somehow when A/B testing, this adds first conversion event

### DIFF
--- a/apps/store/src/components/PriceForm/PriceForm.tsx
+++ b/apps/store/src/components/PriceForm/PriceForm.tsx
@@ -3,9 +3,11 @@ import {
   PriceIntentFragmentFragment,
   usePriceIntentConfirmMutation,
 } from '@/services/apollo/generated'
+import { trackOffer } from '@/services/gtm'
 import { prefillData, updateFormState } from '@/services/PriceForm/PriceForm.helpers'
 import { Form } from '@/services/PriceForm/PriceForm.types'
 import { PriceIntent } from '@/services/priceIntent/priceIntent.types'
+import { useShopSession } from '@/services/shopSession/ShopSessionContext'
 import { AutomaticField } from './AutomaticField'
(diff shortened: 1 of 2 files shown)


apps/store/src/components/PriceForm/PriceForm.tsx
```diff
@@ -21,11 +23,20 @@ type Props = {
...
     onCompleted(data) {
       const updatedPriceIntent = data.priceIntentConfirm.priceIntent
       if (updatedPriceIntent) {
+        // FIXME: pick offer for specific product or track all offers
+        const firstOffer = updatedPriceIntent.offers[0]
+        trackOffer({
+          shopSessionId: shopSession!.id,
+          contractType: firstOffer.variant.typeOfContract,
+          amount: firstOffer.price.amount,
+          currency: firstOffer.price.currencyCode,
```
gustaveen, line 38: nit: do we want to send the offer as a single payload and extract the data in the trackOffer function instead? Guess it comes down to if we we reuse this in more places I guess

gustaveen (approved):
Great first step! 🙌

apps/store/src/components/PriceForm/PriceForm.tsx
alebedev: There's a place for smart tracking service between UI and GTM service. Agree that UI should not know which details we're using in tracking

amandaflod (approved):
Looks good to me! Great progress for tracking in hedvig.com 👍

PR #797, opened 2022-11-02
GRW-1728 / Migrate to new image component

gustaveen:
## Describe your changes
Use the new Image component from Next

The new Image component relies on [native lazy loading](https://developer.mozilla.org/en-US/docs/Web/Performance/Lazy_loading) and outputs much less html which makes it easier to style.

You can read more about the Image component here https://nextjs.org/docs/api-reference/next/image

What we need to consider if we have good enough browser support https://caniuse.com/loading-lazy-attr since it's not supported in Safari below 15.4 It does fallback to eager loading. 

Looking at the split between browser version on our site from 1/9 we have 1,4% sitting on version 14.1.2 and with other versions below 15.4 it's a total around 2,4%. Given that we will go live next year the number will decrease so I think we can go ahead and use it. For onboarding it might not be worthwhile to migrate.

Screenshot 2022-11-02 at 17 31 07

### Html before 
Screenshot 2022-11-02 at 17 18 52

### Html after
Screenshot 2022-11-02 at 17 18 18

[GRW-1728]: https://hedvig.atlassian.net/browse/GRW-1728?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/store/src/blocks/HeroBlock.tsx
+++ b/apps/store/src/blocks/HeroBlock.tsx
@@ -1,6 +1,6 @@
 import styled from '@emotion/styled'
 import { storyblokEditable } from '@storyblok/react'
-import Image from 'next/legacy/image'
+import Image from 'next/image'
 import { ButtonBlock, ButtonBlockProps } from '@/blocks/ButtonBlock'
 import { HeadingBlock, HeadingBlockProps } from '@/blocks/HeadingBlock'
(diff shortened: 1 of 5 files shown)


KarnellSchultz (approved):
nice!

alebedev (approved):
Nice!x

PR #819, opened 2022-11-07
Grw-1702 / feat / PDP: Navigation and Tabs menu sticky together

KarnellSchultz:
## Describe your changes

Add the ability for the nav and the tab menu on PDP to scroll nicely with one another. 

The below video shows the scroll. 

https://user-images.githubusercontent.com/50870173/200324103-df8968bb-b1dd-4072-bae9-adf191371136.mov

This is how you'd expect two sticky position elements to scroll with one another. With one element pushing the other off of the page rather than going behind it. 

https://hedvig-dot-jregh850j-hedvig.vercel.app/se/products/bil

### The reason this is not done with css
For position sticky to work with these two components they'd need to have the same parent container. To do that we'd need to make major changes to our application's markup. Making those changes would come with other issues that are really not worth it / not acceptable. 

### DIFF
--- a/apps/onboarding/package.json
+++ b/apps/onboarding/package.json
@@ -33,7 +33,7 @@
     "graphql-tag": "2.12.6",
     "jsdom": "20.0.2",
     "lodash-es": "4.17.21",
-    "marked": "4.2.1",
+    "marked": "4.2.2",
     "next": "13.0.1",
     "next-i18next": "12.1.0",
     "pino-datadog": "https://github.com/simonauner/pino-datadog#rename-hostname-to-host",
--- a/yarn.lock
+++ b/yarn.lock
@@ -16381,12 +16381,12 @@ __metadata:
   languageName: node
   linkType: hard
 
-"marked@npm:4.2.1":
-  version: 4.2.1
-  resolution: "marked@npm:4.2.1"
+"marked@npm:4.2.2":
+  version: 4.2.2
+  resolution: "marked@npm:4.2.2"
   bin:
     marked: bin/marked.js
-  checksum: 9361bc017f707ed4233ba0cade7065269b2f835d29388bfa170090ebeadfe96f01097b5495b2514f79c1ce13042007d64cc156236cabb841a4e470af977af04d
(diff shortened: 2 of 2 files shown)


apps/store/src/blocks/ImageBlock.tsx
```diff
@@ -32,6 +32,7 @@ const Wrapper = styled.div<{ margins: boolean }>(({ theme, margins = true }) =>
   paddingLeft: margins ? theme.space[4] : 0,
   paddingRight: margins ? theme.space[4] : 0,
   position: 'relative',
+  zIndex: -1,
```
gustaveen, line 35: Should we isolate the zIndex settings to the Tablists instead? Just thinking the ImageBlock or other blocks shouldn't have to know about this feature :)

apps/store/src/components/TopMenu/useTopMenuStickyScroll.tsx
```diff
@@ -0,0 +1,36 @@
...
+  return scrollPosition
+}
+
+export const useTopMenuStickyScroll = () => {
+  const [state, setState] = useState({ distanceFromTop: 0, elementHeight: 0 })
+  const scrollPosition = useScrollPosition()
+
+  useEffect(() => {
+    const tabListElement = document?.getElementById('tab-list')
+    const rect = tabListElement?.getBoundingClientRect()
```
gustaveen, line 23: I was a bit curious if we can use the [Intersection Observer API](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API) for this to make it a bit cheaper on the main thread. But from what I can tell it isn't the optimal use case to update a pixel value on an element like that. Did you find any examples using that API @KarnellSchultz :) 

The current implementation works really nice when testing! ⭐️

KarnellSchultz: @gustaveen  Good call, after looking into Intersection observer that seems to be the way to go. Found some interesting things [about performance](https://itnext.io/1v1-scroll-listener-vs-intersection-observers-469a26ab9eb6) too.

I'll make the swap. 

As for incrementing the pixel. I can't think of another way to move the nav off the view port 🤔

apps/store/src/blocks/ImageBlock.tsx
KarnellSchultz: Without this line the images were in front of the Tablist when scrolling. But let's see if we can sort this out another way.

PR #820, opened 2022-11-07
Co-locate top level z-index values

gustaveen:
## Describe your changes
Co-locate top level z-index values

An attempt to avoid arbitrary z-index values that grows over time. There's a bunch of ways to do this but I kinda liked this one: https://www.nielskrijger.com/posts/2021-08-29/zindex-with-css-in-js/ This approach sets z-index relative to each other, rather than just having an object with hardcoded values. It makes it easy to add values in between existing z-index values.


KarnellSchultz (approved):
Into it 👍

PR #826, opened 2022-11-08
GRW-1702 / feat /pdp: Intersection observer nav play nice with tabs

KarnellSchultz:
## Describe your changes

The Topmenu (aka the nav) now functions as if it was sticky with the tabs menu on the product detail page. 

https://user-images.githubusercontent.com/50870173/200620979-58f52389-8665-4b66-b163-341972aa8a0c.mov

The design for the product page show that this Tabs menu needs to be sticky and replace the normal nav menu in it's sticky position.

[Intersection Observer](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API) was used here instead of watching the users scroll to improve the performance of the changes.  

### downsides

Scrolling really fast can cause issues where the top menu does not move totally off the page. 
Read more about that [here](https://stackoverflow.com/questions/61951380/intersection-observer-fails-sometimes-when-i-scroll-fast)


KarnellSchultz:
Not working in the preview this morning 🤔

apps/store/src/components/TopMenu/useTopOffset.tsx
```diff
@@ -0,0 +1,38 @@
+import { useScroll } from 'framer-motion'
+import { useEffect, useRef, useState } from 'react'
+
+export const useTopOffset = () => {
```
gustaveen, line 4: Hm, is it just me that don't see it, where is this hook being used? :)

KarnellSchultz: good catch, It's supposed to be deleted :-)

gustaveen:
I think we can adress the zIndex with the following :D 

```ts
export const Tabs = styled(RadixTabs.Root)({
  position: 'relative', // Set a new stacking context
  display: 'flex',
  flexDirection: 'column',
})

export const TabsList = styled(RadixTabs.TabsList)(({ theme }) => ({
  display: 'flex',
  borderBottom: `1px solid ${theme.colors.gray500}`,
  background: theme.colors.white,
  position: 'sticky',
  top: 0,
  zIndex: zIndexes.tabs, // 'tabs' can be just above body I think
}))
```

apps/store/src/components/TopMenu/useTabListObserver.tsx
```diff
@@ -0,0 +1,49 @@
+import { useEffect, useState } from 'react'
+
+const buildThresholdList = (steps: number) => {
```
gustaveen, line 3: If feels like Intersection Observer isn't the best use case for sequential updates like this. From what I understood now it isn't recommended to use threshold for observing scroll changes

https://css-tricks.com/an-explanation-of-how-the-intersection-observer-watches/#:~:text=I%20don%E2%80%99t%20recommend%20you%20set%20your%20thresholds%20in%20this%20way%20for%20typical%20use%20in%20projects.

gustaveen (commented):
I tried it out on my phone and it lags behind a bit by the look of it. Not sure Intersection Observer is suited for this type of sequential calculations :/ (that's on me since I brought it up in the first place 🙈) 

Do you think we could make it work with the useScroll solution you also had? 

https://user-images.githubusercontent.com/6661511/200808944-89abdafa-8b2f-4133-bc71-51a52fc43823.MP4

KarnellSchultz:
made a new PR for this one.

PR #829, opened 2022-11-09
GRW-1753 / Feature / Support i18n pluralization

alebedev:
## Describe your changes

Support i18next pluralization by
- fixing config
- post-processing translations after download
- applying special workaround for our Swedish locale

Allows us to use standard way to pluralize texts


apps/store/next-i18next.config.js
```diff
@@ -10,14 +10,21 @@ module.exports = {
   // - Storyblok folder slugs
   // - Dictionaries download script
   i18n: {
-    locales: ['default', 'se', 'en-se', 'dk', 'en-dk', 'no', 'en-no'],
+    locales: ['default', 'se', 'sv-se', 'en-se', 'dk', 'en-dk', 'no', 'en-no'],
```
gustaveen, line 13: So would this be needed for dk and no as well then?

gustaveen (approved):
Works good! 🙌 Up to you if you want to remove the sample code now or later :)

apps/store/next-i18next.config.js
alebedev: Fortunately not.  Even if language code is incorrect, at least it doesn't point to real language with different pluralization rules.  May bite us in other cases, but I don't see them yet

PR #830, opened 2022-11-09
Feature / Enable storyblok cache

alebedev:
## Describe your changes

Make storyblok requests faster by enabling built-in caching

Tested in preview app, subsequent requests are faster

- Speed
- Required for datasource-driven redirects, they are only practical when re-fetching the data is close to free

### DIFF
--- a/apps/store/src/services/storyblok/storyblok.ts
+++ b/apps/store/src/services/storyblok/storyblok.ts
@@ -173,6 +173,13 @@ export const initStoryblok = () => {
 
   storyblokInit({
     accessToken: process.env.NEXT_PUBLIC_STORYBLOK_ACCESS_TOKEN,
+    apiOptions: {
+      // ~290ms -> ~15ms for subsequent page renders
+      cache: {
+        type: 'memory',
+        clear: 'auto',
+      },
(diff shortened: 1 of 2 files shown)


apps/store/src/services/storyblok/storyblok.ts
```diff
@@ -173,6 +173,13 @@ export const initStoryblok = () => {
 
   storyblokInit({
     accessToken: process.env.NEXT_PUBLIC_STORYBLOK_ACCESS_TOKEN,
+    apiOptions: {
+      // ~290ms -> ~15ms for subsequent page renders
+      cache: {
+        type: 'memory',
+        clear: 'auto',
```
gustaveen, line 180: When will the automatic clear cache trigger?

gustaveen (approved):
Great! 🏎

apps/store/src/services/storyblok/storyblok.ts
alebedev: Not really documented, right now it only clears draft versions, according to storyblok-js-client sources.  But may start doing something useful in the future

PR #831, opened 2022-11-09
GRW-1702 / Feat  / Sticky TopMenu and Tablist

KarnellSchultz:
## Describe your changes

(Third time is the charm)

Implemented a css sticky type interaction between the TopMenu and TabList components. 

## Scroll Video
In the logs here you can see that the `onScroll` is not running when the TabList is not intersecting on the page.
That said, it'd be even better if we could not run the onScroll after the TabList replaces the TopNav. It's possible, but may take some time to sort that out.

https://user-images.githubusercontent.com/50870173/200874127-7e045f63-574e-4f4a-8b08-8f22bb6962a5.mov
The 🧽 emoji means that the `onScroll` is not running. 

The 🎽 emoji shows when`onScroll` is running

Designs require the TabList be stuck to the top of the page once a user scrolls past it. 

### DIFF
--- a/apps/store/src/utils/useIsIntersecting.tsx
+++ b/apps/store/src/utils/useIsIntersecting.tsx
@@ -1,6 +1,11 @@
 import { useEffect, useState } from 'react'
 
-export const useIsIntersecting = (node: Element | null, options = {}) => {
+type UseIsIntersectingType = (
+  node: Element | null,
+  options?: IntersectionObserverInit,
+) => readonly boolean[]
+
+export const useIsIntersecting: UseIsIntersectingType = (node, options = {}) => {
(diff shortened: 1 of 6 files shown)


apps/store/src/utils/useIsIntersecting.tsx
```diff
@@ -0,0 +1,21 @@
+import { useEffect, useState } from 'react'
+
```
KarnellSchultz, line 2: Made this one generic in the event that someone else wants to use it.

KarnellSchultz:
related to
https://github.com/HedvigInsurance/racoon/pull/826

gustaveen (approved):
Wohoo, here we go! 🎉 Looks great 🙌

PR #834, opened 2022-11-10
refact(store/Dialog): move it to component library

guilhermespopolin:
📓 This is part of a series of PRs; I believe it's easier to review then Top to Bottom;

## Describe your changes

* Promote `Dialog` component to `ui` library;

As I would need to implement a Dialog that displays the coverage information of a particular promise ([`CoverageDialog`](https://github.com/HedvigInsurance/racoon/blob/GRW-1716/feat/insurance-card/apps/onboarding/src/components/LandingPage/CoverageDialog.tsx)) I've decided to promote our `Dialog` component to the component library so we can re-use it.

Additionally, I figure that it would be better to move `zIndex` config to the global so we can also re-use it.

## Preview (End result)

[onboarding DK](https://onboarding-git-grw-1716-featinsurance-card-hedvig.vercel.app/dk-en/new-member)
[onboarding NO](https://onboarding-git-grw-1716-featinsurance-card-hedvig.vercel.app/no-en/new-member)


guilhermespopolin:
* main
  * **PR #834** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/834" target="_blank">Graphite</a>  👈
    * **PR #835** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/835" target="_blank">Graphite</a> 
      * **PR #836** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/836" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/834?utm_source=stack-comment).

alebedev (approved):
Great!

packages/ui/src/lib/theme/theme.ts
```diff
@@ -1,5 +1,6 @@
 import { colors, UIColor } from './colors'
 import { fonts, fontSizes } from './typography'
+import { zIndexes } from './zIndex'
```
gustaveen, line 3: Hm, my thought for the zIndexes was that it's local to the app since it can be hard to have the same stacking order in different apps. The components in ui kit can't and shouldn't know what context they will be used in IMO :)

gustaveen (commented):
Great to have it in the UI lib! 🙌  Just had some thoughts about the zIndex

packages/ui/src/lib/theme/theme.ts
alebedev: Maybe provide some defaults and let apps override indexes and insert new named entries?

guilhermespopolin: I see. Now that I'm thinking about it I agree with you. I've move it to `theme` so I could move promote `Dialog` to `ui` without "much problem" but it wasn't the correct move. Gonna address that.

packages/ui/src/components/Dialog/Dialog.tsx
```diff
@@ -33,7 +32,6 @@ export const Window = styled.div(({ theme }) => ({
 const StyledContentWrapper = styled.div({
   position: 'fixed',
   inset: 0,
-  zIndex: zIndexes.dialog,
```
guilhermespopolin, line 36: I've realised we're adding z-index here to solve issues that we shouldn't be getting in the first place since _radix_'s `Dialog` supports `Portals` and we're using them. [`Portals`](https://reactjs.org/docs/portals.html) we're invented  to implement modals so if by using them we still have content that stacks upon modals it means there's something off with our app. I fix it by creating a [stacking context](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/The_stacking_context) to our app so there will be no more "z-index war" between our main content and Portals.

I intent to prepare something about _stacking contexts_ so we can talk during a [Tech Forum](https://www.notion.so/hedviginsurance/Tech-Forum-b38c7f2c42be47bf96ceafb4ed4cb02f?p=45220794523e45698c11ae45c866b42e&pm=s)

packages/ui/src/lib/theme/theme.ts
guilhermespopolin: FYI I've reverted `zIndex` to `store` project :)

packages/ui/src/components/Dialog/Dialog.tsx
gustaveen: Ah yes, I agree, portals should take precedence over other stacking contexts in the app 👍

PR #836, opened 2022-11-10
GRW-1716 - refact: extract 'insurances by market' logic into a service

guilhermespopolin:
📓 This is part of a series of PRs; I believe it's easier to review then Top to Bottom;

## Describe your changes

* Refactor `embark` service so its API function in terms of `Market` instead of `MarketLabel`; In order to this to work, I had to include `Market` information in the `LocaleData` record (`src/lib/l10n/locales.ts`).
* Creates an `insurances` service responsible to handle what it was being handled by `LandingPage.constants`/`LandingPage.helpers` before: provides an API (`getInsurancesByLocaleLabel`) to get insurances by locale (in other terms, per market).
* Add a _graphql_ query for fetching coverage info of an insurance. That info is included in the insurance data provided by the `insurances` service (previous item).

## Justify why they are needed

It simplifies things and provides a data requirement for the [`CoverageDialog`](https://github.com/HedvigInsurance/racoon/blob/GRW-1716/feat/insurance-card/apps/onboarding/src/components/LandingPage/CoverageDialog.tsx)

## Preview (End result)

[onboarding DK](https://onboarding-git-grw-1716-featinsurance-card-hedvig.vercel.app/dk-en/new-member)
[onboarding NO](https://onboarding-git-grw-1716-featinsurance-card-hedvig.vercel.app/no-en/new-member)

PR #835, opened 2022-11-10
GRW-1716 - refact: unify main/additional coverage into a single insurance list

guilhermespopolin:
📓 This is part of a series of PRs; I believe it's easier to review then Top to Bottom;

## Describe your changes

* Simplify the information passed to `LandingPage`: instead of splitting the insurances of a market between `mainCoverage` and `additionalCoverage`, now they belong to a single insurance list.

## Justify why they are needed

This whole project is about selling _Travel_ and __Accident_ insurances as standalone products, which means everything is a standalone promise now. Therefore, there is no reason to keep such separation between `mainCoverage` and `additionalCoverage` insurances.

## Preview (End result)

[onboarding DK](https://onboarding-git-grw-1716-featinsurance-card-hedvig.vercel.app/dk-en/new-member)
[onboarding NO](https://onboarding-git-grw-1716-featinsurance-card-hedvig.vercel.app/no-en/new-member)

PR #837, opened 2022-11-10
GRW-1716 - refact(Checkbox): make it more accesible

guilhermespopolin:
📓 This is part of a series of PRs; I believe it's easier to review then Top to Bottom;

## Describe your changes

Updates `ui/Checkbox` component by making it more accessible an easy to work with. Its label is now wrapped by a `<label>` instead of a `<div>` which is the recommended way. It also accepts all `input` props, which makes it easy to use. I believe the more our custom components resemble the native Web API, the more easy to maintain they will become. 

So we can easily reuse it throughout our code base. Please observe tho, that I couldn't use this work in the other PR's down the stream because of an issue with `ui` build process 😢 
That means this work doesn't directly affect the other PRs. I've decided to keep it just to not throw away of what I think to be a valid addition. I've created a [ticket](https://hedvig.atlassian.net/browse/GRW-1760) for checking `ui` build process issue, which I intent to tackle in the next maintenance day.

## Preview

https://user-images.githubusercontent.com/19200662/201149340-21ae350b-8383-4469-afd8-07ef3ea16440.mov


guilhermespopolin:
* main
  * **PR #834** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/834" target="_blank">Graphite</a> 
    * **PR #835** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/835" target="_blank">Graphite</a> 
      * **PR #836** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/836" target="_blank">Graphite</a> 
        * **PR #837** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/837" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/837?utm_source=stack-comment).

packages/ui/src/components/Checkbox/Checkbox.tsx
```diff
@@ -59,70 +44,65 @@ const ControlLabel = styled.div<{ disabled?: boolean }>(
...
+  justifyContent: 'center',
+  alignItems: 'center',
+  width: '1.25rem',
+  height: '1.25rem',
+  border: `1px solid ${theme.colors.gray500}`,
+  borderRadius: circle ? '100%' : '2px',
+  background: theme.colors.white,
+  transition: 'all 150ms',
+  // checked/unchecked backgorund
+  [`${HiddenInput}:checked + &`]: {
```
alebedev, line 73: If you're showing HiddenInput yourself, maybe it's easier to have separate `styled` wrapped for it?  Should solve the build problem, I suspect

alebedev (approved):
Good one!

packages/ui/src/components/Checkbox/Checkbox.tsx
guilhermespopolin: I see. So this implementation takes advantage of a technique that I've saw recently in this CSS course that I've taking.

The general dynamic with native inputs is that they come with a bunch of desired functionality  but it's very hard to customise their look & feel. So wr need to compromise: 

* We create [custom checkboxes](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/checkbox_role), making them look as desired but then we'd need to handle a bunch accessibility stuff that comes for free with the native version of it.
* We accept the general look of checkboxes but then no additional work is required in order to make them work as expected. 

With that said, this technique tries to merge both worlds; It can be illustrated by the image below. Basically it puts an invisible `<input />` in front of the _presentation checkbox_ so we it will "look" as desired but it will also be accessible*.

Additionally I'm styling the _presentational bit_ based on native CSS selectors that takes checkbox state into account: the `:checked` pseudo selector. I've trying to keep everything styled related into CSS. I think it's a very scalable way to implement components. By using `:checked` here it will look accordantly despite the way we use that component: _controlled_ or _uncontrolled_. And it seems to be this movement in the industry where people will more and more try to implement forms with the native Web API, which means move away from _controlled_ components which finally means avoid to control style with JS.
One tradeoff of this approach is some weird  CSS selectors sometimes. But I believe [`has()`](https://developer.mozilla.org/en-US/docs/Web/CSS/:has) selector will really simplify stuff in that matter. So looking forward to have it widely available ([82% today](https://caniuse.com/?search=has)) 🤞🏻 

image

* the solution is not perfect tho, as it never is. One thing that we always need to do with it is to implement the `outline` since the default style for it, that it's provided the browser, will not be shown because the input is invisible. However it seems a small price to pay since one probably will override the default style for it anyway. The important thing is that it has some :)

image

KarnellSchultz: Tack for the info dump 🥔

gustaveen: Nice! This is the way to go IMO, using pseudo selectors to style different input states  🎉

gustaveen (approved):
Good stuff @guilhermespopolin 🙌

packages/ui/src/components/Checkbox/Checkbox.tsx
```diff
@@ -59,70 +44,65 @@ const ControlLabel = styled.div<{ disabled?: boolean }>(
...
+    visibility: 'revert',
+  },
+  [`${HiddenInput}:not(:enabled) + & > ${DisabledTick}`]: {
+    display: 'revert',
+  },
+}))
+
+const CheckboxWrapper = styled.div({
+  position: 'relative',
+  isolation: 'isolate',
```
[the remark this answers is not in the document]
guilhermespopolin: So `CheckboxWrapper` is the wrapper that contains the _hidden input_ and the _presentational bit_. Since I'm making sure the _hidden input_ is on top the the _presentational bit_ by setting a `z-index` for it (`z-index: 1`), I'm isolating them by creating a new stacking context on their parent (`CheckboxWrapper`) to avoid any weird `z-index` bugs.

It seems like a good pratice:

> Whenever a child within a component applies a z-index value, I add isolation: isolate to the component's parent element. This guarantees that we won't see weird "slip-in-between" bugs, like the one we saw with the sticky header. But it doesn't contribute at all to z-index inflation, or force me to pick a value

At first place I've tried to avoid relying on `z-index` to achieve the _hidden input on top of the presentational bit_ thing. Since `HiddenInput` and `StyledCheckboxElement` are both positioned elements their stacking order would respect the DOM order, which means I'd be able to achieve the same thing without any `z-index` if I just switch the order they appear in the DOM: `StyledCheckboxElement`,`HiddenInput` instead of `HiddenInput`,`StyledCheckboxElement` as we have today. The issue is that I need them in the `HiddenInput`,`StyledCheckboxElement` order because of the selectors I'm using: `element + element`

PR #843, opened 2022-11-11
FIx / Improve page load time

alebedev:
## Describe your changes

- Lazy load GTM when browser is idly
- Delay loading intercom script

See intercom problem in Lighthouse on screenshot below

Improves page load by spending less time on non-essential parts


gustaveen (approved):
Speed is our friend! 👌 Only concern is that we load some GTM scripts too late with this strategy. But we can test and see if Sonny see any problems with it

alebedev:
@gustaveen it should be fine - GTM handles data added before its initialization already.  So, it's safe to initialize late, assuming user does not close the page within first few seconds

PR #857, opened 2022-11-14
GRW-1766 / Feature / Use @next/font in store

alebedev:
## Describe your changes

- Add @next/font for loading fonts
- Add font files to repository
- Keep onboarding and storybook with old fonts for now, we should also switch them if we agree to proceed with @next/font
- Less magic with applying global font styles

- Better page load speed, no need to use separate domain for fonts
- More agressive caching policy for fonts -> faster load on repeated visits


apps/store/src/pages/_app.tsx
```diff
@@ -15,6 +16,38 @@ import { SHOP_SESSION_PROP_NAME } from '@/services/shopSession/ShopSession.const
...
+  src: '../../../../packages/ui/fonts/HedvigLetters-Small.woff2',
+  weight: '400',
+  fallback: ['sans-serif'],
+  variable: '--hedvig-font-small',
+  display: 'swap',
+})
+const standardFont = localFont({
+  src: '../../../../packages/ui/fonts/HedvigLetters-Standard.woff2',
+  weight: '400',
+  fallback: ['serif'],
```
gustaveen, line 32:
```suggestion
  fallback: ['sans-serif'],
```

PR #859, opened 2022-11-15
GRW-1759 / fix maxLength of car register

KarnellSchultz:
## Describe your changes

fixed mac length for the car register input field. 

**Before**
Screenshot 2022-11-14 at 15 00 55

**After**
Screenshot 2022-11-14 at 15 00 42

previously you could add as many chars as you'd like. 

[GRW-1759]: https://hedvig.atlassian.net/browse/GRW-1759?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/store/src/components/PriceForm/CarRegistrationField.tsx
+++ b/apps/store/src/components/PriceForm/CarRegistrationField.tsx
@@ -30,7 +30,7 @@ export const CarRegistrationNumberField = ({ field }: RegistrationFieldProps) =>
       label={field.label ? translateLabel(field.label) : undefined}
       pattern={CAR_REGISTRATION_NUMBER_REGEX}
       placeholder="ABC 123"
-      minLength={7}
+      maxLength={7}
       required={field.required}
       value={value}
       defaultValue={field.defaultValue}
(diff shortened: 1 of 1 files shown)


apps/store/src/components/PriceForm/CarRegistrationField.tsx
```diff
@@ -30,7 +30,7 @@ export const CarRegistrationNumberField = ({ field }: RegistrationFieldProps) =>
       label={field.label ? translateLabel(field.label) : undefined}
       pattern={CAR_REGISTRATION_NUMBER_REGEX}
       placeholder="ABC 123"
-      minLength={7}
+      maxLength={7}
```
gustaveen, line 33: Is `minLength` covered by the pattern instead?

gustaveen (approved):
🙌

apps/store/src/components/PriceForm/CarRegistrationField.tsx
KarnellSchultz: It'll throw an error if you try to submit something that does not match the pattern or longer than the pattern. But the `maxLength` prevents the user from entering more than the expected number of chars.

abderrahmane-tj:
@KarnellSchultz 👋🏼 is this repo meant to be open to the public? (and a couple more 👇🏼)
image

PR #860, opened 2022-11-15
Fix / householSize field everywhere

alebedev:
## Describe your changes

- Use good-looking householdSize field in home and student apartment insurance
- Ignoring accident for now, field is likely not needed there

### DIFF
--- a/apps/store/src/services/PriceForm/data/SE_HOUSE.ts
+++ b/apps/store/src/services/PriceForm/data/SE_HOUSE.ts
@@ -129,13 +129,12 @@ export const SE_HOUSE: Template = {
       items: [
         {
           field: {
-            type: 'number',
-            min: 0,
-            max: 5,
+            type: 'householdSize',
             name: 'numberCoInsured',
-            label: { key: 'Number of co-insured' },
+            label: { key: 'Household size' },
             required: true,
             defaultValue: 0,
+            max: 5,
           },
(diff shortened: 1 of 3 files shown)


apps/store/src/services/PriceForm/data/SE_HOUSE.ts
```diff
@@ -129,13 +129,12 @@ export const SE_HOUSE: Template = {
...
           field: {
-            type: 'number',
-            min: 0,
-            max: 5,
+            type: 'householdSize',
             name: 'numberCoInsured',
-            label: { key: 'Number of co-insured' },
+            label: { key: 'Household size' },
             required: true,
             defaultValue: 0,
```
amandaflod, line 136: Should we maybe change the default value to 1 since it is changed to household size?

amandaflod (approved):
Looks good! Just one little comment/question 🙂

apps/store/src/services/PriceForm/data/SE_HOUSE.ts
alebedev: It's still posted as numberCoInsured, I don't want to handle value transformation in this PR.  There's alternative approach to that in @robinandeer, let's discuss this with him first

amandaflod: Ok, sounds good!

PR #868, opened 2022-11-16
GRW-1779 / Feature / Show signing progress and errors

alebedev:
## Describe your changes

- Display signing progress in UI
- Display signing errors

Ability to understand what's going on without console

### DIFF
--- a/apps/store/src/components/CheckoutPage/CheckoutPage.tsx
+++ b/apps/store/src/components/CheckoutPage/CheckoutPage.tsx
@@ -2,6 +2,7 @@ import styled from '@emotion/styled'
 import Link from 'next/link'
 import { ArrowForwardIcon, Button, Heading, HedvigLogo, InputField, Space } from 'ui'
 import { CartInventory } from '@/components/CartInventory/CartInventory'
+import { PersonalNumberField } from '@/components/PersonalNumberField/PersonalNumberField'
 import { CheckoutSigningStatus } from '@/services/apollo/generated'
 import { PageLink } from '@/utils/PageLink'
 import { FormElement } from './CheckoutPage.constants'
@@ -67,10 +68,8 @@ const CheckoutPage = (props: CheckoutPageProps) => {
               <Heading as="h2" variant="standard.24">
                 2. Your contact info
               </Heading>
-              <InputField
-                label="Personal Number"
+              <PersonalNumberField
                 name={FormElement.PersonalNumber}
-                type="text"
                 required
                 defaultValue={prefilledData.personalNumber ?? undefined}
                 errorMessage={userErrors[FormElement.PersonalNumber]}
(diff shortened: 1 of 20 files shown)


apps/store/src/components/CheckoutPage/CheckoutPage.tsx
```diff
@@ -128,6 +127,25 @@ const CheckoutPage = (props: CheckoutPageProps) => {
   )
 }
 
+const SubmitButton = ({
+  loading,
+  signingStatus,
+}: Pick<CheckoutPageProps, 'loading' | 'signingStatus'>) => {
```
robinandeer, line 133: Maybe a bit overengineered ;)

We could just type it from scratch, no?

apps/store/src/components/CheckoutPage/CheckoutPage.tsx
```diff
@@ -128,6 +127,25 @@ const CheckoutPage = (props: CheckoutPageProps) => {
...
+  let label
+  if (signingStatus === CheckoutSigningStatus.Pending) {
+    label = 'Please open your BankID app and sign there'
```
robinandeer, line 136: Did we say anything about using translation keys?

apps/store/src/services/Checkout/useHandleSignCheckout.ts
```diff
@@ -17,18 +18,19 @@ export const useHandleSignCheckout = (params: Params) => {
   // eslint-disable-next-line
   const { checkoutId, checkoutSigningId: initialCheckoutSigningId, onSuccess } = params
   const [checkoutSigningId, setCheckoutSigningId] = useState(initialCheckoutSigningId)
+  const [signingStatus, setSigningStatus] = useState<CheckoutSigningStatus | undefined>()
```
robinandeer, line 21: Why do we need this separate state? We should just pass value from the query itself

```
const { data } = useCheckoutSigningQuery(...)
```

apps/store/src/services/Checkout/useHandleSignCheckout.ts
```diff
@@ -44,24 +46,29 @@ export const useHandleSignCheckout = (params: Params) => {
...
     {
       loading: result.loading || Boolean(checkoutSigningId),
-      userErrors,
+      signingStatus,
+      userErrors: getErrors(result),
     },
   ] as const
 }
+
+const getErrors = (result: CheckoutStartSignMutationResult) => {
```
robinandeer, line 66: Let's discuss a scalable way to handle this with backend!

robinandeer (commented):
Looks good! Just a few suggested updates -- mainly to pass query data instead of using separate react state :)

apps/store/src/components/CheckoutPage/CheckoutPage.tsx
alebedev: Not yet, but we should discuss it soon

apps/store/src/components/CheckoutPage/CheckoutPage.tsx
alebedev: Since we drill down these props and they are declared in different module, I'd argue that this way is easier to maintain

apps/store/src/services/Checkout/useHandleSignCheckout.ts
alebedev: Do you have any starting ideas on how to improve?

robinandeer: Not really, but I can start to collect some common use cases and then see how we can approach it

robinandeer (approved):
Looks good, let's try this out :D

PR #876, opened 2022-11-16
Set domain option for auth session cookie

robinandeer:
## Describe your changes

Set the domain option for the auth session cookie.

We need to to be able to share the cookie between different subdomains.

It's not possible to test this locally which is unfortunate.

Perhaps we can consider a similar solution to how the "payment link" in HOPE works? There we inlude the "exchange token" in the URL which web onboarding uses to get an access token...

### DIFF
--- a/apps/store/src/services/Auth/Auth.ts
+++ b/apps/store/src/services/Auth/Auth.ts
@@ -1,5 +1,6 @@
 import { getCookie, setCookie } from 'cookies-next'
 import { OptionsType } from 'cookies-next/lib/types'
+import { ORIGIN_URL } from '@/utils/PageLink'
 
 const COOKIE_KEY = '_hvsession'
 const ACCESS_TOKEN_SESSION_FIELD = 'token'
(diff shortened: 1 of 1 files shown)


robinandeer:
* main
  * **PR #876** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/876" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/876?utm_source=stack-comment).

apps/store/src/services/Auth/Auth.ts
```diff
@@ -9,13 +10,22 @@ export const save = (accessToken: string) => {
...
       secure: true,
     }),
   })
 }
 
+const getRootDomain = () => {
+  const url = new URL(ORIGIN_URL)
+  const hostname = url.hostname
+  const parts = hostname.split('.')
+  const rootDomtainParts = parts.slice(-2, parts.length + 1)
```
amandaflod, line 25: Just a small typo on rootDomainParts here and the line below :)

amandaflod (approved):
🙌 Left a small comment on a typo if you want to fix that!

apps/store/src/services/Auth/Auth.ts
alebedev, line 25: `.slice(-2)` - no need for end index if you want all remaining elements

alebedev (approved):
Nice!

Didn't know before that leading dot in domain name is not needed for modern browsers

alebedev:
Going to fix and merge it myself to move faster

PR #879, opened 2022-11-17
GRW-1777 / Fix / Payment connection origin

alebedev:
## Describe your changes

Fix origin detection based on env variables so that we actually use env.NEXT_PUBLIC_ORIGIN_URL when it's available

Fixes payment connection flow on staging


amandaflod (approved):
🙌

PR #914, opened 2022-11-22
Refactor PriceCalculator -> PurchaseForm

robinandeer:
## Describe your changes

Rename `PriceCalculator` to `PurchaseForm`.

Add `OfferPresenter` to handle displaying price summar, selecting tier, and adding to cart. This is only shown when we have some offer(s) to present to the user.

The `PurchaseForm` handles more things than calculating a price so we want to reflect that in the name.

We will follow-up with more actions:

- Renaming `PriceForm` to `PriceCalculator` (to match Figma)
- Rename `PriceCalculatorBlock` in Storyblok or remove it


alebedev (approved):
Looks good

PR #918, opened 2022-11-23
Remove start dates from Checkout

robinandeer:
## Describe your changes

Remove the start date inputs from the checkout.

Move the hook to update start date to the `PurchaseForm` scope.

Design changes.

[GRW-1817]: https://hedvig.atlassian.net/browse/GRW-1817?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/store/src/components/ProductPage/PurchaseForm/useUpdateStartDate.ts
```diff
@@ -0,0 +1,70 @@
...
+  return [
+    handleUpdateStartDate,
+    {
+      loading: result.loading,
+      userError,
+    },
+  ] as const
+}
+
+const convertToDate = (value: unknown) => {
```
alebedev, line 58: Is it time to create utils/date?

alebedev (approved):
Looks good, one small comment inside

apps/store/src/components/ProductPage/PurchaseForm/useUpdateStartDate.ts
robinandeer: Perhaps, do we have anywhere else we need this?

alebedev: Already did it locally, so let's consider this resolved

PR #919, opened 2022-11-23
GRW-1817 / Feature / Move startDate field into purchase form

alebedev:


apps/store/src/components/ProductPage/PurchaseForm/useUpdateStartDate.ts
```diff
@@ -1,20 +1,28 @@
...
+  // TODO: Improve this name, it's not event handler
+  const handleUpdateStartDate = async ({
+    offerId,
+    dateValue,
+    onSuccess,
+  }: {
+    offerId: string
+    dateValue: string | null
+    onSuccess: () => void
+  }) => {
```
robinandeer, line 25: I think we should avoid defining complex types inline -- okey for now but we should change later

robinandeer (approved):
Looks good! Just had a small comment

PR #923, opened 2022-11-23
Create fields for Overview and Coverage on PDP and inline PurchaseForm

gustaveen:
## Describe your changes
First step in towards new layout
Create fields for Overview and Coverage on PDP and inline PurchaseForm

## Justify why they are needed
Too support updated layout for PDP

https://user-images.githubusercontent.com/6661511/203581427-c6524252-6fa7-414a-9b6a-6ba2d10ba3c0.mov

## Checklist before requesting a review

- [ ] I have performed a self-review of my code

PR #924, opened 2022-11-23
Render different layout for mobile and desktop.

gustaveen:
## Describe your changes
Render different layout for mobile and desktop.
Since there's no straight forward way to "destroy" the tabs we will render different markup on different viewports.
Either we do it by js breakpoints or css breakpoints. Currently we get hydration warnings when doing it with js breakpoints. 
What do we prefer? I kind of favour hiding it with css if it doesn't come with drawbacks rendering more to the DOM

Next steps:
- Possibly break this out in to another component
- Get tab labels from CMS
- Add scroll to links for desktop, but as I understood design might be revised

### Desktop

https://user-images.githubusercontent.com/6661511/203760710-2c4756fb-00df-4e67-bfc7-3f3ad77d0ada.mov

### Mobile

https://user-images.githubusercontent.com/6661511/203760780-39f0eec1-ba13-4dce-a661-05f214782740.mov

Update layout for desktop

### TODO

- [x] Render section in tabs for mobile
- [x] Render sections on page with scroll to links for desktop

### DIFF
--- a/apps/store/src/blocks/ProductPageBlock.tsx
+++ b/apps/store/src/blocks/ProductPageBlock.tsx
@@ -1,5 +1,6 @@
 import styled from '@emotion/styled'
 import { storyblokEditable, StoryblokComponent, SbBlokData } from '@storyblok/react'
+import { useBreakpoint } from 'ui'
 import { PurchaseForm } from '@/components/ProductPage/PurchaseForm/PurchaseForm'
 import * as Tabs from '@/components/ProductPage/Tabs'
 import { SbBaseBlockProps } from '@/services/storyblok/storyblok'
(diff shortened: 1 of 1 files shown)


gustaveen:
* main
  * **PR #923** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/923" target="_blank">Graphite</a> 
    * **PR #924** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/924" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/924?utm_source=stack-comment).

apps/store/src/blocks/ProductPageBlock.tsx
```diff
@@ -10,15 +12,54 @@ type ProductPageBlockProps = SbBaseBlockProps<{
...
+          <PurchaseForm />
+          <Tabs.Tabs defaultValue="overview">
+            <Tabs.TabsList>
+              {/* TODO: Get tab labels from Storyblok */}
+              <Tabs.TabsTrigger value="overview">Overview</Tabs.TabsTrigger>
+              <Tabs.TabsTrigger value="coverage">Coverage</Tabs.TabsTrigger>
+            </Tabs.TabsList>
+
+            <Tabs.TabsContent value="overview">
+              {blok.overview?.map((nestedBlock) => (
```
gustaveen, line 29: Should we memoize this?

robinandeer (approved):
Great work!

I think resorting to JS for dealing with breakpoints should be a last resort. It's not responsive design. But of course, it needs to start with design so they need to have the same mindset.

gustaveen:
> Great work!
> 
> I think resorting to JS for dealing with breakpoints should be a last resort. It's not responsive design. But of course, it needs to start with design so they need to have the same mindset.

Thanks! I hear you, let's do it with CSS. 🙌 It's mostly the Tabs that makes it a bit problematic. Other than that it would be somewhat straight forward since it's all the same content, so it's responsive in the sense, it's just the layout it's presented in. that changes.

apps/store/src/blocks/ProductPageBlock.tsx
alebedev: Let's see it being visible as a problem in profiler first

alebedev (approved):
Looks like a good start.  Perhaps we can encapsulate js part of responsiveness in something like `ResponsiveTabs` component or block - then even if we'd do it js, it won't leak outside

gustaveen:
> Looks like a good start. Perhaps we can encapsulate js part of responsiveness in something like `ResponsiveTabs` component or block - then even if we'd do it js, it won't leak outside

Cool, let's merge this for now and sort out how we want to best treat these type of media queries.

gustaveen:
**Graphite Merge Job**

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/924?utm_source=stack-comment).


Job Reference: JeC9qaAfvgPRoKcv6BUS

PR #925, opened 2022-11-23
GRW-1798 / Feature / Use authorizationCode instead of shared cookie to pass access info to web-onboarding

alebedev:
## Describe your changes

- Use accessToken to generate new one-time authorizationCode and append it to payment connection URL
- access token cokkie no longer cross-domain

- Makes us independent of DNS setup between store and web-onboarding, allows to test payment connection on any branch
- Reduces risk from 3rd part web apps we may install at hedvig-com subdomains


robinandeer (approved):
[excited](https://app.graphite.dev/macros/HedvigInsurance)

PR #950, opened 2022-11-28
Fix / Connect ApolloDevTools

alebedev:
## Describe your changes

Fix integration with ApolloDevTools Chrome extension

It used to work automatically by relying on __DEV__ global variable from React, now it doesn't.  Setting value explicitly is better than relying on React internals

Fix developer experience


robinandeer (approved):
[likeaboss](https://app.graphite.dev/macros/HedvigInsurance)

PR #954, opened 2022-11-29
GRW-1847 / Feature / Simplified userError API

alebedev:
## Describe your changes

- Switch from userErrors to userError for all mutations other that `priceIntentDataUpdate`
- Extract getMutationError util

See Slack discussion here: https://hedviginsurance.slack.com/archives/G01MARCFUUW/p1669646168923799


robinandeer (approved):
Good stuff! This is much easier to reason about and we can add support for showing errors next to relevant inputs when we have time to implement that :)

PR #984, opened 2022-12-02
Fix layout in `OfferPresenter`

robinandeer:
## Describe your changes

Fix layout issue in `OfferPresenter` when `CancellationForm` is visible.

Avoid controlling checkbox since it leads to strange UX. We should update optimistically and handle errors instead.

Subscribe to `PriceIntent` updates client side so UI is kept up to date.

I refactored a bit to subscribe to `PriceIntent` updates using Apollo Client. Not sure if this is the direction we want to move in generally. Happy to hear feedback on this!

### DIFF
--- a/apps/store/src/pages/products/[product].tsx
+++ b/apps/store/src/pages/products/[product].tsx
@@ -8,6 +8,7 @@ import { ProductPage } from '@/components/ProductPage/ProductPage'
 import { getProductData } from '@/components/ProductPage/ProductPage.helpers'
 import { ProductPageProps } from '@/components/ProductPage/ProductPage.types'
 import { addApolloState, initializeApollo } from '@/services/apollo/client'
+import { usePriceIntentQuery } from '@/services/apollo/generated'
 import logger from '@/services/logger/server'
 import { fetchPriceTemplate } from '@/services/PriceCalculator/PriceCalculator.helpers'
 import { priceIntentServiceInitServerSide } from '@/services/priceIntent/PriceIntent.helpers'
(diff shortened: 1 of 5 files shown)


robinandeer:
* main
  * **PR #984** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/984" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/984?utm_source=stack-comment).

apps/store/src/pages/products/[product].tsx
```diff
@@ -30,16 +32,20 @@ type PageQueryParams = {
...
 }
 
-const NextProductPage: NextPageWithLayout<NextPageProps> = (props) => {
-  const story = useStoryblokState(props.story)
+const NextProductPage: NextPageWithLayout<NextPageProps> = ({ priceIntent, ...pageProps }) => {
+  const story = useStoryblokState(pageProps.story)
+
+  const { locale } = useCurrentLocale()
+  const { data } = usePriceIntentQuery({ variables: { priceIntentId: priceIntent.id, locale } })
+  const livePriceIntent = data?.priceIntent || priceIntent
```
Comment (alebedev) [inline apps/store/src/pages/products/[product].tsx:40]: Nice!  Does it mean we can remove router refresh when purchase form is complete and we're switching to offer display?

alebedev (approved):
Looks good.  One question on future improvements inside

apps/store/src/pages/products/[product].tsx
Comment (robinandeer) [reply apps/store/src/pages/products/[product].tsx:40]: Yes, I believe we can just subscribe to changes instead using Apollo Client and it will handle that for us :)

PR #991, opened 2022-12-02
Fix/ cart's "Go to Store" button link

KarnellSchultz:
## Describe your changes

before the user needed to click on the 'go to store' text to follow the link. 

Now the button will send the user to the store. 

Improved user exp


apps/store/src/components/CartPage/CartPage.tsx
```diff
@@ -130,9 +131,9 @@ const EmptyState = () => {
         </StyledHeading>
         <CenteredParagraph>{t('CART_EMPTY_SUMMARY')}</CenteredParagraph>
         <Footer>
-          <Button fullWidth>
-            <Link href={PageLink.store()}>Go to Store</Link>
-          </Button>
+          <LinkButton fullWidth href={PageLink.store({ locale: routingLocale })}>
```
robinandeer, line 134: Doesn't nextjs handle the locale automatically?

robinandeer (approved):
Nice nice!

apps/store/src/components/CartPage/CartPage.tsx
KarnellSchultz: I thought that'd be the case but for some reason without adding the `routingLocale` it linked to the store without a local i.e. a broken page.

I can play around with it a bit more to see if I can get it to work without the routingLocale though.

PR #1008, opened 2022-12-05
Fix / Update priceIntent.startDate in API

alebedev:
## Describe your changes

- Make priceIntent.startDate required
- Rely on default value from backend

API updatem move logic to backend to be more flexible with start date depending on insurance info we may have

### DIFF
--- a/apps/store/src/components/ProductPage/PurchaseForm/OfferPresenter.tsx
+++ b/apps/store/src/components/ProductPage/PurchaseForm/OfferPresenter.tsx
@@ -73,8 +73,6 @@ export const OfferPresenter = ({
 
   const loading = loadingAddToCart || updateCancellationInfo.loading || updateStartDateInfo.loading
 
-  const dateToday = new Date()
-
   return (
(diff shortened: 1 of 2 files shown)


KarnellSchultz (approved):

apps/store/src/components/ProductPage/PurchaseForm/OfferPresenter.tsx
```diff
@@ -92,7 +90,7 @@ export const OfferPresenter = ({
           <Space y={0.25}>
             <CancellationForm
               option={cancellationOption}
-              startDate={convertToDate(priceIntent.startDate) ?? dateToday}
+              startDate={convertToDate(priceIntent.startDate)!}
```
robinandeer, line 93: Why do we need to force unwrap here? Isn't the start date required on the price intent?

robinandeer (commented):
Nice -- just a question

robinandeer (approved):

PR #1024, opened 2022-12-07
GRW-1641 / Fix / Faster codegen

alebedev:
## Describe your changes

- Remove prettier and eslint steps from
- Simplify turbo config by lifting up common settings

Codegen time on my machine (median of 3 runs)
before: 9.1s
after: 3.8s 

Better developer experience

### DIFF
--- a/apps/onboarding/package.json
+++ b/apps/onboarding/package.json
@@ -13,7 +13,7 @@
     "lint:fix": "next lint --fix --ignore-path .gitignore",
     "clean": "rm -rf .turbo && rm -rf node_modules && rm -rf .next",
     "download-translations": "./scripts/download-translations.sh",
-    "codegen": "graphql-codegen && prettier --write ./src/services/**/{generated,types}.ts && yarn lint:fix",
+    "codegen": "graphql-codegen",
     "cypress": "cypress open",
     "cypress:headless": "cypress run",
     "typecheck": "tsc --noEmit"
--- a/turbo.json
+++ b/turbo.json
@@ -1,33 +1,35 @@
 {
   "$schema": "https://turborepo.org/schema.json",
   "pipeline": {
-    "build": {
+    "codegen": {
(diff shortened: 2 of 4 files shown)


gustaveen (approved):
Nice if this works! 🙌 
Do we know the reason why we ran prettier and lint in codegen, did we need to format it?

turbo.json
```diff
@@ -1,63 +1,45 @@
...
-      "outputs": [".next/**"],
-      "dependsOn": ["^build", "codegen"]
-    },
-    "onboarding#test": {
-      "outputs": ["coverage/**"],
-      "dependsOn": ["^build"]
+    "codegen": {
+      "inputs": ["src/graphql/*.graphql", "codegen.yml"],
+      "outputs": ["src/services/**/generated.ts", "src/services/apollo/types.ts"],
+      "cache": false
```
robinandeer, line 7: Hmm just wondering what this means? Does Turbo do anything with inputs/outputs if we turn off the cache?

alebedev:
I assume the reason we ran prettier and lint on generated files was that it broke.  I find it's better to exclude generated files from check - we're not linting anything in "dist", for example

apps/onboarding/package.json
```diff
@@ -13,7 +13,7 @@
     "lint:fix": "next lint --fix --ignore-path .gitignore",
     "clean": "rm -rf .turbo && rm -rf node_modules && rm -rf .next",
     "download-translations": "./scripts/download-translations.sh",
-    "codegen": "graphql-codegen && prettier --write ./src/services/**/{generated,types}.ts && yarn lint:fix",
+    "codegen": "graphql-codegen",
```
robinandeer, line 16: We don't need to load "dotenv" here?

robinandeer (commented):
Great stuff! Just asking stuff to better understand

apps/onboarding/package.json
alebedev: It's hardcoded, don't want to change it now

PR #1028, opened 2022-12-07
Fix / Cache codegen and force it in CI

alebedev:
## Describe your changes

Cache codegen step.  Force run it in CI

Better developer experience, we don't change schema all that ofter.   Developers will have to force codegen with `yarn dev --force` to flush stale cache


robinandeer (approved):
Nice nice - I think this is a great tradeoff :D

There's also some editor plugins that can help to validate the schema

PR #1031, opened 2022-12-07
GRW-1778 - refact(Checkbox): adding support for receiving a ref

guilhermespopolin:
## Describe your changes

* Update default onboarding flows for DK and NO: `onboarding-DK-v3` and `onboarding-NO-v2`
* Remove old implementation for `LandingPage`
* Update `ui/Checkbox` so it accept a `ref`
* Remove old images for _House_ and _Accident_ promises
* Remove `TRAVEL_ACCIDENT_STANDALONE` feature flag

Cleanup post release of _Travel&Accident Standalone_

## Jira issue(s): [GRW-1778](https://hedvig.atlassian.net/browse/GRW-1778)


guilhermespopolin:
* main
  * **PR #1031** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1031" target="_blank">Graphite</a>  👈
    * **PR #1032** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1032" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1031?utm_source=stack-comment).

packages/ui/src/components/Checkbox/Checkbox.tsx
```diff
@@ -94,52 +94,50 @@ const CheckboxWrapper = styled.div({
...
+        </ControlLabel>
+      )
+    }
+
+    return (
+      <ControlContainer>
+        {prependLabel && LabelComponent}
+        <CheckboxWrapper>
+          <HiddenInput
+            ref={mergeRefs([inputRef, ref])}
```
robinandeer, line 126: We should consider if we should even allow this 😅 but fine for now

robinandeer (approved):
Very nice from what I can tell - great clean-up!

packages/ui/src/components/Checkbox/Checkbox.tsx
guilhermespopolin: Let's discuss it them 😀
In case you're not familiar with the reasoning behind it:

> When developing low level UI components, it is common to have to use a local ref but also support an external one using `React.forwardRef`

That's what happened with me here: `Checkbox` [implementation](https://github.com/HedvigInsurance/racoon/blob/GRW-1778/refact/landing-page/packages/ui/src/components/Checkbox/Checkbox.tsx#L99) make usage of a ref in order to support custom validation. However I also need to be able to pass in a `ref` to that checkbox component from [`InsuranceCard`](https://github.com/HedvigInsurance/racoon/blob/GRW-1778/refact/landing-page/apps/onboarding/src/components/LandingPage/InsuranceCard.tsx#L24) in order to toggle the checkbox by clicking on the card.

I guess the whole issue think came from the fact that one can interact with `InsuranceCard` in several ways:

* It should be possible to click on the checkbox
* It should be possible to click on the "open coverage dialog" button
* **It should be possible to click on the card meaning a click on the checkbox**

If we could ignore the interaction with the whole card, then things would be easier.

Another common solution for such card component would be to control the checkbox inside of it and attach a `handleClick` to the whole card that then, would just call checkboxe's `handleChange`:

```javascript
const Card = () => {
  const [checked, setChecked] = useState(false)

  const handleChange = () => setChecked(!checked)

  return (
    // I guess even with a solution like that, one would need to verify inside this `handleClick` if the `target`
    // is not the Checkbox itself otherwise a double toggle would happen.
    // That would also require a `ref` to the Checkbox 🤷🏻 
    <div onClick={() => handleChange()}>
      <Checkbox checked={checked} onChange={handleChange} />
    </div>
  )
}
```

How would you handle that case?

PR #1033, opened 2022-12-07
Fix / prevent duplicate session creation

alebedev:
## Describe your changes

Prevent duplicate creation of shopSession in strict mode

Reference solution for not running mutations twice.  We should use it sparingly

### DIFF
--- a/apps/store/src/services/shopSession/ShopSessionContext.tsx
+++ b/apps/store/src/services/shopSession/ShopSessionContext.tsx
@@ -1,5 +1,5 @@
 import { QueryResult, useApolloClient } from '@apollo/client'
-import { createContext, PropsWithChildren, useContext, useEffect, useMemo } from 'react'
+import { createContext, PropsWithChildren, useContext, useEffect, useMemo, useRef } from 'react'
 import {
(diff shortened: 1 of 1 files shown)


apps/store/src/services/shopSession/ShopSessionContext.tsx
```diff
@@ -47,13 +48,14 @@ export const ShopSessionProvider = ({ children, shopSessionId: initialShopSessio
     },
   })
 
-  // Has to be wrapped to prevent duplicate execution (Apollo quirk leads do duplicate execution when called directly from render)
   useEffect(() => {
+    // calledRef ensures we don't call the effect twice in strict mode
     // TODO: isBrowser() and ensure code splitting does not break
-    if (typeof window !== 'undefined' && !shopSessionId && !mutationResult.called) {
+    if (typeof window !== 'undefined' && !shopSessionId && !calledRef.current) {
```
robinandeer, line 54: Do we really need to check if we run on the server here?

robinandeer (approved):
I see, escape hatch but a decent one -- sometimes it's gotta be done

PR #1040, opened 2022-12-08
Render all CMS-pages from slug-endpoint

robinandeer:
## Describe your changes

Render all pages (including product pages) from the `[[...slug]]` route.

Conditionally fetch and render product pages.

Remove product page route.

Now we can mix and match content and product pages as we wish from the CMS.

To create a product page all you need to do is create it with the "product" content type.

[GRW-1915]: https://hedvig.atlassian.net/browse/GRW-1915?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/store/src/pages/[[...slug]].tsx
```diff
@@ -14,17 +19,21 @@ import {
...
-import { isRoutingLocale } from '@/utils/l10n/localeUtils'
+import { isProductStory } from '@/services/storyblok/Storyblok.helpers'
+import { isRoutingLocale, toApiLocale } from '@/utils/l10n/localeUtils'
 
-type RoutingPath = {
-  params: {
-    slug: string[]
-  }
-  locale?: string
+type NextStoryblokPageProps = StoryblokPageProps & { type: 'story' }
```
Comment (gustaveen) [inline apps/store/src/pages/[[...slug]].tsx:25]: nit: I wonder if we can name it something else, since ProductPage is also a Storyblok page :D but can't think of anything better haha

gustaveen (approved):
Awesome, works like charm! 🙌

PR #1042, opened 2022-12-08
Fix design so dialogs are scrollable

robinandeer:
## Describe your changes

Fix styles for Dialogs so the content is scrollable on smaller screens.

### Price calculator dialog

Refactor Dialog: remove one nested div that's not needed.


robinandeer:
* main
  * **PR #1042** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1042" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1042?utm_source=stack-comment).

apps/store/src/components/PriceCalculator/CurrentInsuranceField/CurrentInsuranceField.tsx
```diff
@@ -109,13 +109,22 @@ const useUpdateExternalInsurer = (priceIntentId: string) => {
...
+  paddingLeft: theme.space[2],
+  paddingRight: theme.space[2],
+}))
 
 const StyledDialogWindow = styled(Dialog.Window)(({ theme }) => ({
-  marginInline: theme.space[2],
-  marginTop: theme.space[2],
-  borderRadius: theme.radius.xs,
+  width: '100%',
+  maxWidth: 700,
```
guilhermespopolin, line 124: Should we convert these to relative units?

```suggestion
  maxWidth: '43.75rem',
  ...
  maxHeight: '49.38rem',
```

Btw, the following can be a nit trick:

```
maxWidth: 'calc(700rem / 16)'; /* ~700px */
```

It's easier to think about px but in general we might want relative sizes for dimensions (width, height, etc) so this could be a way to "write it in px" but actually use rems, with the tradeoff of giving some calculation to the browser. I've mixed feelings about it. What do you think?

guilhermespopolin (approved):

apps/store/src/components/PriceCalculator/CurrentInsuranceField/CurrentInsuranceField.tsx
robinandeer: Ahh I agree with you.

However, these values are decided by Insurely - I should've commented that better. We don't control them - it's just the max values of the iframe that Insurely has decided.

PR #1071, opened 2022-12-12
Create an Icons story to view icons in one place

KarnellSchultz:
## Describe your changes

Create an Icons story to view icons in one place

Needed an easy way to view all of our icons at once. So I made one. 

Screenshot 2022-12-12 at 13 50 27


alebedev (commented):
Nice!  Can you use `require.context` to automatically include all the icons in the folder?  Right now it's easy to forget that every new icon has to be added to a story

KarnellSchultz:
> 

Never heard of a `require.context` are we using that in any other place that you know of? Looking for an example of how it can be used. 

Although, I think I can sort out another way to automatically import the icons we have. I'll take a look at it in a bit.

alebedev:
No current usage, AFAIK.  I'd start with webpack docs (https://webpack.js.org/guides/dependency-management/#requirecontext).  We should not use it NextJs code, but since Storybook is webpack-based, we can use it there

guilhermespopolin (approved):
Great :)
I'm approving it but the idea of load all icons automatically is also nice. I'll stay tuned for the next chapters on that.

KarnellSchultz:
Screenshot 2022-12-13 at 14 46 44

KarnellSchultz:
@guilhermespopolin  @alebedev I came up with another idea of importing all of the icons we have then showing them on the page. This way each time a new icon is added/exported it'll show up in the story as well.

let me know what ya think

PR #1093, opened 2022-12-15
GRW-1838 / Feature / Better formatting API

alebedev:
## Describe your changes

- Move formatting implementation to pure functions
- Use single hook for all formatting needs
- Add currency to countries config so we can render prices without shop session

- Better developer experiece
- Formatters are easy to unit test now


apps/store/src/components/CartNotification/CartNotification.stories.tsx
```diff
@@ -52,17 +50,17 @@ export const WithDialog = TemplateWithDialog.bind({})
...
   name: 'Hedvig Home',
   price: 179,
-  currencyCode: 'SEK',
 }
 
 export default config
 
-const NotificationContent = ({ name, price, currencyCode }: Props) => {
-  const formatter = useCurrencyFormatter(currencyCode)
+// TODO: use CartToast or nothing - does not make sense to have separate presentational component here in story
```
robinandeer, line 57: hmm I think I just wanted to render the content inline instead of having to open the toast to work on styles for it...

apps/store/src/utils/formatter.ts
```diff
@@ -0,0 +1,51 @@
...
+    return new Intl.RelativeTimeFormat(options.locale, { numeric: 'auto' }).format(0, 'day')
+  }
+  return date.toLocaleDateString(options.locale)
+}
+
+type FormatterOptions = MoneyFormatOptions & { i18n: I18NextClient }
+
+export class Formatter {
+  constructor(public options: FormatterOptions) {
+    // Has to be assigned explicitly for Storybook compatibility, looks like it messes up TS constructor properties
```
robinandeer, line 44: Does this mean we can't use:

```
constructor(
    private readonly options: FormatterOptions,
  ) {}
```

...also what does "public" mean in this context?

apps/store/src/utils/formatter.ts
```diff
@@ -0,0 +1,51 @@
...
+    this.options = options
+  }
+
+  amount = (amount: number) => formatAmount(amount, this.options)
```
robinandeer, line 48: nit: we can mark these as public just to be explicit :)

apps/store/src/utils/useFormatter.ts
```diff
@@ -0,0 +1,12 @@
+import { useTranslation } from 'next-i18next'
+import { useMemo } from 'react'
+import { Formatter } from '@/utils/formatter'
+import { useCurrentCountry } from '@/utils/l10n/useCurrentCountry'
+import { useCurrentLocale } from '@/utils/l10n/useCurrentLocale'
+
+export const useFormatter = (): Formatter => {
+  const { locale } = useCurrentLocale()
+  const { currencyCode } = useCurrentCountry()
+  const { i18n } = useTranslation()
```
robinandeer, line 10: why don't we just pass down the translation function?

robinandeer (approved):
Love it! Much improved developer experience :D

Can't see any downside to adding currency to the country config until we go to markets that support multiple currencies or something but we probably shouldn't optimize for that now.

Had some question just to understand better

apps/store/src/utils/formatter.ts
alebedev: Yes, same problem with default webpack config for Storybook.  Easier to work around than fix

apps/store/src/utils/useFormatter.ts
alebedev: I've had problems in the past with i18n.t relying on this, so calling i18n.t() worked and simply t() didn't.  Will check it here and change to just passing t - would be easier to mock

apps/store/src/components/CartNotification/CartNotification.stories.tsx
alebedev: I feel that maybe we have too much storybooks-specific markup.  Will turn this into Tech forum topc

PR #1094, opened 2022-12-15
FIx / Simplify useTranslateFieldLabel

alebedev:
## Describe your changes

Remove unused logic from useTranslateTextLabel
Rename it to useTranslateFieldLabel

Developer experience


robinandeer (approved):
Nice nice!

For me it's also fine to remove the hook until we know we need something like it :)

alebedev:
> Nice nice!
> 
> For me it's also fine to remove the hook until we know we need something like it :)

It felt nice to get `Field -> string` translation function and don't copy-paste key extraction, namespace and defaultValue across call sites

PR #1096, opened 2022-12-15
GRW-1838 / Feature / Use Money object for price data and formatting

alebedev:
## Describe your changes

Always take currency from money object instead of some context

Explicit is better than implicit.  Use same data shape or our GraphQL API - less transform code, lower mental load

### DIFF
--- a/apps/store/src/components/CartInventory/CartInventory.types.ts
+++ b/apps/store/src/components/CartInventory/CartInventory.types.ts
@@ -1,15 +1,15 @@
+import { Money } from '@/utils/formatter'
+
 export type CartEntry = {
   offerId: string
   title: string
-  cost: number
-  currencyCode: string
+  cost: Money
   startDate?: Date
 }
 
 export type CartCampaign = { id: string; displayName: string }
 
 export type CartCost = {
-  currencyCode: string
-  amount: number
-  crossOutAmount?: number
+  total: Money
+  crossOut?: Money
 }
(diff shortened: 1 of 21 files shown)


apps/store/src/components/CartInventory/CartInventory.types.ts
```diff
@@ -1,15 +1,15 @@
...
   title: string
-  cost: number
-  currencyCode: string
+  cost: Money
   startDate?: Date
 }
 
 export type CartCampaign = { id: string; displayName: string }
 
 export type CartCost = {
```
alebedev, line 12: Maybe we should use CartCost from the API?
```
{ net: Money, gross: Money, discount: Money }
```

robinandeer: Hmm, I prefer to transform data and send it to dumb components 🙃

apps/store/src/utils/formatter.ts
```diff
@@ -2,22 +2,27 @@ import { i18n as I18NextClient } from 'i18next'
 import { CurrencyCode } from '@/services/apollo/generated'
 import { IsoLocale } from '@/utils/l10n/types'
 
-type MoneyFormatOptions = { currencyCode: CurrencyCode; locale: IsoLocale }
+type MoneyFormatOptions = { locale: IsoLocale }
 
-export const formatAmount = (amount: number, options: MoneyFormatOptions): string => {
+export type Money = {
```
robinandeer, line 7: nit: alias the type generated from Apollo?

robinandeer (approved):
I Like It

apps/store/src/utils/formatter.ts
alebedev: The problem with alias is that you now have 2 places to import from - IDEs may pick random one

PR #1098, opened 2022-12-15
Integrate discount API explanations

robinandeer:
## Describe your changes

Integrate updated API to display explanations for discounts

## Checklist before requesting a review

- [ ] I have performed a self-review of my code

PR #1099, opened 2022-12-15
Remove layout from CartPage

robinandeer:
## Describe your changes

Remove layout + menu from `CartPage`.

Handle closing cart by navigating back to previous route (or store).

I'm always scared to rely on "history.back" since it doesn't always behave predictably. Like if the history has changes because of "#" updates...

The current solution might be a bit overengineered but I think it's okey.

Probably in the future we should just have a dialog instead of treating it as a separate page.

### DIFF
--- a/apps/store/src/components/CartPage/CartPage.tsx
+++ b/apps/store/src/components/CartPage/CartPage.tsx
@@ -3,7 +3,7 @@ import { useTranslation } from 'next-i18next'
 import Link from 'next/link'
 import { useRouter } from 'next/router'
 import { ReactNode } from 'react'
-import { Button, Heading, LinkButton, Space } from 'ui'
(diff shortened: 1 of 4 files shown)


robinandeer:
* main
  * **PR #1098** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1098" target="_blank">Graphite</a> 
    * **PR #1099** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1099" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1099?utm_source=stack-comment).

alebedev (approved):
Looks like there may be too much magic, but I don't have any specific suggestions.  Let's do it and see how it works

PR #1104, opened 2022-12-15
Feature / Localize country and language labels

alebedev:
## Describe your changes

Localize country selector page and footer selector block

Screenshot 2022-12-15 at 17 12 19

### DIFF
--- a/apps/store/src/blocks/FooterBlock.tsx
+++ b/apps/store/src/blocks/FooterBlock.tsx
@@ -1,5 +1,6 @@
 import styled from '@emotion/styled'
 import { storyblokEditable } from '@storyblok/react'
+import { useTranslation } from 'next-i18next'
 import Link from 'next/link'
 import { useRouter } from 'next/router'
 import { ChangeEvent, useRef } from 'react'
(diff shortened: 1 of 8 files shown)


apps/store/src/blocks/FooterBlock.tsx
```diff
@@ -68,17 +64,18 @@ export const FooterBlock = ({ blok }: FooterBlockProps) => {
   const formRef = useRef<HTMLFormElement>(null)
   const { language: currentLanguage } = useCurrentLocale()
   const currentCountry = useCurrentCountry()
+  const { t } = useTranslation()
   const cookiePersister = new CookiePersister(LOCALE_COOKIE_KEY)
 
   const countryList = Object.keys(countries).map((country) => ({
-    name: TEMP_TRANSLATIONS[`COUNTRY_LABEL_${country}`],
+    name: t(`COUNTRY_LABEL_${country}`, { defaultValue: country }),
```
robinandeer, line 71: did we discuss if we should avoid dynamic keys?

if we do, we can run code analysis to eliminate unused keys in the future, right?

robinandeer (approved):
Nice improvement :)

apps/store/src/blocks/FooterBlock.tsx
alebedev: Let's use them for now - it's an easy way.  And we can always go to switch statements when we integrate used keys analysis

PR #1127, opened 2022-12-19
GRW-1918 - styles: split perils into columns

guilhermespopolin:
## Describe your changes

* UI changes: split PerilsBlock content into columns

Requested by design

Screenshot 2022-12-19 at 11 18 39
Screenshot 2022-12-19 at 11 18 21

### DIFF
--- a/apps/store/src/components/Perils/Perils.tsx
+++ b/apps/store/src/components/Perils/Perils.tsx
@@ -1,15 +1,38 @@
 import styled from '@emotion/styled'
-import React, { useState, useCallback } from 'react'
+import React, { useState, useCallback, ReactNode } from 'react'
+import { mq } from 'ui'
 import * as Accordion from '@/components/Accordion/Accordion'
 import { SpaceFlex } from '@/components/SpaceFlex/SpaceFlex'
 import { CoverageList } from './CoverageList'
 import { Peril } from './Perils.types'
+
+const MAX_COLS = 4
 
 type Props = {
   items: Array<Peril>
 }
 
 export const Perils = ({ items }: Props) => {
+  const numberOfItemsPerColumn = Math.ceil(items.length / MAX_COLS)
+  const accordions: Array<ReactNode> = []
+  for (let i = 0; i < items.length; i += numberOfItemsPerColumn) {
+    const perils = items.slice(i, i + numberOfItemsPerColumn)
+    accordions.push(<PerilsAccordion key={i} perils={perils} />)
+  }
+
+  return <PerilsAccordionGrid fixedCols={items.length < 4}>{accordions}</PerilsAccordionGrid>
+}
+
+const PerilsAccordionGrid = styled.div(({ fixedCols = false }: { fixedCols?: boolean }) => ({
+  display: 'grid',
+  gap: '0.25rem',
+  [mq.md]: {
+    gridTemplateColumns: `repeat(auto-fit, ${fixedCols ? '20.75rem' : 'minmax(20.75rem, 1fr)'})`,
+    gap: '1rem',
+  },
+}))
+
+const PerilsAccordion = ({ perils }: { perils: Array<Peril> }) => {
   const [openedItems, setOpenedItems] = useState<Array<string>>()
 
   const handleValueChange = useCallback((value: Array<string>) => {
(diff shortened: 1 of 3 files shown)


guilhermespopolin:
* main
  * **PR #1127** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1127" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1127?utm_source=stack-comment).

apps/store/src/components/Perils/Perils.tsx
```diff
@@ -1,15 +1,47 @@
...
 import { Peril } from './Perils.types'
 
+const MAX_COLS = 4
+
 type Props = {
   items: Array<Peril>
 }
 
 export const Perils = ({ items }: Props) => {
+  const matchesMdScreen = useBreakpoint('md')
```
robinandeer, line 16: There's no way around using this, right? Like just stacking the sections on top of one another on mobile? I feel like we can start with that if it means we can handle the responsiveness in CSS.

robinandeer (commented):
Good stuff! Just added a comment if it's possible to deal with responsiveness using CSS

gustaveen:
Nice @guilhermespopolin 🎉
I saw you had a discussion about this on friday. The problem was that expanding one accordion pushed content down in other columns of the grid right if I understood it correctly?

apps/store/src/components/Perils/Perils.tsx
```diff
@@ -1,15 +1,47 @@
...
+      <PerilsAccordion perils={items} />
+    </PerilsAccordionGrid>
+  )
+}
+
+const PerilsAccordionGrid = styled.div(({ fixedCols = false }: { fixedCols?: boolean }) => ({
+  display: 'grid',
+  [mq.md]: {
+    gridTemplateColumns: `repeat(auto-fit, ${fixedCols ? '20.75rem' : 'minmax(20.75rem, 1fr)'})`,
+    gap: 16,
```
gustaveen, line 40: Slightly adjust the row-gap for md breakpoints :) 
Screenshot 2022-12-19 at 11 47 23
```suggestion
    gap: 0.25rem 1rem,
```

apps/store/src/components/Perils/Perils.tsx
guilhermespopolin: Yeah, I couldn't find a way of solving it only using CSS 😞 
That's because the responsive design on this case also mean change the markup: we start with every peril being part of the the same accordion, but then change on desktop where we have "groups" (columns) where each one of them has its own accordion.

guilhermespopolin:
> Nice @guilhermespopolin 🎉 I saw you had a discussion about this on friday. The problem was that expanding one accordion pushed content down in other columns of the grid right if I understood it correctly?

Precis. To be honest I'm sure if that slack discussion is still valid because I had miss interpreted how that block was suppose to work (design wise). I thought we're aiming for a _[masonry layout](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Grid_Layout/Masonry_Layout#:~:text=Masonry%20layout%20is%20a%20layout,to%20completely%20fill%20the%20gaps.)_ but that seems not to be the case.

apps/store/src/components/Perils/Perils.tsx
gustaveen: Gotcha! Yeah omit this if this is what you decided with Zak and co 🙌

gustaveen:
> > Nice @guilhermespopolin 🎉 I saw you had a discussion about this on friday. The problem was that expanding one accordion pushed content down in other columns of the grid right if I understood it correctly?
> 
> Precis. However, to be honest I'm not sure if that slack discussion is still valid because I had miss interpreted how that block was suppose to work (design wise). I thought we're aiming for a _[masonry layout](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Grid_Layout/Masonry_Layout#:~:text=Masonry%20layout%20is%20a%20layout,to%20completely%20fill%20the%20gaps.)_ but that seems not to be the case.

I see, yeah I though we were going masonry as well :D

apps/store/src/components/Perils/Perils.tsx
gustaveen: True, that could work right? Since we don't have any "close all" functionality, it wouldn't matter if they belonged to multiple groups I suppose :)

guilhermespopolin: Yep. That works. I'm updating it. Thanks :)

robinandeer (approved):
Nice! You made it work ^^

gustaveen (approved):
GJ @guilhermespopolin 🎉

PR #1137, opened 2022-12-20
Avoid infinite loop of creating price intents after API error

robinandeer:
## Describe your changes

Avoid causing infinite loop when getting API error after creating new price intents.

We didn't consider the case where the API would return an error after creating a new price intent. This would cause an infinite loop.

We still have the problem that we run the mutation twice, but that's a separate issue.


robinandeer:
* main
  * **PR #1137** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1137" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1137?utm_source=stack-comment).

KarnellSchultz (approved):
loooks good to me 🔄

PR #1152, opened 2022-12-21
GRW-1693 / Feature / Update experiment_impression tracking

alebedev:
## Describe your changes

- Support single experiment AB testing vs old site
- Track experiment_impression events based on local cookie, based on redirect query parameter

- Support new AB testing model
- Don't rely on parent domain cookie, it's painful to test.  We've got enough when testing payment connection


robinandeer (approved):
Looks good as far as I can tell -- will be fun to hear an introduction!

PR #1156, opened 2022-12-22
Add Hedvig.com Button ✨

robinandeer:
## Describe your changes

Implement button from *New* .com UI Kit.

Some translucent colors are missing from theme so used solid colors for now.

Add slight transition + focus-visible states.


robinandeer:
* main
  * **PR #1156** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1156" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1156?utm_source=stack-comment).

guilhermespopolin (approved):
Nice 💅🏻

apps/store/src/components/Button/Button.tsx
```diff
@@ -0,0 +1,164 @@
...
+    fontSize: theme.fontSizes[1],
+  }),
+
+  cursor: 'pointer',
+  '&:disabled': {
+    cursor: 'default',
+  },
+
+  '&:focus-visible': {
+    boxShadow: `0 0 0 2px ${theme.colors.textPrimary}`,
```
gustaveen, line 83: nit: let's add a semantic color for shadows or a general primary color :)

apps/store/src/components/Button/Button.tsx
```diff
@@ -0,0 +1,164 @@
...
+const StyledButton = styled.button<CustomProps>(({ theme, size = 'large' }) => ({
+  // opt out of double tap to zoom to immediately respond to taps
+  touchAction: 'manipulation',
+
+  borderRadius: theme.radius.sm,
+
+  whiteSpace: 'nowrap',
+  lineHeight: 1,
+  transition:
+    'background 0.1s ease-out 0s, box-shadow 0.15s cubic-bezier(0.47, 0.03, 0.49, 1.38) 0s',
```
gustaveen, line 54: just curious, where did we get the transition settings from :)

apps/store/src/components/Button/Button.tsx
```diff
@@ -0,0 +1,164 @@
...
+
+  ...(size === 'large' && {
+    width: '100%',
```
gustaveen, line 57: Should we have a `fullWidth` prop instead? I think we want to be able to have a large button displayed inline right :)

gustaveen (approved):
Wohoo - great work Robin!! 🎉
Let's move this one into UI kit and move the button we have there to onboarding

apps/store/src/components/Button/Button.tsx
robinandeer: We should check with Zak -- didn't seem like it from Figma

apps/store/src/components/Button/Button.tsx
robinandeer: Yep! I actually think I found this from Figma though 😅

apps/store/src/components/Button/Button.tsx
robinandeer: Hmmm from atlassian: https://atlassian.design/components/button/examples

robinandeer:
Agreed that we should replace the button from UI kit -- I will follow up with that!

PR #1157, opened 2022-12-22
GRW-1798 / Feature / Tracking service

alebedev:
## Describe your changes

- Track -> Tracking service
- Create instance explicitly and inject through provider, similar to ShopSession service
- Move event tracking from GTM to Tracking, leave GTM as low-level adapter

Provide better abstraction for reporting analytics events


robinandeer (commented):
Great work!

I added a bunch of comments -- mostly related to me being worried about setting up a bunch of `useEffect` hooks that are hard to reason about. I feel like we can find more accurate places to track things :)

apps/store/src/components/ProductPage/PurchaseForm/PurchaseForm.tsx
```diff
@@ -182,14 +182,7 @@ const EditingState = (props: EditingStateProps) => {
...
       if (updatedPriceIntent) {
-        // FIXME: pick offer for specific product or track all offers
-        const firstOffer = updatedPriceIntent.offers[0]
-        trackOffer({
-          shopSessionId: shopSession!.id,
-          contractType: firstOffer.variant.typeOfContract,
-          amount: firstOffer.price.amount,
-          currency: firstOffer.price.currencyCode,
-        })
+        updatedPriceIntent.offers.forEach((offer) => tracking.reportOffer(offer))
```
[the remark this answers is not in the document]
alebedev: Backend cannot report to Google Analytics, AFAIK

apps/store/src/services/Tracking/TrackingContext.tsx
```diff
@@ -0,0 +1,51 @@
...
+}
+
+export const useTrackShopSession = (tracking: Tracking) => {
+  const { shopSession } = useShopSession()
+  const shopSessionId = shopSession?.id
+  useEffect(() => {
+    if (shopSessionId) {
+      tracking.setContext('shopSessionId', shopSessionId)
+    }
+  }, [shopSessionId, tracking])
```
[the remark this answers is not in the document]
alebedev: Makes sense, to be fixed in next PR

apps/store/src/services/Tracking/TrackingContext.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+  return <TrackingContext.Provider value={tracking}>{children}</TrackingContext.Provider>
+}
+
+const useMaintainCountryContext = (tracking: Tracking) => {
+  const { countryCode } = useCurrentCountry()
+  tracking.setContext('countryCode', countryCode)
+  useEffect(() => {
+    setGtmContext(countryCode)
+  }, [countryCode])
```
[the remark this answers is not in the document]
alebedev: This is what we're effectively doing here, but reusing existing `useCountry` that depends on router.  Do you see an easy way to handle this outside React?

guilhermespopolin:
> Great work!
> 
> I added a bunch of comments -- mostly related to me being worried about setting up a bunch of `useEffect` hooks that are hard to reason about. I feel like we can find more accurate places to track things :)

Just to give more food for the discussion _if_ I properly understood your point, the issue with tracking stuff by setting up a bunch of `useEffect` hooks is that they end up being to vague when it comes about when to track things - track when some value changes. With that said, can we agree that the ideal situation would be being more explicitly on the matter by

* Track things imperatively and/or;
* Track things as a response to an event?

apps/store/src/pages/_app.tsx
```diff
@@ -3,32 +3,45 @@ import { Global, ThemeProvider } from '@emotion/react'
...
 if (process.env.NEXT_PUBLIC_API_MOCKING === 'enabled') {
   require('../mocks')
 }
 
-if (typeof window === 'undefined') {
-  Datadog.initDatadog()
-} else {
-  trackPageView(window.location.pathname)
-  trackNewSiteExperimentImpression()
+const tracking = new Tracking()
```
robinandeer, line 31: Very nice! 😄

apps/store/src/services/Tracking/Tracking.ts
```diff
@@ -0,0 +1,89 @@
+import { ProductOfferFragment } from '@/services/apollo/generated'
+import { AppTrackingContext, pushToGTMDataLayer, setGtmContext } from '@/services/gtm'
+import { isBrowser } from '@/utils/env'
+import { newSiteAbTest } from '../../newSiteAbTest'
```
robinandeer, line 4: where is this imported from? Did you miss `@/services/`?

apps/store/src/services/Tracking/Tracking.ts
```diff
@@ -0,0 +1,89 @@
...
+    this.context[key] = value
+  }
+
+  public setAppContext = (context: AppTrackingContext) => {
+    this.setContext('countryCode', context.countryCode)
+    setGtmContext(context)
+  }
+
+  public reportPageView(urlPath: string) {
+    this.ensureBrowserEnvironment()
```
robinandeer, line 31: Should we really throw an error here? Would be really bad UX to crash the app just because we setup analytics correctly, no?

apps/store/src/services/Tracking/TrackingContext.tsx
```diff
@@ -0,0 +1,21 @@
...
+export const TrackingContext = createContext(new Tracking())
+
+export const TrackingProvider = ({ value: tracking, children }: ProviderProps<Tracking>) => {
+  useTrackShopSession(tracking)
+
+  return <TrackingContext.Provider value={tracking}>{children}</TrackingContext.Provider>
+}
+
+export const useTrackShopSession = (tracking: Tracking) => {
+  const { shopSession } = useShopSession()
```
robinandeer, line 14: nit: I would expose a param `onCompleted` here that returns the up-to-date shop session so we can set tracking context :)

apps/store/src/services/Tracking/newSiteExperimentTracking.ts
```diff
@@ -0,0 +1,26 @@
+// Track experiment impression based on local cookie set by router
+import { getCookie, setCookie } from 'cookies-next'
+import router from 'next/router'
+import { Tracking } from '@/services/Tracking/Tracking'
```
robinandeer, line 4: This could be a relative import since it's within the same service :)

apps/store/src/services/Tracking/trackPageViews.ts
```diff
@@ -0,0 +1,13 @@
+import router from 'next/router'
+import { Tracking } from '@/services/Tracking/Tracking'
+
+export const trackPageViews = (tracking: Tracking) => {
+  router.ready(() => {
+    tracking.reportPageView(window.location.pathname)
+    router.events.on('routeChangeComplete', (url: string, { shallow = false } = {}) => {
```
robinandeer, line 7: is this safe? We never clean up this subscription, right?

Why not set this up inside react useEffect?

robinandeer (approved):
Much improved, I find it pretty easy to understand the setup now :D

Just left some comments how to improve even further

apps/store/src/services/Tracking/TrackingContext.tsx
```diff
@@ -0,0 +1,21 @@
+import { createContext, ProviderProps, useEffect } from 'react'
+import { useShopSession } from '@/services/shopSession/ShopSessionContext'
+import { Tracking } from '@/services/Tracking/Tracking'
+
+export const TrackingContext = createContext(new Tracking())
+
+export const TrackingProvider = ({ value: tracking, children }: ProviderProps<Tracking>) => {
+  useTrackShopSession(tracking)
```
guilhermespopolin, line 8: 🤔 It feels this call belongs to `_app/MyApp` component. Not a big deal though.

guilhermespopolin (approved):
👌🏻

apps/store/src/services/Tracking/TrackingContext.tsx
alebedev: Agree, will do in ShopSession PR

apps/store/src/services/Tracking/Tracking.ts
alebedev: Assuming we call analytics only from event handlers, this should never be called server-side.  If it does, it's programmer error and sooner we crash, sooner we detect and fix it

apps/store/src/services/Tracking/TrackingContext.tsx
alebedev: Needs to be inside ShopSessionProvider to see the session

PR #1159, opened 2022-12-22
GRW-1967/Display discount applied state

KarnellSchultz:
## Describe your changes

- Added logic to show a discount duration message for several states
- updated the Storybook

Screenshot 2022-12-22 at 15 28 00

Match the designs

[GRW-1967]: https://hedvig.atlassian.net/browse/GRW-1967?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/store/src/components/CartInventory/CartInventory.helpers.tsx
```diff
@@ -27,3 +28,24 @@ export const useGetDiscountExplaination = () => {
...
+
+export const useGetDiscountDurationExplanation = () => {
+  const { t } = useTranslation('cart')
+  const formatter = useFormatter()
+
+  return (discount: CampaignDiscount, total: Money) => {
+    switch (discount.type) {
+      case CampaignDiscountType.FreeMonths:
+      case CampaignDiscountType.MonthlyPercentage:
+        // A workaround to escape the '/' in the monthly price.
```
KarnellSchultz, line 40: @robinandeer and I had a chat about this. Looking for feedback/ideas/solutions.

**The issue**
when attempting to do something like this: 
```
t('DISCOUNT_DURATION_EXPLANATION', {
          count: discount.months,
          total:formatter.monthlyPrice(total)
        })
```
The forward slash is not properly escaped. [i18next docs](https://www.i18next.com/translation-function/interpolation)

Any ideas on the best way to handle that? 

Screenshot 2022-12-22 at 10 49 10

KarnellSchultz:
Found a bug with the free months discount during our standup demo. Fix it and here is a video of how it works now. 


https://user-images.githubusercontent.com/50870173/209308472-6e5b92d0-a147-4843-bb52-a773d1ba5907.mov

apps/store/src/components/CartInventory/CartInventory.helpers.tsx
alebedev: What happens with `{{- value}}` and `interpolation: { escapeValue: false }` options from the docs?

apps/store/public/locales/default/cart.json
```diff
@@ -9,5 +9,7 @@
   "DISCOUNT_STATE_FREE_MONTHS": "Free for {{count}} mth.",
   "DISCOUNT_STATE_MONTHLY_PERCENTAGE": "{{percentage}}% discount for {{count}} mth.",
   "DISCOUNT_STATE_INDEFINITE_PERCENTAGE": "{{percentage}}% discount",
-  "CHECKOUT_ADD_DISCOUNT_BUTTON": "Add"
-}
\ No newline at end of file
+  "CHECKOUT_ADD_DISCOUNT_BUTTON": "Add",
+  "DISCOUNT_DURATION_EXPLANATION": "For {{count}} months, then",
```
alebedev, line 13: Let's add singular form to localize, so it would be "1 month" and "2 months"

apps/store/src/components/CartInventory/CartInventory.helpers.tsx
KarnellSchultz: I was playing with that previously and it works. My issues with that were that we were somehome opening ourselves up to XSS attacks like they mention in the docs. Also it's not super visible.

apps/store/src/components/CartInventory/CartInventory.stories.tsx
```diff
@@ -21,17 +21,153 @@ export default {
...
+const MOCK_CART_TWO: CartFragmentFragment = {
+  id: '950119de-7639-4c56-bdc4-613c1a4e1705',
+  cost: {
+    gross: {
+      amount: 307,
+      currencyCode: CurrencyCode.Sek,
+      __typename: 'Money',
+    },
+    net: {
+      amount: 307,
```
alebedev, line 48: Nit: gross and discount don't add up to net here

apps/store/src/components/CartInventory/CartInventory.tsx
```diff
@@ -20,12 +27,26 @@ export const CartInventory = ({ cart, children }: Props) => {
...
       <Footer>
-        <Heading as="h3" variant="standard.18">
-          Total
-        </Heading>
-        <Heading as="h3" variant="standard.18">
-          {formatter.monthlyPrice(cart.cost.net)}
-        </Heading>
+        <TotalWrapper>
+          <Heading as="h3" variant="standard.18">
+            Total
```
alebedev, line 32: Should we localize it?

apps/store/src/components/CartInventory/CartInventory.tsx
```diff
@@ -20,12 +27,26 @@ export const CartInventory = ({ cart, children }: Props) => {
...
+          </Heading>
+          <PriceWrapper>
+            <OriginalPrice as="h3" variant="standard.18">
+              {formatter.monthlyPrice(cart.cost.net)}
```
alebedev, line 36: Isn't original cost `gross` and becoming `net` after discount?

apps/store/src/components/CartInventory/CartInventory.tsx
```diff
@@ -20,12 +27,26 @@ export const CartInventory = ({ cart, children }: Props) => {
...
+            </OriginalPrice>
+            <Heading as="h3" variant="standard.18">
+              {formatter.monthlyPrice(cart.cost.discount)}
```
alebedev, line 39: What if we don't have any discounts?  Should we just display net price without gross and discount?

apps/store/src/components/CartInventory/CostSummary.tsx
```diff
@@ -1,33 +1,62 @@
...
 }
 
-const SpaceBetween = styled.div({
+const SpaceBetweenWrapper = styled.div({
   display: 'flex',
   justifyContent: 'space-between',
-  alignItems: 'center',
+})
+
+const CostWrapper = styled.div({
```
alebedev, line 45: Nit: can we use some SpaceFlex or styled(Space) to avoid extra wrapping in DOM?

apps/store/src/components/CartInventory/CostSummary.tsx
```diff
@@ -1,33 +1,62 @@
...
+  justifyContent: 'space-between',
+  flexDirection: 'column',
 })
 
 const CrossOutText = styled(Text)({ textDecoration: 'line-through' })
+
+const PriceWrapper = styled.div(({ theme }) => ({
+  display: 'flex',
+  justifyContent: 'flex-end',
+  gap: theme.space[2],
```
alebedev, line 56: Does gap work out of display: grid context?

apps/store/src/pages/cart.tsx
```diff
@@ -35,21 +40,42 @@ const NextCartPage: NextPageWithLayout<Props> = (props) => {
...
+      shopSession.cart.redeemedCampaigns[0].discount,
+      shopSession.cart.cost.gross,
+    ),
   }))
 
   const cartCost = shopSession.cart.cost
-  const crossOut = cartCost.gross.amount !== cartCost.net.amount ? cartCost.gross : undefined
+  const hasDiscount = shopSession.cart.redeemedCampaigns.length
+
+  const total =
```
alebedev, line 53: Multiline ternaries are tricky to get right.  Can you simplify it or replace with if/else?

apps/store/src/pages/cart.tsx
```diff
@@ -35,21 +40,42 @@ const NextCartPage: NextPageWithLayout<Props> = (props) => {
...
+    hasDiscount &&
+    shopSession.cart.redeemedCampaigns[0].discount.type === CampaignDiscountType.FreeMonths
```
alebedev, line 55: Can silently become wrong if we have multiple discounts and FreeMonths is in position 2+

Ideas
- Check all values with `.some()`
- Document that we only expect single discount and print warning if there're multiple.  I guess it's OK for launch - no need to spend time supporting multiple discounts until we have them

Related slack discussion: https://hedviginsurance.slack.com/archives/C041SD67G82/p1671789253334779

apps/store/src/pages/cart.tsx
```diff
@@ -35,21 +40,42 @@ const NextCartPage: NextPageWithLayout<Props> = (props) => {
...
+      ? cartCost.discount
+      : cartCost.net
+
+  const getCrossOut = () => {
+    if (hasDiscount) {
```
alebedev, line 60: Nit: easier to read with early return
```
if (!hasDiscount) return
```

alebedev (commented):
Great work here! 

Sorry for number of comments.  I suggest we make math and texts right here and fix styles in follow-up PR to make this one smaller

apps/store/src/components/CartInventory/CartInventory.tsx
KarnellSchultz: Updated that now.

In the case of a percentage discount that's correct. 

However in the "free months" case. The net and gross are the same and the user would be paying the discount price, which is zero.

apps/store/src/components/CartInventory/CostSummary.tsx
KarnellSchultz: Yeah, it works with flex and grid 👍

apps/store/src/pages/cart.tsx
KarnellSchultz: Turned this into a if statement to simplify it.

apps/store/src/pages/cart.tsx
KarnellSchultz: Great point - Going to leave a comment here and come back to this a bit further down the line.

apps/store/src/components/CartInventory/CartInventory.stories.tsx
KarnellSchultz: Fixed that now

apps/store/src/components/CartInventory/CartInventory.tsx
KarnellSchultz: Ended up creating a new component to handle the logic for when there is no discount

KarnellSchultz:
@alebedev Thanks for taking the time to make comments on this PR. I've attempted to address each of your comments. Please continue to comment as you see fit 👍 and let me know if you have any further suggestions.

alebedev (approved):
Great work!

apps/store/src/components/CartInventory/CartInventory.stories.tsx
```diff
@@ -21,25 +21,161 @@ export default {
...
+        {(offer) => <div>{offer.variant.product.displayNameFull}</div>}
+      </CartInventory>
+    </>
   )
 }
 
 export const Default = Template.bind({})
 Default.args = {}
 
+const MOCK_CART_TWO: CartFragmentFragment = {
```
guilhermespopolin, line 39: It's just OCD 😅 
Shouldn't `MOCK_CART_TWO` be placed bellow `MOCK_CART` hehe ?

apps/store/src/components/CartInventory/CartInventory.tsx
```diff
@@ -1,16 +1,23 @@
...
 
 type Props = {
   cart: CartFragmentFragment
   children: (offer: ProductOfferFragment) => ReactNode
 }
 
 export const CartInventory = ({ cart, children }: Props) => {
-  const formatter = useFormatter()
+  const { t } = useTranslation('cart')
+  const discountDurationExplanation = useGetDiscountDurationExplanation()
```
guilhermespopolin, line 16:
```suggestion
  const getDiscountDurationExplanation = useGetDiscountDurationExplanation()
```

apps/store/src/components/CartInventory/CartInventory.tsx
```diff
@@ -20,17 +27,57 @@ export const CartInventory = ({ cart, children }: Props) => {
...
+          </Heading>
+          <DisplayTotalAmount cost={cart.cost} hasDiscount={hasDiscount} />
+        </TotalWrapper>
+        {campaigns
+          ? campaigns.map((campaign) => (
+              <DiscountMessage key={campaign.discount.type}>
+                {discountDurationExplanation(campaign.discount, cart.cost.gross)}
+              </DiscountMessage>
+            ))
+          : null}
```
guilhermespopolin, line 42: I guess we don't need to check the "presence" of `campaigns` here:

```suggestion
        {campaigns.map((campaign) => (
              <DiscountMessage key={campaign.discount.type}>
                {discountDurationExplanation(campaign.discount, cart.cost.gross)}
              </DiscountMessage>
            ))}
```

apps/store/src/components/CartInventory/CartInventory.tsx
```diff
@@ -20,17 +27,57 @@ export const CartInventory = ({ cart, children }: Props) => {
...
       </Footer>
     </Space>
   )
 }
 
+type DisplayTotalAmountProps = {
+  cost: CartFragmentFragment['cost']
+  hasDiscount: boolean
```
guilhermespopolin, line 50: very nit: Since we're defining a a default value for `hasDiscount` maybe it should be optional.

```suggestion
  hasDiscount?: boolean
```

apps/store/src/components/CartInventory/CartInventory.tsx
```diff
@@ -20,17 +27,57 @@ export const CartInventory = ({ cart, children }: Props) => {
...
+      {!hasDiscount ? (
+        <Text>{formatter.monthlyPrice(gross)}</Text>
+      ) : (
+        <SpaceBetweenPrice>
+          <LineThroughPrice>{formatter.monthlyPrice(gross)}</LineThroughPrice>
+          <Text>{formatter.monthlyPrice(net)}</Text>
+        </SpaceBetweenPrice>
+      )}
+    </>
+  )
```
guilhermespopolin, line 67: Just a suggestion:

```suggestion
if (hasDiscount) {
  <SpaceBetweenPrice>
    <LineThroughPrice>{formatter.monthlyPrice(gross)}</LineThroughPrice>
    <Text>{formatter.monthlyPrice(net)}</Text>
  </SpaceBetweenPrice>
}

return <Text>{formatter.monthlyPrice(gross)}</Text>
```

guilhermespopolin (approved):
Great stuff 👌🏻 
Just added a few comments for optional small things.

apps/store/src/components/CartInventory/CartInventory.stories.tsx
KarnellSchultz: Yeah for sure

apps/store/src/components/CartInventory/CartInventory.tsx
KarnellSchultz: I like that 👍

apps/store/src/components/CartInventory/CartInventory.tsx
KarnellSchultz: I've got a little habit of adding default values. 

Here I decided to remove the default value and keep `hasDiscount` as a required prop.

PR #1165, opened 2022-12-23
Replace UI Kit Button

robinandeer:
## Describe your changes

Move Button from "store" to "ui".
Update all buttons to use the new Button component.

Move Button from "ui" to "onboarding".

I had to add some properties to the legacy theme - otherwise the typecheck fails. Let me know if there's a way to avoid this.

I talked to Aleks about not exposing Emotion from the "ui" package. Therefore we don't use the "as" prop but instead wrap the Button in the Next link component as pass down the href prop.

I probably didn't catch all the edge cases but better to merge this now and fix them later.


robinandeer:
* main
  * **PR #1165** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1165" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1165?utm_source=stack-comment).

packages/ui/src/components/Button/Button.tsx
```diff
@@ -1,147 +1,173 @@
...
-    type?: 'button' | 'submit'
-    disabled?: boolean
-    form?: string
-  }
+import { ButtonHTMLAttributes, forwardRef } from 'react'
+import { LoadingSpinner } from './LoadingSpinner'
 
-type IconWrapperProps = {
-  padded?: boolean
+export type CustomButtonProps = {
```
alebedev, line 6: StoreButtonProps?

packages/ui/src/components/Button/Button.tsx
```diff
@@ -1,147 +1,173 @@
...
-  marginLeft: padded ? '0.5rem' : 0,
-  display: 'flex',
-}))
+type Props = ButtonHTMLAttributes<HTMLButtonElement> & CustomButtonProps
+
+export const Button = forwardRef<HTMLButtonElement, Props>((props, ref) => {
+  const { variant = 'primary', loading, children, ...baseProps } = props
+
+  const theme = useTheme()
+  const buttonChildren = loading ? (
```
alebedev, line 19: We have a lot of uses with `disabled={loading}` do we want to change them to loading in the future?

packages/ui/src/components/Button/Button.tsx
```diff
@@ -1,147 +1,173 @@
...
+  href?: string
 }
 
-const IconWrapper = styled.div<IconWrapperProps>(({ padded }) => ({
-  marginLeft: padded ? '0.5rem' : 0,
-  display: 'flex',
-}))
+type Props = ButtonHTMLAttributes<HTMLButtonElement> & CustomButtonProps
+
+export const Button = forwardRef<HTMLButtonElement, Props>((props, ref) => {
```
alebedev, line 15: Design question - is it better to have one big component that does everything or export some specialized versions like `LinkButton` or `LoadingButton`?

alebedev (approved):
Looks good, big one!  Some questions inside

apps/store/src/components/CartNotification/CartToast.tsx
```diff
@@ -46,19 +46,15 @@ const CartNotificationContent = ({ name, price, onClose }: Props) => {
         <ProductItem name={name} price={price} />
 
         <Space y={0.5}>
-          <LinkButton as={Link} href={PageLink.cart()} fullWidth>
-            Proceed to cart ({cartLineCount})
-          </LinkButton>
+          <Link href={PageLink.cart()} passHref legacyBehavior>
```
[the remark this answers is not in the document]
robinandeer: Yes I think it's a good idea to wrap it! Will do that in a follow up PR

packages/ui/src/components/Button/Button.tsx
robinandeer: Hmmm what do you mean? This is from the UI kit now :)

Perhaps deserves a better name tho... any ideas?

packages/ui/src/components/Button/Button.tsx
robinandeer: I wanted to export a "LinkButton" but couldn't work out the types so ended up with this

Loading is a state of the button so that should be part of the same component I think :)

packages/ui/src/components/Button/Button.tsx
robinandeer: Yes - showing pending state by disabling the button was only a placeholder

packages/ui/src/components/Button/Button.tsx
alebedev: UiButtonProps is my best idea and I don't like it )

Let's keep it for now

PR #1168, opened 2022-12-23
Fix / Fix ShopSessionContext memozing stale value

alebedev:
## Describe your changes

Replace useMemo with ref in ShopSessionContext to avoid returning stale values

Reduce number of renders by utilizing lazy query which we run after we populate Apollo cache from outside React

Stepping stone to fixing double run of priceIntent mutations

### DIFF
--- a/apps/store/src/services/shopSession/ShopSessionContext.tsx
+++ b/apps/store/src/services/shopSession/ShopSessionContext.tsx
@@ -1,6 +1,7 @@
 import { useApolloClient } from '@apollo/client'
-import { createContext, PropsWithChildren, useContext, useEffect, useMemo } from 'react'
-import { ShopSessionQueryResult, useShopSessionQuery } from '@/services/apollo/generated'
+import { createContext, PropsWithChildren, useContext, useEffect, useMemo, useRef } from 'react'
+import { ShopSessionQueryResult, useShopSessionLazyQuery } from '@/services/apollo/generated'
+import { ShopSession } from '@/services/shopSession/ShopSession.types'
 import { isBrowser } from '@/utils/env'
 import { useCurrentCountry } from '@/utils/l10n/useCurrentCountry'
 import { setupShopSessionServiceClientSide } from './ShopSession.helpers'
@@ -14,7 +15,9 @@ export const ShopSessionProvider = ({ children, shopSessionId: initialShopSessio
   return <ShopSessionContext.Provider value={contextValue}>{children}</ShopSessionContext.Provider>
 }
 
-export const useShopSession = () => {
+type ShopSessionResult = ShopSessionQueryResult & { shopSession?: ShopSession }
+
+export const useShopSession = (): ShopSessionResult => {
   const { countryCode } = useCurrentCountry()
 
   const queryResult = useContext(ShopSessionContext)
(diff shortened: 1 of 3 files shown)


robinandeer (commented):
Really appreciate your work on this!

Just a question when using ref -- does that still cause components to re-render when they need to?

apps/store/src/services/shopSession/ShopSessionContext.tsx
```diff
@@ -24,37 +27,45 @@ export const useShopSession = () => {
...
-    return {
-      ...queryResult,
-      // Ignore session from different country.
-      // This probably means old session was fetched and new one is being created
-      shopSession: shopSession?.countryCode === countryCode ? shopSession : undefined,
-    }
-  }, [queryResult, countryCode])
+  // Make sure result object is stable
+  const result = useRef<ShopSessionResult>({} as ShopSessionResult)
+  Object.assign(result.current, queryResult)
```
guilhermespopolin, line 32: Just curious: why not just do `result.current = queryResult`? 🤔

alebedev: We want to maintain identity between renders, reassigning result.current will break "should I render" checks down in child components using session context

apps/store/src/services/shopSession/ShopSessionService.ts
```diff
@@ -75,6 +75,17 @@ export class ShopSessionService {
     >({
       mutation: ShopSessionCreateDocument,
       variables,
+      // TODO: Investigate if we can do it by returning shopSession recond on top instead of shopSessionCreate
```
robinandeer, line 78: nit: `rescond` => `response`

apps/store/src/services/shopSession/ShopSessionContext.tsx
```diff
@@ -24,37 +27,46 @@ export const useShopSession = () => {
...
+        })
+      })
     }
-  }, [countryCode, shopSessionService])
+  }, [shopSessionServiceClientSide, countryCode, fetchSession])
+
+  // Fetch once during SSR
+  if (!isBrowser() && initialShopSessionId && !queryResult.called) {
+    fetchSession({ variables: { shopSessionId: initialShopSessionId } })
+  }
```
robinandeer, line 69: Can you help me understand why this is better than the recommended apollo approach?

`useShopSeesionQuery({ ssr: true })`

robinandeer (commented):
This is not too many changes! Looking promising :D

apps/store/src/services/shopSession/ShopSessionContext.tsx
robinandeer: So I see in the logs that we do two renders (on server and client)

1. queryResult = undefined
2. queryResult = defined

Because React has started coalescing state updates this doesn't cause any visual artifacts (my guess) => i.e. the returned shop session is always defined when we prefetch it server-side.

This does feel a little scary though, that we are relying on a React feature to coalesce state updates to avoid rendering before we had time to fetch the result from cache.

robinandeer: React components are supposed to be pure functions and we violate this here... I'm just trying to wrap my head around things 😅

PR #1190, opened 2022-12-28
Grw 1953/checkout discount applied states

KarnellSchultz:
## Describe your changes

- Update the /checkout page to show the applied discount states. 
- Create hooks to get the `total` and `crossOut` prices

Screenshot 2022-12-28 at 15 19 01
Screenshot 2022-12-28 at 15 19 09

- The user would like to see what price they're expected to be paying on the checkout page.

[GRW-1953]: https://hedvig.atlassian.net/browse/GRW-1953?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/store/src/components/CartInventory/CartInventory.helpers.tsx
+++ b/apps/store/src/components/CartInventory/CartInventory.helpers.tsx
@@ -1,5 +1,6 @@
 import { useTranslation } from 'next-i18next'
 import { CampaignDiscount, CampaignDiscountType } from '@/services/apollo/generated'
+import { useShopSession } from '@/services/shopSession/ShopSessionContext'
 import { Money } from '@/utils/formatter'
 import { useFormatter } from '@/utils/useFormatter'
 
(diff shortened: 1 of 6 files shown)


apps/store/src/components/CartInventory/CartInventory.helpers.tsx
```diff
@@ -49,3 +50,32 @@ export const useGetDiscountDurationExplanation = () => {
...
   }
 }
+
+export const useGetTotal = () => {
+  const { shopSession } = useShopSession()
+  const hasDiscount = shopSession?.cart.redeemedCampaigns.length !== 0
+
+  if (!hasDiscount) return shopSession.cart.cost.net
+  // Only expecting one discount right now. Going forward we'd need to make this work for multi discounts.
+  switch (shopSession?.cart.redeemedCampaigns[0].discount.type) {
```
KarnellSchultz, line 60: Looking for input here.

Since `shopSession` can return `undefined` this function either returns `Money` or `undefined` which is causing me to check if `total` exists when the prop is passed down. 

Seems like there is a better way to do this, but after thinking about it I was not able to come up with anything good. What say you? 🙇‍♂️

apps/store/src/components/CartInventory/CostSummary.tsx
```diff
@@ -11,6 +11,8 @@ export const CostSummary = ({ total, crossOut, campaigns }: Props) => {
   const formatter = useFormatter()
   const { t } = useTranslation('cart')
 
+  if (!total) return null
```
robinandeer, line 14: Why would this ever happen? What's the reason to render `CostSummary` if we don't know the total cost? 😅

apps/store/src/components/CartInventory/CartInventory.helpers.tsx
robinandeer: I would pass in only the information necessary -- then it's up to the caller to make sure the we have a valid shop session :)

On both cart and checkout page we fetch shop session server side so we know it will be defined - so should be no problem, right?

apps/store/src/components/CartInventory/CostSummary.tsx
KarnellSchultz: Right! It won't happen. That's kinda how I knew there was an issue with what I was doing here.

KarnellSchultz:
@robinandeer Thanks for the tips. Made some changes that I'm more happy with. Let me know if you have any other suggestions 👍

alebedev (approved):
Looks good!

PR #1191, opened 2022-12-28
GRW-1983 / Fix / Prevent duplicate creation of PriceIntent

alebedev:
## Describe your changes

- Move PriceIntent handling up to its own provider
- Run data manipulation outside React, remove duplicate in-react implementation (same idea as ShopSessionContent)
- Make PurchaseForm safe to render twice which we need for responsive UI

Fix duplicate creation bug.  Simplify code, make it more reliable


apps/store/src/components/ProductPage/PriceIntentContext.tsx
```diff
@@ -0,0 +1,55 @@
...
+export const PriceIntentContextProvider = ({ children }: Props) => {
+  const contextValue = usePriceIntentContextValue()
+  return <PriceIntentContext.Provider value={contextValue}>{children}</PriceIntentContext.Provider>
+}
+
+const usePriceIntentContextValue = () => {
+  const { priceTemplate, productData } = useProductPageContext()
+  const apolloClient = useApolloClient()
+  const { shopSession, onReady } = useShopSession()
+  const [fetchQuery, queryResult] = usePriceIntentLazyQuery({
```
robinandeer, line 24: Curious why we don't use "skip" solution here? Is it because we first want to validate the price intent before loading it?

apps/store/src/components/ProductPage/PriceIntentContext.tsx
```diff
@@ -0,0 +1,55 @@
...
+  const productName = productData.name
+  const priceIntentService = useMemo(() => {
+    if (isBrowser() && shopSession) {
+      return priceIntentServiceInitClientSide({ apolloClient, shopSession })
+    }
+  }, [apolloClient, shopSession])
+
+  useEffect(() => {
+    return onReady(() => {
+      priceIntentService?.getOrCreate({ priceTemplate, productName }).then((priceIntent) => {
```
robinandeer, line 38: Shouldn't we return shop session as a param from `onReady`?

That way we can use it to setup `priceIntentService` in this code block and we don't have to use optional chaining. It feels a bit disconnected right now.

apps/store/src/services/priceIntent/PriceIntentService.ts
```diff
@@ -1,39 +1,63 @@
...
+          })
+        }
+      },
     })
+    if (!result.data?.priceIntentCreate) throw new Error('Could not create price intent')
 
-    this.persister.save(priceIntent.id, this.getPriceIntentKey(priceTemplate.name))
+    let priceIntent: PriceIntent = result.data.priceIntentCreate
 
     if (priceTemplate.initialData) {
```
robinandeer, line 52: Isn't this already implemented on the backend? We should be able to remove it :)

robinandeer (approved):
Good stuff -- just added some improvement ideas

apps/store/src/components/ProductPage/PriceIntentContext.tsx
alebedev: Since priceIntent is only used in client side rendering, we know we need to wait for session.  Lazy query means we don't need to keep extra React state just to trigger skip

apps/store/src/components/ProductPage/PriceIntentContext.tsx
alebedev: Good one.  I think I'll just refactor priceIntentService to not require shopSession in constructor

apps/store/src/services/priceIntent/PriceIntentService.ts
alebedev: Not yet, created JIRA task to make sure we don't forget it

PR #1200, opened 2022-12-29
GRW-1981 / Feature / Report tracking events to datadog

alebedev:
## Describe your changes

Tracking service reports events to Datadog as well

This is how offer_created log looks like in Datadog:
Screenshot 2022-12-29 at 11 29 29

More debugging info for Analytics, we may need it in the future

### DIFF
--- a/apps/store/src/services/Tracking/Tracking.ts
+++ b/apps/store/src/services/Tracking/Tracking.ts
@@ -1,5 +1,8 @@
+import { datadogLogs } from '@datadog/browser-logs'
+import { ProductData } from '@/components/ProductPage/ProductPage.types'
 import { ProductOfferFragment } from '@/services/apollo/generated'
 import { AppTrackingContext, pushToGTMDataLayer, setGtmContext } from '@/services/gtm'
+import { PriceIntent } from '@/services/priceIntent/priceIntent.types'
 import { isBrowser } from '@/utils/env'
 import { newSiteAbTest } from '../../newSiteAbTest'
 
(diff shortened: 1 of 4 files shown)


apps/store/src/services/Tracking/Tracking.ts
```diff
@@ -11,22 +14,44 @@ export enum TrackingEvent {
   PageView = 'virtual_page_view',
 }
 
+datadogLogs.createLogger('tracking')
```
robinandeer, line 17: nit: use a constant to keep track the logger name

or does `createLogger` return a logger instance we can use somewhere else?

apps/store/src/services/Tracking/Tracking.ts
```diff
@@ -11,22 +14,44 @@ export enum TrackingEvent {
...
   constructor(public context: TrackingContext = {}) {}
 
+  private logger = datadogLogs.getLogger('tracking')!
+
   public setContext = (key: string, value: unknown) => {
     if (this.context[key] !== value) {
       console.debug(`tracking context ${key}:`, value)
     }
-    this.context[key] = value
+    if (value != null) {
```
robinandeer, line 29: nit: just used to strict comparison -- shouldn't we always do that (!==)?

apps/store/src/services/logger/client.ts
```diff
@@ -17,7 +17,11 @@ export const initDatadog = () => {
       datadogLogsConfig.beforeSend = (event) => {
         // Must exclude console origin to avoid endless loop.  Feel free to experiment with other values
         if (event.origin === 'logger') {
-          console[event.status](event.message)
+          if (event.logger?.name == 'tracking') {
```
robinandeer, line 20: nit: use the constant to keep track of this string value
nit #2: use strict comparison

apps/store/src/services/Tracking/Tracking.ts
```diff
@@ -11,22 +14,44 @@ export enum TrackingEvent {
...
   }
 
+  public setProductContext = (product: ProductData) => {
+    this.setContext('has_home', product.name.includes('_APARTMENT'))
+    this.setContext('has_house', product.name.includes('_HOUSE'))
+    this.setContext('has_card', product.name.includes('_CAR'))
+    this.setContext('has_accident', product.name.includes('_ACCIDENT'))
+    this.setContext('has_travel', false)
+    this.setContext('is_student', product.name.includes('STUDENT'))
+  }
```
robinandeer, line 48: This is starting to look very promise specific -- I'm sure we can do better ;)

But feel free to tackle it later

robinandeer (approved):
Great stuff -- left a bunch of comments, happy to discuss

apps/store/src/services/Tracking/Tracking.ts
alebedev: The only other usage is less than 10 lines below, where we initialize `Tracking.logger`

I think adding constant does not improve readability here

apps/store/src/services/Tracking/Tracking.ts
alebedev: This is intentional, called can pass null or undefined

apps/store/src/services/logger/client.ts
alebedev: Ok, this is another usage, I'm creating a constant

apps/store/src/services/Tracking/Tracking.ts
alebedev: Sonny asked for it to be backwards compatible with current site analytics.  I don't like it, but at least we're keeping this mapping in one place

PR #1201, opened 2022-12-29
GRW-1904 - feat: added ProductVariantSelector

guilhermespopolin:
## Describe your changes

* Adds `ProductVariantSelector`

## Screenshot

**Disclaimer**:  `ProductVariant.displayName` is currently not being populated by the API. Empty string is returned for each variant, so that's why the select is "empty". So as soon as we get that value, we'll be showing it.

https://user-images.githubusercontent.com/19200662/209939675-27f719f0-3aaa-4773-a6a0-0775b78cf401.mov

### DIFF
--- a/apps/store/src/components/InputSelect/InputSelect.tsx
+++ b/apps/store/src/components/InputSelect/InputSelect.tsx
@@ -21,7 +21,7 @@ const StyledSelect = styled(motion.select)(({ theme }) => ({
   fontSize: theme.fontSizes[5],
   width: '100%',
   borderRadius: theme.radius.sm,
-  padding: `${theme.space[2]} ${theme.space[4]}`,
+  padding: `${theme.space[2]} ${theme.space[7]} ${theme.space[2]} ${theme.space[4]}`,
 
   '&:hover': {
     cursor: 'pointer',
@@ -45,6 +45,7 @@ type InputSelectProps = InputBaseProps & {
   required?: boolean
   placeholder?: string
   autoFocus?: boolean
+  className?: string
 }
 
 export const InputSelect = ({
@@ -55,6 +56,7 @@ export const InputSelect = ({
   defaultValue,
   placeholder,
   label,
+  className,
   ...rest
 }: InputSelectProps) => {
   const { highlight, animationProps } = useHighlightAnimation()
@@ -69,7 +71,7 @@ export const InputSelect = ({
   return (
     <InputBase {...rest}>
       {() => (
-        <Wrapper>
+        <Wrapper className={className}>
           <StyledSelect
             name={name}
             onChange={handleChange}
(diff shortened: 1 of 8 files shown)


guilhermespopolin:
* main
  * **PR #1201** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1201" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1201?utm_source=stack-comment).

apps/store/src/components/ProductVariantSelector/ProductVariantSelector.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+  const handleChange: ChangeEventHandler<HTMLSelectElement> = (event) => {
+    const newSelectedVariant = productData.variants.find(
+      (variant) => variant.typeOfContract === event.target.value,
+    )
+
+    if (newSelectedVariant) {
+      selectedVariantUpdate(newSelectedVariant)
+    } else {
+      // TODO: do tracking here
```
guilhermespopolin, line 29: [Help wanted]
I think we might want to track/notify if this 'else' branch gets reached. But I'm not sure what exactly I should call here. Any ideas?

alebedev: Not really needed, we want to track when priced offer is selected

apps/store/src/components/InputSelect/InputSelect.tsx
```diff
@@ -21,7 +21,7 @@ const StyledSelect = styled(motion.select)(({ theme }) => ({
   fontSize: theme.fontSizes[5],
   width: '100%',
   borderRadius: theme.radius.sm,
-  padding: `${theme.space[2]} ${theme.space[4]}`,
+  padding: `${theme.space[2]} ${theme.space[7]} ${theme.space[2]} ${theme.space[4]}`,
```
alebedev, line 24: Won't this affect mileage select and other use cases in purchase forms?

alebedev (approved):
Nice!   Can you link `ProductDocumentsBlock` to selected variant?  Currently it shows documents for all variants

apps/store/src/components/InputSelect/InputSelect.tsx
guilhermespopolin: It shouldn't but I'm gonna check.
The reasoning behind it is that without creating enough room for the absolut positioned icon with padding we might end up in the situation where select content gets displayed under the icon.

image

guilhermespopolin:
> Nice! Can you link `ProductDocumentsBlock` to selected variant? Currently it shows documents for all variants

Will do!

apps/store/src/components/InputSelect/InputSelect.tsx
alebedev: Makes sense!

PR #1205, opened 2022-12-29
GRW-1988 / Feature / Report Ecommerce events to Google Analytics

alebedev:
## Describe your changes

Report view_item and add_to_cart events via GTM

Would allow us to get better data in Analytics.  And by using recommened event structure we reduce amount of manual configuration required


apps/store/src/components/ProductPage/PurchaseForm/OfferPresenter.tsx
```diff
@@ -48,6 +49,14 @@ export const OfferPresenter = (props: Props) => {
     setSelectedOffer(offer)
   }
 
+  // TODO: Deal with duplicate tracking from purchase form responisve layout
+  //   Perhaps only track when offer becomes visible via intersectionObserver?
+  const tracking = useTracking()
```
alebedev, line 54: @robinandeer We need to track initial view in addition to change.  I want to experiment with visibility detection for this.  The right moment to track view event is when user have seen something, not when we've set a state to render it

apps/store/src/components/ProductPage/PurchaseForm/useHandleSubmitAddToCart.ts
```diff
@@ -11,24 +13,50 @@ type Params = {
...
       },
     })
   }
 
   return [handleSubmit, loading] as const
 }
+
+type CartEntryAddOptions = Parameters<typeof useCartEntryAddMutation>[0]
+
+const useCartEntryAdd = (mutationOptions: CartEntryAddOptions = {}) => {
```
alebedev, line 37: @robinandeer separated mutation and success handler into their own hook as you suggested

guilhermespopolin (approved):

apps/store/src/services/gtm.tsx
```diff
@@ -39,17 +39,24 @@ type GTMPageData = {
...
   eventData?: Record<string, string>
   pageData?: GTMPageData
+  ecommerce?: GTMEcommerceData
 }
 
 // Needed in case event is sent before GTM is loaded, see https://github.com/HedvigInsurance/racoon/commit/38dbb73d552a590f652bbbe537d4d8ed4b0399f8
 export const pushToGTMDataLayer = (obj: DataLayerObject) => {
   if (!window.dataLayer) window.dataLayer = []
+  // Clear the previous ecommerce object
+  if (typeof obj.ecommerce !== 'undefined') {
```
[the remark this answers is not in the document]
alebedev: This is recommendation from Google.  I guess if you don't do it, your values may be merged with previous ones

hasOwnProperty only makes sense when property name is dynamic, to avoid things that come from prototypes

apps/store/src/components/ProductPage/PurchaseForm/OfferPresenter.tsx
robinandeer: that's true, although we probably don't want to send multiple events if you scroll up and down the page

According to react docs it's okey to use `useEffect` for tracking since double render only happens in dev and analytics should be turned off in dev anyways...

robinandeer (approved):
Looks good to me!

PR #1208, opened 2022-12-29
GRW-1904 - fix(ProductDocuments): only show documents of the selected product variant

guilhermespopolin:
## Describe your changes

* Update `ProductDocumentsBlock` so it displays only the documents for the selected variant

### DIFF
--- a/apps/store/src/blocks/ProductDocumentsBlock.tsx
+++ b/apps/store/src/blocks/ProductDocumentsBlock.tsx
@@ -7,13 +7,17 @@ type Props = SbBaseBlockProps<{
   heading: string
 }>
 
-// TODO: Also use variant context and show docs from selected variant only
 export const ProductDocumentsBlock = ({ blok }: Props) => {
-  const { productData } = useProductPageContext()
+  const { productData, selectedVariant } = useProductPageContext()
   const { heading } = blok
-  return (
-    <ProductDocuments heading={heading} productData={productData} {...storyblokEditable(blok)} />
-  )
+
+  const docs = selectedVariant?.documents ?? productData.variants[0].documents
+  if (docs.length < 1) {
+    console.warn('No documents to show; hiding productDocuments')
+    return null
+  }
+
+  return <ProductDocuments heading={heading} docs={docs} {...storyblokEditable(blok)} />
 }
 
 ProductDocumentsBlock.blockName = 'productDocuments'
--- a/apps/store/src/components/ProductPage/ProductDocuments.tsx
+++ b/apps/store/src/components/ProductPage/ProductDocuments.tsx
@@ -1,28 +1,21 @@
 import styled from '@emotion/styled'
 import { Card, CardContent } from 'ui/src/components/Card/Card'
 import { getColor, HeadingLabel, Space, Text } from 'ui'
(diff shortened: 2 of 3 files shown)


guilhermespopolin:
* main
  * **PR #1201** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1201" target="_blank">Graphite</a> 
    * **PR #1208** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1208" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1208?utm_source=stack-comment).

alebedev (approved):
Nice, thanks!

apps/store/src/blocks/ProductDocumentsBlock.tsx
```diff
@@ -7,13 +7,17 @@ type Props = SbBaseBlockProps<{
...
-// TODO: Also use variant context and show docs from selected variant only
 export const ProductDocumentsBlock = ({ blok }: Props) => {
-  const { productData } = useProductPageContext()
+  const { productData, selectedVariant } = useProductPageContext()
   const { heading } = blok
-  return (
-    <ProductDocuments heading={heading} productData={productData} {...storyblokEditable(blok)} />
-  )
+
+  const docs = selectedVariant?.documents ?? productData.variants[0].documents
```
guilhermespopolin, line 14: This line might be interesting. I believe we want to display the documents for the first variant, which appears to be default selected one. However, I'm not sure if this definition of "default selected variant" is mature enough by setting the initial selected variant directly in [`ProductPageContext`](https://github.com/HedvigInsurance/racoon/blob/main/apps/store/src/components/ProductPage/ProductPageContext.tsx#L21). 

I've saw that [`PerilsBlock`](https://github.com/HedvigInsurance/racoon/blob/main/apps/store/src/blocks/PerilsBlock.tsx) also does the same in a sense of making sure to display something when the component first renders. But in a different way. Since the one I'm using in `ProductDocumentsBlock` is simpler, it makes me thing if I'm missing something 🤔 

Any inputs @robinandeer ?

apps/store/src/components/ProductPage/ProductDocuments.tsx
```diff
@@ -1,28 +1,21 @@
...
     <ProductDocumentsWrapper>
       <Space y={1}>
         <HeadingLabel>{heading}</HeadingLabel>
-        {productData.variants.flatMap((variant) =>
-          variant.documents.map((doc, i) => (
-            <ProductDocument key={`${variant.typeOfContract}_${i}`} doc={doc} />
-          )),
-        )}
+        {docs.map((doc, index) => (
+          <ProductDocument key={index} doc={doc} />
```
guilhermespopolin, line 17: I'm using `index` as the key here. I know that may be problematic, but the usage of `index` as key ["is not recommended **if the order of the items may change**"](https://reactjs.org/docs/lists-and-keys.html#:~:text=We%20don%E2%80%99t%20recommend%20using%20indexes%20for%20keys%20if%20the%20order%20of%20items%20may%20change). So I believe it's safe here.

PR #1220, opened 2023-01-02
GRW-2009 / Feature / Track cart and checkout analytics events

alebedev:
## Describe your changes

Track view_cart, add_to_cart, remove_from_cart, begin_checkout

Report ecommerce events to Google Analytics

### DIFF
--- a/apps/store/src/components/CartInventory/CartEntryItem.tsx
+++ b/apps/store/src/components/CartInventory/CartEntryItem.tsx
@@ -3,10 +3,10 @@ import { useTranslation } from 'next-i18next'
 import Link, { LinkProps } from 'next/link'
 import { FormEvent } from 'react'
 import { Button, Dialog, Space, Text } from 'ui'
+import { useRemoveCartEntry } from '@/components/CartInventory/useRemoveCartEntry'
 import * as FullscreenDialog from '@/components/FullscreenDialog/FullscreenDialog'
 import { Pillow } from '@/components/Pillow/Pillow'
 import { SpaceFlex } from '@/components/SpaceFlex/SpaceFlex'
-import { useCartEntryRemoveMutation } from '@/services/apollo/generated'
 import { useFormatter } from '@/utils/useFormatter'
 import { CartEntry } from './CartInventory.types'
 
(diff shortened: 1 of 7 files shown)


apps/store/src/components/CartInventory/CartEntryItem.tsx
```diff
@@ -3,10 +3,10 @@ import { useTranslation } from 'next-i18next'
 import Link, { LinkProps } from 'next/link'
 import { FormEvent } from 'react'
 import { Button, Dialog, Space, Text } from 'ui'
+import { useRemoveCartEntry } from '@/components/CartInventory/useRemoveCartEntry'
```
robinandeer, line 6: nit: this could be a relative import since it's from the same component

apps/store/src/components/CartInventory/useRemoveCartEntry.ts
```diff
@@ -0,0 +1,35 @@
...
+import { useShopSession } from '@/services/shopSession/ShopSessionContext'
+import { useTracking } from '@/services/Tracking/useTracking'
+
+export const useRemoveCartEntry = ({ cartId, offerId }: { cartId: string; offerId: string }) => {
+  const { shopSession } = useShopSession()
+
+  const tracking = useTracking()
+  const [runMutation, mutationResult] = useCartEntryRemoveMutation({
+    refetchQueries: 'active',
+    awaitRefetchQueries: true,
```
robinandeer, line 13: 🤔 do we need these?

apps/store/src/components/CartPage/CartPage.tsx
```diff
@@ -1,22 +1,36 @@
...
 
+  const tracking = useTracking()
+  useEffect(() => {
+    const { cart } = shopSession ?? {}
+    if (cart) {
+      tracking.reportViewCart(cart)
+    } else {
+      datadogLogs.logger.error('No cart data on cartPage')
+    }
+  }, [tracking, shopSession])
```
robinandeer, line 32: won't this track a new event every time the shop session changes?

we also track page views -- wouldn't that cover the same thing?

robinandeer (approved):
Looks good! Just had a question about "view cart event" 😉

apps/store/src/components/CartInventory/useRemoveCartEntry.ts
alebedev: Valid point, will investigate in separate PR

apps/store/src/components/CartPage/CartPage.tsx
alebedev: Payload is important, page view for opening cart won't have value or ecommerce field with items list.  For this reason we may sometimes track multiple analytics events for same thing. For example, virtual_page_view and view_cart here, purchase and signed_customer on signing success

alebedev: Fixed according to Slack discussion

PR #1223, opened 2023-01-02
GRW-2006 / Feature / Track view_item when offer becomes visible

alebedev:
## Describe your changes

Use intersection observer util from framer-motion to track view_item event when selectedOffer first becomes visible

- Solves double event from responsive rendering of OfferPresenter
- Solves race condition with progress bar and event timing
- Useful example of tracking something being visible, we'll likely need to track more such things in the future

### DIFF
--- a/apps/store/src/components/ProductPage/PurchaseForm/OfferPresenter.tsx
+++ b/apps/store/src/components/ProductPage/PurchaseForm/OfferPresenter.tsx
@@ -1,7 +1,8 @@
 import { datadogLogs } from '@datadog/browser-logs'
 import styled from '@emotion/styled'
+import { useInView } from 'framer-motion'
 import { useTranslation } from 'next-i18next'
-import { RefObject, useEffect, useMemo, useState } from 'react'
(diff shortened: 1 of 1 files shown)


apps/store/src/components/ProductPage/PurchaseForm/OfferPresenter.tsx
```diff
@@ -49,13 +50,14 @@ export const OfferPresenter = (props: Props) => {
...
-  //   Perhaps only track when offer becomes visible via intersectionObserver?
+  const offerRef = useRef(null)
   const tracking = useTracking()
-  useEffect(
-    () => selectedOffer && tracking.reportViewItem(selectedOffer),
-    [selectedOffer, tracking],
-  )
+  const isInView = useInView(offerRef, { once: true })
+  useEffect(() => {
+    if (selectedOffer && isInView) {
```
robinandeer, line 57: `selectedOffer` is always defined -- why do we need to check here?

robinandeer (approved):
Cool, doesn't look to complicated! And guess this is a proper use case for `useEffect`

PR #1233, opened 2023-01-03
Show name inputs in Checkout if missing

robinandeer:
## Describe your changes

Show first name and last name input in Checkout if they are missing.

First and last name are required to sign insurance. If the lookup based on SSN fails it makes sense to allow users to add their own name instead.

### DIFF
--- a/apps/store/src/components/CheckoutPage/CheckoutPage.tsx
+++ b/apps/store/src/components/CheckoutPage/CheckoutPage.tsx
@@ -24,7 +24,7 @@ import { CheckoutPageProps } from './CheckoutPage.types'
 import { useHandleSubmitCheckout } from './useHandleSubmitCheckout'
 
 const CheckoutPage = (props: CheckoutPageProps) => {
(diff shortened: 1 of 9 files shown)


apps/store/src/components/CheckoutPage/CheckoutPage.tsx
```diff
@@ -99,6 +99,24 @@ const CheckoutPage = (props: CheckoutPageProps) => {
                 readOnly
                 disabled
               />
+              {collectName && (
```
alebedev, line 102: Would it be more explicit if we'd always shown these fields?  Perhaps as readonly if we want to prevent editing

alebedev (approved):
Looks good, one question

apps/store/src/components/CheckoutPage/CheckoutPage.tsx
robinandeer: Yeah I was also thinking that -- asked in Figma :)

To start I just wanted to stick as close to the design as possible.

PR #1234, opened 2023-01-03
Fix / Optimize cart refresh on item remove

alebedev:
## Describe your changes

Don't refetch queries after removing cart entry - we get updated cart in response and it's automatically used to update Apollo cache

- Faster UI, one less network roundtrip
- Less code


robinandeer (approved):
[cathack](https://app.graphite.dev/macros/HedvigInsurance)

guilhermespopolin (approved):

PR #1245, opened 2023-01-04
GRW-2017 / Feature / Sort translations alphabetically

alebedev:
## Describe your changes

Subj

Allows external tools like source scanners to keep translations ordered.  Relying on localize metadata only works for localize cli


robinandeer (approved):
[approvedarthvader](https://app.graphite.dev/macros/HedvigInsurance)

guilhermespopolin (approved):
Trivia
Given their meaning, doesn't also bother you the fact that "after" comes alphabetically before of "before"? 🤔

PR #1246, opened 2023-01-04
GRW-2018 / Feature / Rename 'default' locale to 'en'

alebedev:
## Describe your changes

Rename magic 'default' locale to 'en'

Explicit is better than implicit, simplify tools integration


robinandeer (commented):
Interesting

We don't support any default language/locale. I think that's why it made sense to not use value that matches an actual language. In this context, I think not using "en" is the more explicit option, what to you think about that?

"default" is also a bad name perhaps - it's just what Next.js docs recommend.

Does it make a difference that we have a country selector page?

guilhermespopolin (approved):
👌🏻

PR #1250, opened 2023-01-04
GRW-1598 / Feature / Scan for translation keys

alebedev:
## Describe your changes

Add scan-translations script to detect new and unused translations in apps/store

Annotate purchase form configs to correctly detect used keys and namespaces

Replace key concatenation with switch statements

Better developer experience

Ability to detect unused translation keys to help with append-only dictionaries


apps/store/public/locales/en/cart.json
```diff
@@ -8,9 +8,15 @@
   "CHECKOUT_BUTTON": "Checkout",
   "CHECKOUT_PRICE_TOTAL": "Total",
   "DISCOUNT_DURATION_EXPLANATION": "For {{count}} month, then",
+  "DISCOUNT_DURATION_EXPLANATION_one": "",
```
robinandeer, line 11: Hmm I wonder if we should avoid "count" variable unless we explicitly want to pluralize translations...?

apps/store/src/services/PriceCalculator/data/SE_ACCIDENT.ts
```diff
@@ -6,16 +6,19 @@ import {
   streetAddressField,
   yourFamilySection,
 } from '@/services/PriceCalculator/formFragments'
+import { setI18nNamespace, tKey } from '@/utils/i18n'
 import { Template } from '../PriceCalculator.types'
 
+setI18nNamespace('purchase-form')
```
robinandeer, line 12: aha does this work outside react? I guess it's safe it we reload the page every time we switch locale, right?

I imagine we move this to the server in the future -- but then I guess server would just give us translated content right away

alebedev: This function is no-op for scanner.  We need to make sure it's in sync with the call site where we translate the keys

apps/store/public/locales/en/cart.json
alebedev: We probably should - created follow-up task to investigate

PR #1260, opened 2023-01-05
Separate out `EntryRemoveDialog` component

robinandeer:
## Describe your changes

Separate out a `RemoveEntryDialog` component from `CartInventory`.

## Justify why they are needed

Avoid nesting Dialog root components.

## Checklist before requesting a review

- [ ] I have performed a self-review of my code

PR #1263, opened 2023-01-05
GRW-2016/Desktop layout - Cart Entry

KarnellSchultz:
## Describe your changes

Created a desktop layout of cart entry items. This component is seen on the /cart and /checkout pages.

Attempted to do this in several ways. Thought this was the least complicated, although I don't _love_ how the show/hide of things like the "title" (`<MobileTitle />` & `<DesktopTitle />`) are done with css. 

checkout-entries
cart-entries

Matching the designs in Figma.
## Figma
Screenshot 2023-01-05 at 15 07 31

[GRW-2016]: https://hedvig.atlassian.net/browse/GRW-2016?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/store/src/components/CartInventory/CartEntryItem.tsx
+++ b/apps/store/src/components/CartInventory/CartEntryItem.tsx
@@ -1,6 +1,9 @@
 import styled from '@emotion/styled'
 import { useTranslation } from 'next-i18next'
-import { Button, Dialog, Space, Text } from 'ui'
+import Link, { LinkProps } from 'next/link'
+import { FormEvent } from 'react'
+import { Button, Dialog, mq, Space, Text } from 'ui'
+import * as FullscreenDialog from '@/components/FullscreenDialog/FullscreenDialog'
 import { Pillow } from '@/components/Pillow/Pillow'
 import { SpaceFlex } from '@/components/SpaceFlex/SpaceFlex'
 import { useFormatter } from '@/utils/useFormatter'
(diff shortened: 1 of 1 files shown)


apps/store/src/components/CartInventory/CartEntryItem.tsx
```diff
@@ -1,58 +1,135 @@
...
+    display: 'grid',
+    gridTemplateColumns: '3fr 2fr 2fr',
+    gap: theme.space.sm,
+  },
+}))
+
+const PillowWrapper = styled.div({
+  display: 'flex',
+  alignItems: 'flex-start',
+  [mq.lg]: {
```
alebedev, line 112: Doesn't this block duplicate styles above?

alebedev (commented):
This probably needs a merge with PR separating EntryRemoveDialog (@source:lambda  START)

apps/store/src/components/CartInventory/CartEntryItem.tsx
```diff
@@ -1,58 +1,124 @@
...
+  gap: theme.space.sm,
+  [mq.lg]: {
+    gridTemplateColumns: '3fr 2fr 2fr',
+  },
+}))
+
+const DesktopTitle = styled(Text)({
+  display: 'none',
+  [mq.lg]: {
+    display: 'block',
```
guilhermespopolin, line 110:
```suggestion
    display: 'revert',
```
This is just a small suggestion. I've seen this "pattern" a lot recently. The advantage is that one doesn't need to worry about checking what would be the default display value of an element.

guilhermespopolin (commented):
I think that's fine the way you've implemented show/hide titles with CSS. However I agree with Aleks that we probably want to reuse [`RemoveEntryDialog`](https://github.com/HedvigInsurance/racoon/pull/1260/files#diff-0b9d91d939c55dc07c7f4a6b04544f5c0a63eff33571b551273131208bcc9035) here.

apps/store/src/components/CartInventory/CartEntryItem.tsx
```diff
@@ -1,58 +1,124 @@
...
-            </Dialog.Trigger>
-          </DetailsSheetDialog>
-          <RemoveEntryDialog cartId={cartId} {...cartEntry}>
+        <Space y={1}>
+          <div>
+            <MobileTitle size="md">{title}</MobileTitle>
+            <Text size="md" color="textSecondary">
+              {/* @TODO: display "automatically switches" if cancellation is requested" */}
+              {startDate
+                ? t('CART_ENTRY_DATE_LABEL', { date: formatter.fromNow(startDate), ns: 'cart' })
```
alebedev, line 45: Nit: don't need explicit ns if it matches argument of `useTranslation()`

KarnellSchultz:
@alebedev @guilhermespopolin  Thanks for the comments. Looks like I mixed up the merge conflicts when working on this PR and left out the stuff from Robins [PR](https://github.com/HedvigInsurance/racoon/pull/1260/files#).

Should be sorted now. I'm guessing some of this stuff is going to change as we get more information on breakpoints from designers.

alebedev (approved):
Looks good!

PR #1267, opened 2023-01-05
Refactor Cart + Campaign Code stuff

robinandeer:
## Describe your changes

- Use `Text` component in-favor manually styling text
- Make sure add campaign button doesn't shrink when in "loading" state
- Add focus styles to remove campaign button
- Use a form to start checkout

- Without a form, there's no way for assitive devices to know what actions can be performed on a page.

### DIFF
--- a/apps/store/src/components/CartPage/CartPage.tsx
+++ b/apps/store/src/components/CartPage/CartPage.tsx
@@ -3,7 +3,7 @@ import styled from '@emotion/styled'
 import { useTranslation } from 'next-i18next'
 import Link from 'next/link'
 import { useRouter } from 'next/router'
-import { ReactNode, useEffect } from 'react'
+import { FormEventHandler, ReactNode, useEffect } from 'react'
 import { Button, Heading, mq, Space, Text } from 'ui'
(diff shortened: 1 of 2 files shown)


apps/store/src/components/CartPage/CartPage.tsx
```diff
@@ -69,9 +74,11 @@ export const CartPage = (props: CartPageProps) => {
         <CampaignCodeList cartId={cartId} campaigns={campaigns} />
         <HorizontalLine />
         <CostSummary {...cost} campaigns={campaigns} />
-        <Button onClick={() => startCheckout()} disabled={loadingStartCheckout}>
-          {t('CHECKOUT_BUTTON')}
-        </Button>
+        <form onSubmit={handleSubmit}>
+          <Button disabled={loading} loading={loading}>
```
alebedev, line 78: Nit: do we need to pass disabled when it's always equal to loading?

robinandeer: Yeah I've been thinking about that -- perhaps we can assume that when loading you want the button to be disabled...

PR #1268, opened 2023-01-05
GRW-1985 / Feat / Ensure we do full page reload on country change

alebedev:
## Describe your changes

Add hook to ensure we reload a page if country changes during client side navigation

Make country selector controls and page use full reload when changing country

Ensure our assumptions about never changing country during client lifetime are satisfied, prevent bugs

This gives us simpler app model at expense of small increase in user wait time when country changes (should be rare)

### DIFF
--- a/apps/store/src/blocks/FooterBlock.tsx
+++ b/apps/store/src/blocks/FooterBlock.tsx
@@ -20,9 +20,10 @@ import { LocaleField, LOCALE_COOKIE_MAX_AGE, LOCALE_COOKIE_KEY } from '@/utils/l
 import {
   getLocaleOrFallback,
   toRoutingLocale,
+  translateCountryName,
   translateLanguageName,
 } from '@/utils/l10n/localeUtils'
-import { IsoLocale } from '@/utils/l10n/types'
+import { CountryLabel, IsoLocale } from '@/utils/l10n/types'
 import { useCurrentCountry } from '@/utils/l10n/useCurrentCountry'
 import { useCurrentLocale } from '@/utils/l10n/useCurrentLocale'
 
@@ -76,7 +77,7 @@ export const FooterBlock = ({ blok }: FooterBlockProps) => {
   const apolloClient = useApolloClient()
 
   const countryList = Object.keys(countries).map((country) => ({
-    name: t(`COUNTRY_LABEL_${country}`, { defaultValue: country }),
+    name: translateCountryName(country as CountryLabel, t),
     value: country,
   }))
 
--- a/apps/store/src/blocks/FooterBlock.tsx
+++ b/apps/store/src/blocks/FooterBlock.tsx
@@ -1,6 +1,8 @@
+import { useApolloClient } from '@apollo/client'
 import styled from '@emotion/styled'
 import { storyblokEditable } from '@storyblok/react'
 import { useTranslation } from 'next-i18next'
+import { addLocale } from 'next/dist/client/add-locale'
 import Link from 'next/link'
 import { useRouter } from 'next/router'
 import { ChangeEvent, useRef } from 'react'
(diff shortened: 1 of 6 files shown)


alebedev:
* main
  * **PR #1268** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1268" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1268?utm_source=stack-comment).

apps/store/src/utils/useReloadOnCountryChange.ts
```diff
@@ -0,0 +1,16 @@
+import { datadogLogs } from '@datadog/browser-logs'
+import { useEffect, useRef } from 'react'
+import { useCurrentCountry } from '@/utils/l10n/useCurrentCountry'
+
+export const useReloadOnCountryChange = () => {
+  const { countryCode } = useCurrentCountry()
+  const countryRef = useRef(countryCode)
+  useEffect(() => {
```
robinandeer, line 8: Hehe I feel like a broken record but I was inspired by David: https://www.youtube.com/watch?v=HPoC-k7Rxwo - goodbye use effect

Why not just subscribe to some router event here and perform the check? That is guaranteed to fire exactly when we care about it :)

apps/store/src/components/CountrySelectorPage/CountrySelectorPage.tsx
```diff
@@ -25,16 +23,13 @@ export const CountrySelectorPage = (props: StoryblokPageProps) => {
...
                 href={`/${toRoutingLocale(countryData.defaultLocale)}`}
-                passHref
-                legacyBehavior
+                onClick={() => onHandleClick(countryData.defaultLocale)}
               >
-                <Button onClick={() => onHandleClick(countryData.defaultLocale)}>
-                  {t(`COUNTRY_LABEL_${country}`, { defaultValue: `MISSING ${country}` })}
-                </Button>
-              </Link>
+                {t(`COUNTRY_LABEL_${country}`, { defaultValue: `MISSING ${country}` })}
```
robinandeer, line 31: nit: refactor so not use dynamic translation key

apps/store/src/blocks/FooterBlock.tsx
```diff
@@ -90,8 +94,17 @@ export const FooterBlock = ({ blok }: FooterBlockProps) => {
...
-    router.push(router.asPath, undefined, { locale: toRoutingLocale(locale) })
+    const nextLocale = toRoutingLocale(locale)
+    cookiePersister.save(nextLocale, undefined, { maxAge: LOCALE_COOKIE_MAX_AGE })
+    const nextCountry = getCountryByLocale(nextLocale)
+    if (nextCountry === currentCountry) {
+      router.push(router.asPath, undefined, { locale: nextLocale })
+    } else {
+      // Country change should be full app reload to maintain our programming assumptions
+      // We may clean any previous shop session while we're at it
+      setupShopSessionServiceClientSide(apolloClient).reset()
```
robinandeer, line 105: We take care of this when we validate shop session, no? I guess it might still be nice to do clean up here as well

apps/store/src/components/CountrySelectorPage/CountrySelectorPage.tsx
```diff
@@ -1,15 +1,13 @@
...
 import { Space, Heading, Button } from 'ui'
 import { CookiePersister } from '@/services/persister/CookiePersister'
-import { StoryblokPageProps } from '@/services/storyblok/storyblok'
 import { countries } from '@/utils/l10n/countries'
 import { LOCALE_COOKIE_MAX_AGE, LOCALE_COOKIE_KEY } from '@/utils/l10n/locales'
 import { toRoutingLocale } from '@/utils/l10n/localeUtils'
 import { IsoLocale } from '@/utils/l10n/types'
 
-export const CountrySelectorPage = (props: StoryblokPageProps) => {
+export const CountrySelectorPage = (props: { className?: string }) => {
```
robinandeer, line 10: I'm just confused; where does this class name come from? What do we use it for?

robinandeer (approved):
Good stuff, the more I understand, the more I like the approach -- I just question the use effect hook -- no surprises there ;)

apps/store/src/components/CountrySelectorPage/CountrySelectorPage.tsx
alebedev: This comes from App component - `className={contentFontClassName}`

Usually layout component takes this prop, but here page is its own layout

The reason I changed this place was to prevent passing other props like `_i18nInstance` to DOM

apps/store/src/blocks/FooterBlock.tsx
alebedev: I'm saving one network roundtrip for pages that don't use server side rendering

PR #1291, opened 2023-01-09
GRW-1907 -  feat: added BannerBlock

guilhermespopolin:
## Describe your changes

* Adds a `BannerBlock`

Can be tested on `en-se/insurances/home-insurance/rental` (I've tried to publish the changes but for some reason I still couldn't see them on staging) 🤷🏻 

---
Screenshot 2023-01-09 at 18 51 18
Screenshot 2023-01-09 at 18 50 36

## Jira issue(s): [GRW-1907](https://hedvig.atlassian.net/browse/GRW-1907)

[GRW-1907]: https://hedvig.atlassian.net/browse/GRW-1907?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/store/src/blocks/BannerBlock.tsx
+++ b/apps/store/src/blocks/BannerBlock.tsx
@@ -1,0 +1,83 @@
+import styled from '@emotion/styled'
+import { storyblokEditable } from '@storyblok/react'
+import { default as NextImage } from 'next/image'
+import { mq } from 'ui'
+import { HeadingBlock, HeadingBlockProps, isHeadingBlock } from '@/blocks/HeadingBlock'
+import { TextBlock, TextBlockProps, isTextBlock } from '@/blocks/TextBlock'
+import { SbBaseBlockProps, StoryblokAsset } from '@/services/storyblok/storyblok'
+
+type ImageSize = {
+  aspectRatioLandscape?: '4 / 3' | '16 / 9'
+  aspectRatioPortrait?: '1 / 1' | '4 / 5'
+}
+
+type BannerBlockProps = SbBaseBlockProps<
+  {
+    image: StoryblokAsset
+    fullBleed?: boolean
+    body?: Array<HeadingBlockProps['blok'] | TextBlockProps['blok']>
+  } & ImageSize
+>
+
+export const BannerBlock = ({ blok }: BannerBlockProps) => {
+  const headingBlocks = blok.body?.filter(isHeadingBlock)
+  const textBlocks = blok.body?.filter(isTextBlock)
+
+  return (
+    <Wrapper
+      aspectRatioLandscape={blok.aspectRatioLandscape ?? '16 / 9'}
+      aspectRatioPortrait={blok.aspectRatioPortrait ?? '1 / 1'}
+      {...storyblokEditable(blok)}
+    >
+      <Image src={blok.image.filename} alt={blok.image.alt} fill={true} />
+      <BodyWrapper>
+        {headingBlocks?.map((nestedBlock) => (
+          <HeadingBlock key={nestedBlock._uid} blok={nestedBlock} />
+        ))}
+        {textBlocks?.map((nestedBlock) => (
+          <TextBlock key={nestedBlock._uid} blok={nestedBlock} />
+        ))}
+      </BodyWrapper>
+    </Wrapper>
+  )
+}
+BannerBlock.blockName = 'banner'
+
+const Wrapper = styled('div')<ImageSize>(
+  ({ theme, aspectRatioLandscape, aspectRatioPortrait }) => ({
+    position: 'relative',
+    marginInline: theme.space.xs,
+    borderRadius: theme.radius.md,
+    overflow: 'hidden',
+
+    [mq.md]: {
+      marginInline: theme.space.md,
+      borderRadius: theme.radius.lg,
+    },
+
+    ['@media (orientation: portrait)']: {
+      ...(aspectRatioPortrait && { aspectRatio: aspectRatioPortrait }),
+    },
+    ['@media (orientation: landscape)']: {
+      ...(aspectRatioLandscape && { aspectRatio: aspectRatioLandscape }),
+    },
+  }),
+)
+
+const Image = styled(NextImage)({
+  objectFit: 'cover',
+})
+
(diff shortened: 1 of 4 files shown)


apps/store/src/blocks/BannerBlock.tsx
```diff
@@ -0,0 +1,83 @@
...
+type ImageSize = {
+  aspectRatioLandscape?: '4 / 3' | '16 / 9'
+  aspectRatioPortrait?: '1 / 1' | '4 / 5'
+}
+
+type BannerBlockProps = SbBaseBlockProps<
+  {
+    image: StoryblokAsset
+    fullBleed?: boolean
+    body?: Array<HeadingBlockProps['blok'] | TextBlockProps['blok']>
```
guilhermespopolin, line 18: Other blocks like [`ImageTextBlock`](https://github.com/HedvigInsurance/racoon/blob/main/apps/store/src/blocks/ImageTextBlock.tsx#L13) make usage of `ExpectedBlockType` to define the props. One issue I've find tho is that the resulting type doesn't seem to be accurate:

`body?: ExpectedBlockType<HeadingBlockProps | TextBlockProps> --> HeadingBlockProps[] | TextBlockProps[] | undefined` which infers a block can get a list of `HeadingBlock`s **or** a list of `TextBlock`s; I believe we want `body` being a list that can contain both types as its items, which we can achieve with `body?: Array<HeadingBlockProps['blok'] | TextBlockProps['blok']>`.

apps/store/src/blocks/BannerBlock.tsx
```diff
@@ -0,0 +1,83 @@
...
+  } & ImageSize
+>
+
+export const BannerBlock = ({ blok }: BannerBlockProps) => {
+  const headingBlocks = blok.body?.filter(isHeadingBlock)
```
guilhermespopolin, line 23: Here, other blocks like [`HeroBlock`](https://github.com/HedvigInsurance/racoon/blob/main/apps/store/src/blocks/HeroBlock.tsx#L16) uses `filterByBlockType` in order to manipulate the block a block can be nested with. However that function only works properly (in terms of typing) if the list provided to it can assume one single type of value.

Maybe one thing to do here would be update `BannerBlock` so it get its nested blocks separated... 🤔

apps/store/src/blocks/BannerBlock.tsx
```diff
@@ -0,0 +1,83 @@
...
+      aspectRatioPortrait={blok.aspectRatioPortrait ?? '1 / 1'}
+      {...storyblokEditable(blok)}
+    >
+      <Image src={blok.image.filename} alt={blok.image.alt} fill={true} />
+      <BodyWrapper>
+        {headingBlocks?.map((nestedBlock) => (
+          <HeadingBlock key={nestedBlock._uid} blok={nestedBlock} />
+        ))}
+        {textBlocks?.map((nestedBlock) => (
+          <TextBlock key={nestedBlock._uid} blok={nestedBlock} />
```
guilhermespopolin, line 38: In terms of styling we can't have too much control with `TextBlock`; That's why final result is not exactly the same as the design. So, should we update `TextBlock` so I can remove more control like fontSize and color or should we use the `ui/Text` component here 🤔 ?

apps/store/src/blocks/BannerBlock.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+const Image = styled(NextImage)({
+  objectFit: 'cover',
+})
+
+const BodyWrapper = styled.div(({ theme }) => ({
+  position: 'absolute',
+  inset: 0,
+  display: 'grid',
+  gridTemplateRows: 'auto 1fr auto',
```
guilhermespopolin, line 78: Was implemented with `grid` so the removal of one of the items (title or description) would still make things keep on their respective places.

apps/store/src/blocks/BannerBlock.tsx
```diff
@@ -0,0 +1,102 @@
...
+    borderRadius: theme.radius.md,
+    overflow: 'hidden',
+
+    [mq.md]: {
+      marginInline: theme.space.md,
+      borderRadius: theme.radius.lg,
+    },
+
+    ['@media (orientation: portrait)']: {
+      ...(aspectRatioPortrait && { aspectRatio: aspectRatioPortrait }),
```
alebedev, line 62: You're always providing default value above, no need to check for it here

alebedev (approved):
Looks good

apps/store/src/blocks/BannerBlock.tsx
```diff
@@ -0,0 +1,102 @@
...
+      <Image src={blok.image.filename} alt={blok.image.alt} fill={true} />
+      {hasContent && (
+        <BodyWrapper>
+          {blok.title && (
+            <Title as="h1" align="center" color="textNegative" variant="serif.32">
+              {blok.title}
+            </Title>
+          )}
+          {blok.description && (
+            <Description color="textNegative" size={{ _: 'md', lg: 'lg' }} align="center">
```
guilhermespopolin, line 38: I've implemented this with `ui/Text` component instead of `TextBlock` because we can't controls things like font size with `TextBlock`. Should we update it instead so we can use it here?
@gustaveen

apps/store/src/blocks/BannerBlock.tsx
```diff
@@ -0,0 +1,102 @@
+import styled from '@emotion/styled'
+import { storyblokEditable } from '@storyblok/react'
+import { default as NextImage } from 'next/image'
+import { Heading, Text, mq } from 'ui'
+import { SbBaseBlockProps, StoryblokAsset } from '@/services/storyblok/storyblok'
+
+type ImageSize = {
+  aspectRatioLandscape?: '4 / 3' | '16 / 9'
+  aspectRatioPortrait?: '1 / 1' | '4 / 5'
+}
```
gustaveen, line 10: Let's check what ratios they want to work with later but this works great for starters 👍

gustaveen (approved):
Great job! 🎉
Let's see if we should add some overlay to the image to increase contrast of the text.

PR #1300, opened 2023-01-10
fix(csp): added google fonts and storyblok

guilhermespopolin:
## Describe your changes

* Added CSP rules for google fonts and storyblok

So their resources don't get block


robinandeer (approved):
great

PR #1301, opened 2023-01-10
Integrate API for data table in cart details sheet

robinandeer:
## Describe your changes

Integrate API in the details sheet (cart + checkout) - specifically the user form data.

Setup generic "data table" interface to display the data.

Update apollo config to exclude generated graphql file - otherwise we get conflicts.

The data in the table differs between products. Therefore we need some configutation so we can avoid referencing invididual products in the code.

I used a similar setup to the price calculator template. It's very naive and we can generalize it if we need to. However, it's likely that this will be replaced by some edit form in the future.


robinandeer:
* main
  * **PR #1301** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1301" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1301?utm_source=stack-comment).

apps/store/src/components/CartInventory/DataTable/DataTable.ts
```diff
@@ -0,0 +1,61 @@
...
+  key: 'livingSpace',
+  label: tKey('DATA_TABLE_LIVING_SPACE_LABEL'),
+} as const
+
+const HOUSEHOLD_SIZE_ROW = {
+  type: 'HOUSEHOLD_SIZE',
+  label: tKey('DATA_TABLE_HOUSEHOLD_SIZE_LABEL'),
+} as const
+
+const DATA_TABLE = {
```
alebedev, line 28: Maybe it should be located in purchase forms config?  If we add a new field, we probably want to update single field

apps/store/src/components/CartInventory/DetailsSheetDialog.tsx
```diff
@@ -76,3 +85,40 @@ export const DetailsSheetDialog = ({ children, pillow, title, cost, documents }:
...
+const Capitalize = styled.span({ textTransform: 'capitalize' })
+
+type AutomaticDataTableRowProps = { row: DataTableRow; data: CartEntry['data'] }
+
+const AutomaticDataTableRow = ({ row, data }: AutomaticDataTableRowProps) => {
+  const { t } = useTranslation('cart')
+
+  let value: string | number = ''
+
+  if (row.type === 'STRING') {
```
alebedev, line 98: Switch statement would be safer - we'll know if we missed any types

alebedev (approved):
Looks good

apps/store/src/components/CartInventory/DataTable/DataTable.ts
robinandeer: I discussed this a bit with @guilhermespopolin and I think we agreed that it would be nice to limit the scope of the form configs to just what they are now -- what do you think?

Also since this is a feature mostly for the launch -- then we will add edit functionality which means we can delete this code. That's easier if we keep it separate from the form configs.

PR #1306, opened 2023-01-11
api / session / create small refactor

KarnellSchultz:
## Describe your changes

- Renamed a few variables to clarify them. Reworked how the `promise` is handled. 
- Added a 3rd item to the cart when this endpoint is used. 

Make this code easier to change in the future. 

### DIFF
--- a/apps/store/src/pages/api/session/create.ts
+++ b/apps/store/src/pages/api/session/create.ts
@@ -18,7 +18,7 @@ import { ORIGIN_URL, PageLink } from '@/utils/PageLink'
 
 const LOGGER = logger.child({ module: 'api/session/create' })
 const TEST_SSN = '199808302393'
-const productNames = ['SE_APARTMENT_RENT', 'SE_ACCIDENT', 'SE_APARTMENT_BRF'] as const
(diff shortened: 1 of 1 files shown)


apps/store/src/pages/api/session/create.ts
```diff
@@ -18,6 +18,7 @@ import { ORIGIN_URL, PageLink } from '@/utils/PageLink'
 
 const LOGGER = logger.child({ module: 'api/session/create' })
 const TEST_SSN = '199808302393'
+const productNames = ['SE_APARTMENT_RENT', 'SE_ACCIDENT', 'SE_APARTMENT_BRF'] as const
```
robinandeer, line 21: I feel like adding both RENT + BRF is an odd combination -- no real user would do that, right? You either live in a rental OR brf so I would remove one of them

robinandeer (approved):
Nice refactor!

apps/store/src/pages/api/session/create.ts
KarnellSchultz: Very true

PR #1307, opened 2023-01-11
GRW-2023 / Feature / Log CSP violations

alebedev:
## Describe your changes

Send CSP violations to /api/csp-reports
Log those reports so we can see them in Datadog

Ability to detect CSP errors out of development

### DIFF
--- a/apps/store/src/pages/api/csp-reports.ts
+++ b/apps/store/src/pages/api/csp-reports.ts
@@ -1,0 +1,16 @@
+import { NextApiRequest, NextApiResponse } from 'next'
+import logger from '@/services/logger/server'
+
+const cspReports = (req: NextApiRequest, res: NextApiResponse<void>) => {
+  try {
+    const report = JSON.parse(req.body)['csp-report']
+    if (report) {
+      logger.child({ csp: report }).info('csp violation')
+    }
+  } catch (err) {
+    logger.error('Failed to parse CSP report', err)
+  }
+  return res.status(204).send()
+}
+
+export default cspReports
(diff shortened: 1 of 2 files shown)


alebedev:
* main
  * **PR #1307** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1307" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1307?utm_source=stack-comment).

apps/store/src/pages/api/csp-reports.ts
```diff
@@ -0,0 +1,16 @@
+import { NextApiRequest, NextApiResponse } from 'next'
+import logger from '@/services/logger/server'
+
+const cspReports = (req: NextApiRequest, res: NextApiResponse<void>) => {
+  try {
+    const report = JSON.parse(req.body)['csp-report']
```
robinandeer, line 6: Doesn't this get called with content-type headers? In that case Next.js will automatically parse the body

apps/store/src/pages/api/csp-reports.ts
```diff
@@ -0,0 +1,16 @@
...
+    if (report) {
+      logger.child({ csp: report }).info('csp violation')
```
robinandeer, line 8: Aren't we missing logging something here that lets us identify which domain that was blocked?

Should we log info or warn?

robinandeer (approved):
Awesome @alebedev!

apps/store/src/pages/api/csp-reports.ts
alebedev: This is added when sending info to Datadog

PR #1313, opened 2023-01-11
feat(TextBlock): enable color changing

guilhermespopolin:


gustaveen (approved):
Awesome! 🙌 We probably don't need `textDisabled` as an option, check with Zak what colors he wants 👌

KarnellSchultz (approved):

guilhermespopolin:
> Awesome! 🙌 We probably don't need `textDisabled` as an option, check with Zak what colors he wants 👌

It turns out we also don't need `textNegative` as well.

PR #1315, opened 2023-01-11
chore(csp): added basic documentation

guilhermespopolin:
## Describe your changes

* Added basic documentation that explain the reasoning behind some rules

So we remember their purpose

### DIFF
--- a/apps/store/next.config.js
+++ b/apps/store/next.config.js
@@ -24,22 +24,78 @@ module.exports = {
   },
 }
 
+const scriptSrcDirectives = [
+  'https://app.storyblok.com',
+  'https://widget.intercom.io',
+  'https://js.intercomcdn.com',
+  'https://www.googletagmanager.com',
+  'https://dc.insurely.com',
+  'https://vercel.live',
+  "'unsafe-inline'",
+  "'unsafe-eval'",
+  "'self'",
+]
+const styleSrcDirectives = [
+  'fonts.googleapis.com', // GTM preview mode
+  "'unsafe-inline'",
+  "'self'",
+]
+const fontSrcDirectives = [
+  'https://fonts.intercomcdn.com',
+  'fonts.gstatic.com', // GTM preview mode
+  "'self'",
+]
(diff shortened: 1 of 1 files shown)


guilhermespopolin:
* main
  * **PR #1315** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1315" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1315?utm_source=stack-comment).

apps/store/next.config.js
```diff
@@ -26,6 +26,16 @@ module.exports = {
 
 // script-src: NOT SAFE - https://www.hyperxiao.top/en/posts/6
 // style-src: consider emotion + nonce: https://emotion.sh/docs/@emotion/cache#nonce
+/**
+ * style-src:
+ *  - fonts.googleapis.com: required for GTM preview mode
```
guilhermespopolin, line 31: I though about using this format but I'm open for suggestions :)

alebedev: I think it would be better to locale comments close to rules.  Maybe make header components into an array so we can comment on each item next to it.  Then just join an array for final header value

robinandeer: Yeah agree with @alebedev -- we don't need to keep the template string implementation!

apps/store/next.config.js
```diff
@@ -24,22 +24,78 @@ module.exports = {
...
+  'https://js.intercomcdn.com',
+  'https://www.googletagmanager.com',
+  'https://dc.insurely.com',
+  'https://vercel.live',
+  "'unsafe-inline'",
+  "'unsafe-eval'",
+  "'self'",
+]
+const styleSrcDirectives = [
+  'fonts.googleapis.com', // Required for GTM preview mode
```
robinandeer, line 39: nit: just "GTM preview mode" is enough I think

robinandeer (approved):
Nice good improvement

PR #1318, opened 2023-01-12
Implement login endpoint for Swedish BankID method

robinandeer:
## Describe your changes

Integrate the auth service to be able to use the "member login" endpoint.

Add a new API route to make the request to auth service.

This new API route accepts an SSN + optional "next" parameter to redirect the user after successful login. It also sets the auth session cookie.

This will be useful in the future but more so if we want to handle existing members in the checkout. The idea is to make them login before going through the checkout.

I added an API route beacuse CORS blocks us from calling auth service directly from client.

We don't know what the UX will look like so this PR doesn't contain any UI stuff.

I want to co-locate all auth stuff in the same "service" module -- will do in separate PR.

### DIFF
--- a/apps/store/src/pages/api/auth/login/se.ts
+++ b/apps/store/src/pages/api/auth/login/se.ts
@@ -1,0 +1,26 @@
+import { NextApiRequest, NextApiResponse } from 'next'
+import * as Auth from '@/services/Auth/Auth'
+import { memberLoginSE } from '@/utils/auth'
+import { ORIGIN_URL, PageLink } from '@/utils/PageLink'
+
+const handler = async (req: NextApiRequest, res: NextApiResponse) => {
+  console.info('Logging in with BankID')
+
+  const ssn = req.body.ssn
+  if (typeof ssn !== 'string') throw new Error('No ssn provided')
+
+  const accessToken = await memberLoginSE(ssn)
+  console.debug('Login successful')
+
+  Auth.save(accessToken, { req, res })
+
+  const nextQueryParam = req.query['next']
+  const url = typeof nextQueryParam === 'string' ? nextQueryParam : PageLink.home({ locale: 'se' })
+  const nextURL = new URL(url, ORIGIN_URL)
+
+  const destination = nextURL.toString()
+  console.debug(`Re-directing to destination: ${destination}`)
+  return res.redirect(destination)
+}
+
+export default handler
--- a/apps/store/src/utils/auth.ts
+++ b/apps/store/src/utils/auth.ts
@@ -1,30 +1,114 @@
 import { CookieParams, getAuthHeader } from '@/services/Auth/Auth'
 
+enum MemberLoginMethod {
+  SE_BANKID = 'SE_BANKID',
+}
+
+const BASE_URL = process.env.NEXT_PUBLIC_AUTH_ENDPOINT
+const Endpoint = {
+  TOKEN: `${BASE_URL}/oauth/token`,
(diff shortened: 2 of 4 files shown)


robinandeer:
* main
  * **PR #1318** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1318" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1318?utm_source=stack-comment).

alebedev:
General comment - we can allow CORS on per-endpoint basis in auth server.  If this is the only reason to make request server-side, we can move it to the client

apps/store/src/pages/api/auth/login/se.ts
```diff
@@ -0,0 +1,33 @@
...
+  const ssn = req.body.ssn
+  if (typeof ssn !== 'string') throw new Error('No ssn provided')
+
+  const accessToken = await memberLoginSE(ssn)
+  LOGGER.debug('Login successful')
+
+  Auth.save(accessToken, { req, res })
+
+  const nextURL = new URL(ORIGIN_URL)
+  nextURL.pathname = PageLink.home({ locale: 'se' })
```
alebedev, line 21: Can we somehow take client locale?

apps/store/src/pages/api/auth/login/se.ts
```diff
@@ -0,0 +1,33 @@
...
+
+  const nextQueryParam = req.query['next']
+  if (typeof nextQueryParam === 'string') {
```
alebedev, line 24: Is it intentional that we're ignoring any query params in next URL?

apps/store/src/utils/auth.ts
```diff
@@ -1,32 +1,116 @@
 import { CookieParams, getAuthHeader } from '@/services/Auth/Auth'
 
+enum MemberLoginMethod {
+  SE_BANKID = 'SE_BANKID',
+}
+
+const BASE_URL = process.env.NEXT_PUBLIC_AUTH_ENDPOINT
+const Endpoint = {
```
alebedev, line 8: nice!

alebedev (commented):
Looks good.  Only major question I have is client vs backend request

apps/store/src/pages/api/auth/login/se.ts
robinandeer: We would have to pass it in or read from a cookie I guess

Let's worry about it if we decide we need this endpoint in the first place.

apps/store/src/pages/api/auth/login/se.ts
robinandeer: What do you mean? What are we ignoring?

robinandeer:
@alebedev yeah I figured we could enable CORS as well but just wanted to try it out with the setup we have at the moment

apps/store/src/pages/api/auth/login/se.ts
alebedev: Let's say we want `?next=/path?q=1` (properly urlencoded).  The way you set pathname, you'll redirect to `/path` without any query params

alebedev (approved):
Looks good, did round of pair improvements

robinandeer:
Stack job ID: s3qmEzbk8MIL7p5CwkCI.
See details on [graphite.dev](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1318)

PR #1352, opened 2023-01-16
Add size settings to TextBlock

gustaveen:
## Describe your changes
- Refactor Text component to be able to render rich text content
- Update gray color palette
- Add font size settings. I used datasources for this, that way we can define it once and use in multiple places, neat IMO :) 

Different font sizes are needed at different breakpoints. Since there will be different needs depending on context we can't just use a single size (like `sm`) and use a predefined size scale.

Screenshot 2023-01-16 at 17 21 46

Screenshot 2023-01-16 at 17 21 28

Screenshot 2023-01-16 at 17 49 55


gustaveen:
* main
  * **PR #1352** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1352" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1352?utm_source=stack-comment).

apps/store/src/blocks/TextBlock.tsx
```diff
@@ -1,26 +1,29 @@
-import styled from '@emotion/styled'
 import { ISbRichtext, renderRichText, storyblokEditable } from '@storyblok/react'
 import { useMemo } from 'react'
-import { Space, UIColors, getColor } from 'ui'
+import { FontSizes } from 'ui/src/lib/theme/typography'
```
alebedev, line 3: Can it be imported from 'ui' ?

apps/store/src/blocks/TextBlock.tsx
```diff
@@ -1,26 +1,29 @@
...
 
 export const TextBlock = ({ blok }: TextBlockProps) => {
   const contentHtml = useMemo(() => renderRichText(blok.body), [blok.body])
   return (
     <Text
       {...storyblokEditable(blok)}
+      as="div"
       color={blok.color ?? 'textPrimary'}
       y={1}
+      size={{ _: blok.fontSize ?? 'md', ...(blok.fontSizeDesktop && { lg: blok.fontSizeDesktop }) }}
```
alebedev, line 24: New object every render, can we memo it?

alebedev (approved):
Great job, especially editing experience!

apps/store/src/blocks/TextBlock.tsx
gustaveen: ah, good catch! 👌

apps/store/src/blocks/TextBlock.tsx
gustaveen: In these cases I always wonder if there's any performance benefit memoizing? :D

guilhermespopolin (approved):
Well done :)

PR #1359, opened 2023-01-17
refact(Heading): enabled responsive variant

guilhermespopolin:


guilhermespopolin:
* main
  * **PR #1359** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1359" target="_blank">Graphite</a>  👈
    * **PR #1360** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1360" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1359?utm_source=stack-comment).

packages/ui/src/components/Heading/Heading.helpers.ts
```diff
@@ -1,100 +1,112 @@
...
-      72: {
-        fontSize: theme.fontSizes[10],
-        fontFamily: theme.fonts.big,
-      },
-      96: {
-        fontSize: theme.fontSizes[11],
-        fontFamily: theme.fonts.big,
-      },
-    },
+export const getHeadingVariantStyles = (variant: HeadingVariant) => {
```
alebedev, line 84: Can we reuse `getFontSize()` here?  Looks like it solves the same problem

alebedev (approved):
Good change, let's make sure we have same API and implementation as for Text

packages/ui/src/components/Heading/Heading.helpers.ts
guilhermespopolin: The issue is that in case of `Heading` we're changing the variant in a responsive way, meaning changing more than one thing at the same time: fontSize and lineHeight for example.
Something that we could maybe do is not implement those utilities and rely on something like [`@styled-system/typography`](https://github.com/styled-system/styled-system/tree/master/packages/typography) and [`@styled-system/variant`](https://github.com/styled-system/styled-system/tree/master/packages/variant)

packages/ui/src/components/Heading/Heading.helpers.ts
```diff
@@ -1,100 +1,112 @@
...
+import { theme } from '../../lib/theme/theme'
 
-type HeadingSize = '18' | '20' | '24' | '32' | '40' | '48' | '56' | '72' | '96'
+type StandardHeadingSize = '18' | '20' | '24' | '32' | '40' | '48' | '56' | '72' | '96'
+type SerifHeadingSize = Exclude<StandardHeadingSize, '18'>
+type PossibleHeadingVariant = `standard.${StandardHeadingSize}` | `serif.${SerifHeadingSize}`
 
-type HeadingStyles = {
-  [key in HeadingSize]: object
+const headings: Record<PossibleHeadingVariant, CSSObject> = {
```
gustaveen, line 9: I suppose it was easier to have it as a flat object :D

packages/ui/src/components/Heading/Heading.helpers.ts
gustaveen: It's a bit trickier with headings since they also set different font-family and letter-spacing. But I suppose we could refactor it to cater to both use cases

gustaveen (approved):
Awesome! 🙌 Let's edit the storybook story for Headings to display the use cases as well :)

PR #1388, opened 2023-01-20
Fix background color on header

gustaveen:
## Describe your changes
Update bg color to be same as rest of the page


amandaflod (approved):
💅

PR #1411, opened 2023-01-23
Add Checkout breadcrumbs

robinandeer:
## Describe your changes

Update Checkout page with breadcrumbs from Figma.

## Checklist before requesting a review

- [ ] I have performed a self-review of my code

PR #1410, opened 2023-01-23
Restore lint-staged settings

robinandeer:
## Describe your changes

Run eslint as part of lint-staged.

Remove eslint directly from pre-commit hook.

## Justify why they are needed

Not sure why this is still here, I thought Aleks had removed it.

## Checklist before requesting a review

- [ ] I have performed a self-review of my code

PR #1409, opened 2023-01-23
Update details sheet

robinandeer:
## Describe your changes

Update cart item details sheet to match the new design.

Move documents to top of page.

## Checklist before requesting a review

- [ ] I have performed a self-review of my code

PR #1412, opened 2023-01-23
Add BankID modal to Checkout page

robinandeer:
## Describe your changes

Implement the new BankID modal while waiting for BankID signing to complete.


robinandeer:
* main
  * **PR #1409** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1409" target="_blank">Graphite</a> 
    * **PR #1410** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1410" target="_blank">Graphite</a> 
      * **PR #1411** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1411" target="_blank">Graphite</a> 
        * **PR #1412** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1412" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1412?utm_source=stack-comment).

apps/store/src/services/Checkout/useHandleSignShopSession.ts
```diff
@@ -32,7 +32,6 @@ export const useHandleSignShopSession = (params: Params) => {
     async onCompleted(data) {
       const { status, completion } = data.shopSessionSigning
       if (status === ShopSessionSigningStatus.Signed && completion) {
-        setShopSessionSigningId(null)
```
robinandeer, line 35: This causes the query result to be cleared but we want to keep it around to read the signing status

apps/store/src/services/Checkout/useHandleSignShopSession.ts
```diff
@@ -43,6 +42,7 @@ export const useHandleSignShopSession = (params: Params) => {
     },
     onError(error) {
       datadogLogs.logger.warn('Checkout | SigningQuery | Failed to sign', { error })
+      setShopSessionSigningId(null)
```
robinandeer, line 45: We probably need to create a new signing if this fails

apps/store/src/services/Checkout/useHandleSignShopSession.ts
```diff
@@ -62,11 +62,15 @@ export const useHandleSignShopSession = (params: Params) => {
     },
   })
 
+  const signingStatus = queryResult.data?.shopSessionSigning.status
+  const signingFailed = signingStatus === ShopSessionSigningStatus.Failed
+  const isSigning = Boolean(shopSessionSigningId) && !signingFailed
```
robinandeer, line 67: Have to update this since we now keep the `shopSessionId` state after success

apps/store/src/components/CheckoutPage/CheckoutPage.tsx
```diff
@@ -140,21 +149,51 @@ const CheckoutPage = (props: CheckoutPageProps) => {
...
-            </Button>
-          </FullscreenDialog.Close>
-        }
-      >
-        <ErrorPrompt size={{ _: 'md', lg: 'lg' }} align="center">
-          {t('ERROR_GENERAL_DIALOG_PROMPT')}
-        </ErrorPrompt>
-      </FullscreenDialog.Modal>
-    </FullscreenDialog.Root>
+      <FullscreenDialog.Root open={showLoading} onOpenChange={(open) => setHideLoading(!open)}>
```
robinandeer, line 152: We could refactor to defined this in a separate file

apps/store/src/components/CheckoutPage/CheckoutPage.tsx
```diff
@@ -49,13 +52,19 @@ const CheckoutPage = (props: CheckoutPageProps) => {
...
     },
   })
 
-  const submitButtonMessage = loading ? t('OPEN_BANKID_DESCRIPTION') : t('SIGN_DISCLAIMER')
+  const signFailed = signingStatus === ShopSessionSigningStatus.Failed
+  const signFailedMessage = signFailed ? t('UNKNOWN_ERROR_MESSAGE', { ns: 'common' }) : undefined
+  const userErrorMessage = userError?.message ?? signFailedMessage
+
+  const isSigned = signingStatus === ShopSessionSigningStatus.Signed
+  const showLoading = (loading || isSigned) && !hideLoading
```
robinandeer, line 64: We want to keep showing the loading if signing is successful until we navigate to the next page

apps/store/src/components/CheckoutPage/CheckoutPage.tsx
```diff
@@ -34,7 +36,8 @@ const CheckoutPage = (props: CheckoutPageProps) => {
   const router = useRouter()
   const apolloClient = useApolloClient()
   const tracking = useTracking()
-  const [handleSubmitSign, { loading, userError }] = useHandleSubmitCheckout({
+  const [hideLoading, setHideLoading] = useState(false)
```
robinandeer, line 39: Allow the user to hide the loading modal even if there's an ongoing request

apps/store/src/components/CheckoutPage/BankIdIcon.tsx
```diff
@@ -1,42 +1,51 @@
...
+import { theme } from 'ui'
 
-export const BankIdIcon = () => {
-  const theme = useTheme()
+type Props = {
+  color: keyof typeof theme.colors
+  size?: string
+}
 
+export const BankIdIcon = ({ color, size = '1.125rem' }: Props) => {
```
alebedev, line 8: Nit:  You can use color="currentColor" for monochrome icons, then set color on parent.  This way icon itself would not need theme

apps/store/src/components/CheckoutPage/CheckoutPage.tsx
alebedev: What would it mean if we hide a modal?  Could lead to confusion when request succeeds in background

apps/store/src/components/CheckoutPage/CheckoutPage.tsx
alebedev, line 64: Looks complicated.  Let's try refactoring it to explicit state enum

apps/store/src/components/CheckoutPage/CheckoutPage.tsx
```diff
@@ -140,21 +149,51 @@ const CheckoutPage = (props: CheckoutPageProps) => {
...
+                    {t('BANKID_MODAL_CANCEL')}
+                  </Button>
+                </FullscreenDialog.Close>
+              </>
+            )}
+          </SpaceFlex>
+        </FullscreenDialog.Modal>
+      </FullscreenDialog.Root>
+
+      <FullscreenDialog.Root open={showSignError} onOpenChange={setShowSignError}>
```
alebedev, line 180: Can we reuse same modal for error?

apps/store/src/components/CheckoutPage/TickIcon.tsx
```diff
@@ -0,0 +1,25 @@
...
+      height={size}
+      viewBox="0 0 16 16"
+      fill="none"
+      xmlns="http://www.w3.org/2000/svg"
+    >
+      <path
+        fill-rule="evenodd"
+        clip-rule="evenodd"
+        d="M15.5117 3.45171C15.8146 3.73433 15.8309 4.20893 15.5483 4.51174L7.67707 12.9452C6.58989 14.11 4.74345 14.11 3.65627 12.9452L0.451711 9.51174C0.169086 9.20893 0.185451 8.73433 0.488264 8.45171C0.791077 8.16908 1.26567 8.18545 1.54829 8.48826L4.75285 11.9217C5.24702 12.4512 6.08632 12.4512 6.58049 11.9217L14.4517 3.48826C14.7343 3.18545 15.2089 3.16908 15.5117 3.45171Z"
+        fill={theme.colors[color]}
```
alebedev, line 21: You can use currentColor, as above

apps/store/src/services/Checkout/useHandleSignShopSession.ts
alebedev, line 67: Maybe we can explicitly check for "in progress" signing statuses instead?

alebedev (approved):
Looks good, some thoughts about possible improvements in comments

We can merge this now and improve later if you think it's the fastest way to go forward

apps/store/src/components/CheckoutPage/CheckoutPage.tsx
robinandeer: After success, we don't allow hiding the modal -- it's accounted for below

apps/store/src/components/CheckoutPage/CheckoutPage.tsx
robinandeer: Right, yes good point

apps/store/src/components/CheckoutPage/CheckoutPage.tsx
robinandeer: Hmm but what if there's error while loading?

Would be nice to not mix these up too much -- I was thinking of making two components out of these so it doesn't look cluttered

apps/store/src/services/Checkout/useHandleSignShopSession.ts
robinandeer: Hmm does that cover everything tho?

We want to show "singing" also after mutation and before the first "signing" query returns => then "signing status" would be undefined

PR #1421, opened 2023-01-23
Confirmation page step content ✅

KarnellSchultz:
## Describe your changes

- Created a `<CheckList />` component to be used on the confirmation page. This component is presentational with 'checkboxes' that are not interactive.
The Component will also show users the google play / apple app store button based on which platform the user is on. Or show both to let the user make the decision on which is best for them.

TODO: Translations coming in future PR. 

Screenshot 2023-01-23 at 16 18 01

Per latest designs for the confirmation page.

### DIFF
--- a/apps/store/src/components/ConfirmationPage/CheckList.tsx
+++ b/apps/store/src/components/ConfirmationPage/CheckList.tsx
@@ -7,35 +7,32 @@ type ListItemProps = {
   children?: React.ReactNode
 }
 
-export const ListItemChecked = ({ title }: ListItemProps) => {
-  return (
-    <CheckListItem checked>
-      <SpaceFlex align="center" space={0.5}>
-        <PresentationCheckboxChecked>
-          <CheckIcon size="1rem" />
-        </PresentationCheckboxChecked>
-        <Text>{title}</Text>
-      </SpaceFlex>
-    </CheckListItem>
-  )
-}
-export const ListItemUnchecked = ({ children, title }: ListItemProps) => {
-  return (
-    <CheckListItem>
-      <SpaceFlex align="center" space={0.5}>
-        <PresentationCheckboxUnchecked />
-        <Text>{title}</Text>
-      </SpaceFlex>
-      <CheckListItemContent>{children}</CheckListItemContent>
-    </CheckListItem>
-  )
-}
+export const CheckedListItem = ({ title }: ListItemProps) => (
+  <CheckListItem checked>
+    <SpaceFlex align="center" space={0.5}>
+      <PresentationCheckboxChecked>
+        <CheckIcon size="1rem" />
+      </PresentationCheckboxChecked>
+      <Text>{title}</Text>
+    </SpaceFlex>
+  </CheckListItem>
+)
+
+export const UncheckedListItem = ({ children, title }: ListItemProps) => (
+  <CheckListItem>
+    <SpaceFlex align="center" space={0.5}>
+      <PresentationCheckboxUnchecked />
+      <Text>{title}</Text>
+    </SpaceFlex>
+    <CheckListItemContent>{children}</CheckListItemContent>
+  </CheckListItem>
+)
 
 export const StyledCheckList = styled.ul({
-  listStyle: 'none',
   display: 'flex',
   flexDirection: 'column',
   gap: theme.space.xxs,
+  listStyle: 'none',
 })
 
 const CheckListItem = styled.li<{ checked?: boolean }>(({ checked = false }) => ({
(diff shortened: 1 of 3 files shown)


apps/store/src/components/ConfirmationPage/CheckList.tsx
```diff
@@ -0,0 +1,76 @@
+import styled from '@emotion/styled'
+import { CheckIcon, Text, theme } from 'ui'
+import { SpaceFlex } from '@/components/SpaceFlex/SpaceFlex'
+
+type ListItemProps = {
+  title: string
+  children?: React.ReactNode
+}
+
+export const CheckedListItem = ({ title }: ListItemProps) => (
```
alebedev, line 10: I wonder if it's good idea to have different components with different API (children vs no children) for checked and unchecked items

Would single component that takes `checked, title, children?` be easier to use?

apps/store/src/components/ConfirmationPage/CheckList.tsx
```diff
@@ -0,0 +1,76 @@
...
+})
+
+const CheckListItem = styled.li<{ checked?: boolean }>(({ checked = false }) => ({
+  display: 'flex',
+  flexDirection: 'column',
+  justifyContent: 'center',
+  backgroundColor: theme.colors.gray100,
+  borderRadius: theme.radius.sm,
+  padding: theme.space.md,
+  paddingInline: theme.space.md,
```
alebedev, line 45: Same as padding, is it needed?

apps/store/src/components/ConfirmationPage/CheckList.tsx
```diff
@@ -0,0 +1,76 @@
...
+    color: theme.colors.textDisabled,
+    textDecoration: 'line-through',
+  }),
+}))
+
+const CheckListItemContent = styled.div({
+  paddingLeft: `calc(${theme.space.lg} + ${theme.space.xs})`,
+})
+
+const PresentationCheckboxBase = styled.div({
```
alebedev, line 60: If you're not using it as component, can you just reuse CSS fragment?  I think I've seen something like this in emotion

apps/store/src/components/ConfirmationPage/CheckList.tsx
```diff
@@ -0,0 +1,76 @@
...
+    </SpaceFlex>
+    <CheckListItemContent>{children}</CheckListItemContent>
+  </CheckListItem>
+)
+
+export const StyledCheckList = styled.ul({
+  display: 'flex',
+  flexDirection: 'column',
+  gap: theme.space.xxs,
+  listStyle: 'none',
```
robinandeer, line 35: nit: isn't this already part of the CSS reset?

apps/store/src/components/ConfirmationPage/ConfirmationPage.tsx
```diff
@@ -1,10 +1,22 @@
 import styled from '@emotion/styled'
+import Link from 'next/link'
 import { Heading, mq, Space, Text, theme } from 'ui'
 import { CartInventory } from '@/components/CartInventory/CartInventory'
+import { useCurrentLocale } from '@/utils/l10n/useCurrentLocale'
+import { AppStoreBadge } from '../AppStoreBadge/AppStoreBadge'
+import { SpaceFlex } from '../SpaceFlex/SpaceFlex'
```
robinandeer, line 7: nit: update import to include `@/components`

apps/store/src/components/ConfirmationPage/ConfirmationPage.tsx
```diff
@@ -17,8 +29,42 @@ export const ConfirmationPage = (props: ConfirmationPageProps) => {
...
+                    Ladda ner Hedvig appen till din mobil och kom igång med din nya försäkring.
+                  </Text>
+                </div>
+                <Space y={0.5}>
+                  <StyledCheckList>
+                    <CheckedListItem title="Teckna Hedvig" />
+                    <CheckedListItem title="Koppla autogiro" />
+                    <UncheckedListItem title="Ladda ner Hedvig appen">
+                      {platform ? (
+                        <AppStoreBadge type={platform} locale={locale} />
```
alebedev, line 52: Should it be a link too?  Would be nice to open appstore with single touch

apps/store/src/components/ConfirmationPage/ConfirmationPage.tsx
```diff
@@ -17,8 +29,42 @@ export const ConfirmationPage = (props: ConfirmationPageProps) => {
...
+                      ) : (
+                        <SpaceFlex space={0.5}>
+                          <Link href={appStoreLinks.apple} passHref>
+                            <AppStoreBadge type={'apple'} locale={locale} />
```
robinandeer, line 56: nit: update to `type="apple"` 😛

alebedev (changes requested):
Good work!  Some comments on implementation details

robinandeer (approved):
[ismellmoney](https://app.graphite.dev/macros/HedvigInsurance)

apps/store/src/components/ConfirmationPage/ConfirmationPage.tsx
KarnellSchultz: 🍎

apps/store/src/components/ConfirmationPage/ConfirmationPage.tsx
KarnellSchultz: Yes, should be linked for sure 👍

apps/store/src/components/ConfirmationPage/CheckList.tsx
KarnellSchultz: 🧽

apps/store/src/components/ConfirmationPage/CheckList.tsx
KarnellSchultz: Thanks, I did not know that was a thing 👍

apps/store/src/components/ConfirmationPage/CheckList.tsx
KarnellSchultz: I reworked this component in an effort to make it more simple to use and to maintain going forward. 

_A few things that I was thinking about when putting the component together_
At the moment this is only being used in the one place so it's not like we're making the most reusable component ever, but I agree with you that the two different API's is not ideal. I updated that so they're the same. 
I think it's useful to have these two components to be separate because when we end up changing or updating the component with future design changes the person doing that work will not have to think about any if/else kind of logic. They'd be able to just work on the checked or unchecked component directly. 

I ended up making the component a compound component and removing the `checked` prop passing that was being done within the component. I _think_ I'm made it more simple. But perhaps I've overcooked it. 

Let me know if you have any more feedback or think I should go another way with this one.

tack tack

alebedev (approved):
Looks good!

PR #1425, opened 2023-01-24
Update country-selector page styles

KarnellSchultz:
## Describe your changes

Added the wrapper from pages like the cart/confirmation page to the country-selector page. Seems like these should all have the same or similar widths.

Screenshot 2023-01-24 at 10 31 45

Consistant  styles across the app. 


robinandeer (approved):
Nice :D

When we implement this as a reusable component -- I wonder if we should just make it a 12 column grid like it is described in Figma 🤔

KarnellSchultz:
> Nice :D
> 
> When we implement this as a reusable component -- I wonder if we should just make it a 12 column grid like it is described in Figma 🤔

Did not notice the designers updated the page width things yet. but yeah, sounds like a good idea for sure 👍

PR #1430, opened 2023-01-24
Track select_item ecommerce event when user visits product page

alebedev:
## Describe your changes

Subj

Fist step in conversion funnel.  Easier to analyze when we have separate event for product pages instead of relying on virtual_page_view


alebedev:
* main
  * **PR #1430** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1430" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1430?utm_source=stack-comment).

robinandeer (approved):
Let's keep this for now but one of the things I remember hearing from React core devs is this:

They advocate handling e.g. this event when someone clicks on the link to the product page rather than waiting for the product page to load and then send the analytics event.

I think this does seem to make sense -- we execute side effects in the event handler rather than as a result of rendering.

I also understand that it is a bit of a different way of thinking and might be hard to implement in places. However, I would say we should always consider such a solution :) makes sense?

PR #1490, opened 2023-01-30
Setup Storyblok to be used on the confirmation page

KarnellSchultz:
## Describe your changes

Looking to create a page where some of the content is hardcoded and the other is managed in Storyblok. 

- Creating a `confirmation` component in Storyblok and in code.

Designers want the ability to changes things around on parts of the confirmation page, these changes would be a step to allowing that to happen.

### DIFF
--- a/apps/store/src/components/ConfirmationPage/ConfirmationPage.tsx
+++ b/apps/store/src/components/ConfirmationPage/ConfirmationPage.tsx
@@ -4,13 +4,22 @@ import { Heading, mq, Space, Text, theme } from 'ui'
 import { ConfirmationPageBlock } from '@/blocks/ConfirmationPageBlock'
 import { CartInventory } from '@/components/CartInventory/CartInventory'
 import { SpaceFlex } from '@/components/SpaceFlex/SpaceFlex'
-import { appStoreLinks } from '@/utils/appStoreLinks'
+import { ConfirmationStory } from '@/services/storyblok/storyblok'
 import { useCurrentLocale } from '@/utils/l10n/useCurrentLocale'
 import { AppStoreBadge } from '../AppStoreBadge/AppStoreBadge'
 import { CheckList, CheckListItem } from './CheckList'
 import { ConfirmationPageProps } from './ConfirmationPage.types'
 
-export const ConfirmationPage = (props: ConfirmationPageProps) => {
+const appStoreLinks = {
+  apple: 'https://apps.apple.com/se/app/id1303668531?l=en',
+  google: 'https://play.google.com/store/apps/details?id=com.hedvig.app',
+} as const
+
+type Props = ConfirmationPageProps & {
+  story: ConfirmationStory
+}
+
+export const ConfirmationPage = (props: Props) => {
   const { locale } = useCurrentLocale()
   const { platform, cart, story } = props
 
--- a/apps/store/src/components/ConfirmationPage/ConfirmationPage.tsx
+++ b/apps/store/src/components/ConfirmationPage/ConfirmationPage.tsx
@@ -5,15 +5,11 @@ import { ConfirmationPageBlock } from '@/blocks/ConfirmationPageBlock'
 import { CartInventory } from '@/components/CartInventory/CartInventory'
 import { SpaceFlex } from '@/components/SpaceFlex/SpaceFlex'
 import { ConfirmationStory } from '@/services/storyblok/storyblok'
+import { appStoreLinks } from '@/utils/appStoreLinks'
 import { useCurrentLocale } from '@/utils/l10n/useCurrentLocale'
 import { AppStoreBadge } from '../AppStoreBadge/AppStoreBadge'
 import { CheckList, CheckListItem } from './CheckList'
 import { ConfirmationPageProps } from './ConfirmationPage.types'
-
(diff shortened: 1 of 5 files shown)


apps/store/src/components/ConfirmationPage/ConfirmationPage.tsx
```diff
@@ -1,17 +1,27 @@
...
 import { SpaceFlex } from '@/components/SpaceFlex/SpaceFlex'
-import { appStoreLinks } from '@/utils/appStoreLinks'
+import { ConfirmationStory } from '@/services/storyblok/storyblok'
 import { useCurrentLocale } from '@/utils/l10n/useCurrentLocale'
 import { AppStoreBadge } from '../AppStoreBadge/AppStoreBadge'
 import { CheckList, CheckListItem } from './CheckList'
 import { ConfirmationPageProps } from './ConfirmationPage.types'
 
-export const ConfirmationPage = (props: ConfirmationPageProps) => {
+const appStoreLinks = {
```
robinandeer, line 13: why not import this?

apps/store/src/pages/confirmation/[shopSessionId].tsx
```diff
@@ -44,13 +47,14 @@ export const getServerSideProps: GetServerSideProps<ConfirmationPageProps, Param
...
     },
   })
 }
 
-const CheckoutConfirmationPage: NextPageWithLayout<ConfirmationPageProps> = (props) => {
-  return <ConfirmationPage {...props} />
-}
+const CheckoutConfirmationPage: NextPageWithLayout<
+  ConfirmationPageProps & { story: ConfirmationStory }
+> = (props) => <ConfirmationPage {...props} />
```
Comment (robinandeer) [inline apps/store/src/pages/confirmation/[shopSessionId].tsx:57]: I would consider exporting props from the page component instead of redefining them here

apps/store/src/services/storyblok/storyblok.ts
```diff
@@ -150,6 +157,7 @@ export const initStoryblok = () => {
     CheckListBlock,
     ContactSupportBlock,
     ContentBlock,
+    ConfirmationPageBlock,
```
robinandeer, line 160: Is this really a block that can be reused across different pages or rather just a different content type/page?

If it's the former we don't need to add it this this list 🙂

robinandeer (commented):
Nice work -- will be interesting to see how this works!

apps/store/src/services/storyblok/storyblok.ts
gustaveen: Isn't it need to show the preview in Storyblok though? :)

apps/store/src/pages/confirmation/[shopSessionId].tsx
Comment (KarnellSchultz) [reply apps/store/src/pages/confirmation/[shopSessionId].tsx:57]: I did it this way because I could not figure out how to make TS happy on the `getServerSideProps` when adding `{ story: ConfirmaitonStory }` to `ConfirmationPageProps`. Spend a good deal of time trying to sort it to no avail. 

If you've got some time to check it out with me let me know

apps/store/src/pages/confirmation/[shopSessionId].tsx
```diff
@@ -3,32 +3,35 @@ import { serverSideTranslations } from 'next-i18next/serverSideTranslations'
...
 import { addApolloState, initializeApollo } from '@/services/apollo/client'
 import { SHOP_SESSION_PROP_NAME } from '@/services/shopSession/ShopSession.constants'
 import { setupShopSessionServiceServerSide } from '@/services/shopSession/ShopSession.helpers'
-import { getGlobalStory } from '@/services/storyblok/storyblok'
+import { ConfirmationStory, getGlobalStory, getStoryBySlug } from '@/services/storyblok/storyblok'
 import { GLOBAL_STORY_PROP_NAME } from '@/services/storyblok/Storyblok.constant'
 import { getMobilePlatform } from '@/utils/getMobilePlatform'
 import { isRoutingLocale } from '@/utils/l10n/localeUtils'
 
+const CONFIRMATION_PAGE_SLUG = 'confirmation'
```
Comment (gustaveen) [inline apps/store/src/pages/confirmation/[shopSessionId].tsx:14]: I suppose it would be somewhat tricky to make the slug dynamic. But this is probably not a page that's crucial to translate the url

gustaveen (commented):
Great job Karnell! 🙌

apps/store/src/services/storyblok/storyblok.ts
KarnellSchultz: That's what I was thinking @gustaveen. If I remove it from here then it's not going to show up in SB.

apps/store/src/pages/confirmation/[shopSessionId].tsx
Comment (KarnellSchultz) [reply apps/store/src/pages/confirmation/[shopSessionId].tsx:14]: We could pick the slug off by doing something like this.

```
const slug = resolvedUrl.split('/').filter((item) => item !== '')[0]
```

or just setup some kind of function to get the name of the slug based on the locale

What do you think?

Comment (gustaveen) [reply apps/store/src/pages/confirmation/[shopSessionId].tsx:14]: Right! Some function to get the slug based on locale could work, since this will only be used for one page/routes right? Otherwise some thing more dynamic

gustaveen (approved):
🙌

PR #1492, opened 2023-01-30
Pause video when purchase form is open (mobile)

robinandeer:
## Describe your changes

See title.

Refactor purchase form dialog a little bit.

We don't want video to be playing in the background.

Wanted to come up with a solution that doesn't cause coupling between components so went with sending custom events.

### DIFF
--- a/apps/store/src/components/ProductPage/PurchaseForm/PurchaseForm.tsx
+++ b/apps/store/src/components/ProductPage/PurchaseForm/PurchaseForm.tsx
@@ -20,6 +20,7 @@ import { ShopSession } from '@/services/shopSession/ShopSession.types'
 import { useShopSession } from '@/services/shopSession/ShopSessionContext'
 import { TrackingContextKey } from '@/services/Tracking/Tracking'
 import { useTracking } from '@/services/Tracking/useTracking'
+import { sendDialogEvent } from '@/utils/dialogEvent'
 import { useFormatter } from '@/utils/useFormatter'
 import { ScrollPast } from '../ScrollPast/ScrollPast'
 import { CircledHSuperscript } from './CircledHSuperscript'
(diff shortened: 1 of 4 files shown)


robinandeer:
* main
  * **PR #1492** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1492" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1492?utm_source=stack-comment).

apps/store/src/components/Video/Video.tsx
```diff
@@ -75,6 +85,26 @@ export const Video = ({
     [setState, onPause],
   )
 
+  useEffect(() => {
+    const blurListener = () => {
+      if (state === State.Playing) {
+        pauseVideo()
+      }
+    }
+    window.addEventListener('dialog:open', blurListener)
```
alebedev, line 94: Can we use enum here?

apps/store/src/components/ProductPage/PurchaseForm/PurchaseForm.tsx
```diff
@@ -38,13 +39,20 @@ export const PurchaseForm = () => {
   const formatter = useFormatter()
   const [{ priceIntent }, setupPriceIntent] = usePriceIntent()
   const tracking = useTracking()
+  const isLarge = useBreakpoint('lg')
 
+  const editForm = () => {
+    setFormState('EDIT')
+    if (!isLarge) {
+      sendDialogEvent('open')
```
alebedev, line 47: Can we do it on dialog state change instead of here?

robinandeer: hmm yeah I also feel like it's awkward here but this is the event handler that opens the modal -- so in that sense it is the correct place, right?

Or do you mean we should add a `useEffect` hook inside the dialog component?

alebedev: I was thinking about emiting event in `onOpenChange` of dialog.  We can even make it part of our dialog wrapper in `ui` to get it working on every dialog by default

alebedev (approved):

apps/store/src/components/ProductPage/PurchaseForm/PurchaseForm.tsx
robinandeer: Right, yes I tried that but unfortunately it's only for "managed events" like clicking the trigger, pressing escape, etc.

When the event originates from our code it doesn't get called unfortunately :P

robinandeer:
Stack job ID: 6FxEzOYadpudcV0mnNmD.
See details on [graphite.dev](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1492)

PR #1502, opened 2023-01-31
Remove email field for returning members

alebedev:
## Describe your changes

For returning members
- remove email field from price calculator
- remove email field from checkout

We're not going to use it, so let's remove email from UI


alebedev:
* main
  * **PR #1502** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1502" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1502?utm_source=stack-comment).

apps/store/src/components/CheckoutPage/CheckoutPage.tsx
```diff
@@ -32,8 +32,8 @@ const CheckoutPage = (props: CheckoutPageProps) => {
   const {
     cart,
     ssn,
-    prefilledData,
-    collectName,
+    shouldCollectEmail,
+    shouldCollectName,
```
robinandeer, line 36: Future: perhaps we should reverse this and just list which fields we should show instead of a bunch of booleans

apps/store/src/utils/customer.ts
```diff
@@ -0,0 +1,12 @@
+import { ShopSessionAuthenticationStatus, ShopSessionCustomer } from '@/services/apollo/generated'
+
+const isNewMember = (customer: ShopSessionCustomer) =>
+  customer.authenticationStatus === ShopSessionAuthenticationStatus.None
+
+export const shouldCollectEmail = (customer?: ShopSessionCustomer | null) => {
```
robinandeer, line 6: nit: I would rename these as `getShouldCollectEmail` or similar to differentiate from the variable names

apps/store/src/components/PriceCalculator/PriceCalculator.tsx
```diff
@@ -48,8 +49,12 @@ export const PriceCalculator = (props: Props) => {
 
   const [handleSubmit, handleSubmitSection, isLoading] = useHandleSubmitPriceCalculator({
     onSuccess({ priceIntent, customer }) {
-      const userData = { ...customer, ...priceIntent.data }
-      const form = setupForm(priceTemplate, userData, priceIntent.suggestedData)
+      const form = setupForm({
+        customer: shopSession.customer,
+        suggestedData: priceIntent.suggestedData,
+        template: priceTemplate,
+        userData: { ...customer, ...priceIntent.data },
```
robinandeer, line 56: on line 40 we use a function `getFormData` -- why not here?

apps/store/src/components/PriceCalculator/PriceCalculator.tsx
```diff
@@ -48,8 +49,12 @@ export const PriceCalculator = (props: Props) => {
 
   const [handleSubmit, handleSubmitSection, isLoading] = useHandleSubmitPriceCalculator({
     onSuccess({ priceIntent, customer }) {
-      const userData = { ...customer, ...priceIntent.data }
-      const form = setupForm(priceTemplate, userData, priceIntent.suggestedData)
+      const form = setupForm({
+        customer: shopSession.customer,
```
robinandeer, line 53: nit: we already pass in `customer` inside `userData` -- perhaps it will confuse things that we also pass it here?

robinandeer (approved):
Think it looks good! Great that you managed to work it into the current solution without too much refactoring even if it's needed

apps/store/src/components/CheckoutPage/CheckoutPage.tsx
alebedev: I agree. Let's see what new requirements we're going to get before figuring out the right abstraction

PR #1503, opened 2023-01-31
Fix / Add shadow to nav content

amandaflod:
## Describe your changes

Add shadow to theme and use on nav bar content list


robinandeer (commented):
Whoops I think I did the same thing -- but just added it directly to the nav 😅

PR #1504, opened 2023-01-31
Update insurance list in navigation

robinandeer:
## Describe your changes

Update insurances section in main navigation.

Refactor navigation a little bit better match Figma design.

I focused on getting it to work and behave like design. However, I believe there's a lot of complexity in this part of the code base. Therefore, we should refactor it soon. I think we should define mobile and desktop styles separetly since a lot of complexity is just because we are trying to use the same components for two different layouts.

I excluded the button block list since design only supports a single link that leads to the store page.


robinandeer:
* main
  * **PR #1504** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1504" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1504?utm_source=stack-comment).

apps/store/src/components/Header/NavigationLink.tsx
```diff
@@ -1,33 +1,31 @@
...
   flexShrink: 0,
+
   [mq.lg]: {
     padding: `${theme.space.md} ${theme.space.lg}`,
     borderRadius: theme.radius.sm,
-    ':hover': {
-      backgroundColor: theme.colors.gray100,
+    '@media (hover: hover)': {
+      '&:hover': {
+        backgroundColor: theme.colors.grayTranslucent100,
```
amandaflod, line 27: Thank you for catching these! :)

amandaflod (approved):
Great job! I agree that we should get rid of some of the complexity here, since the components are soooo different 🤠

PR #1507, opened 2023-02-01
Maintenance Day - Swap to theme imports

KarnellSchultz:
## Describe your changes

Went from something like this: 
```
const Paragraph = styled.p(({ theme }) => ({
  color: theme.colors.gray700,
}))
```
to something like this: 

```
import {theme} from 'ui'

const Paragraph = styled.p({
  color: theme.colors.gray700,
})
```

As a team we talked about moving away from using the function syntax with Emotion and importing our theme directly from our ui package. This PR is a step in that direction.

### DIFF
--- a/apps/store/src/components/CheckList/CheckList.tsx
+++ b/apps/store/src/components/CheckList/CheckList.tsx
@@ -1,6 +1,6 @@
 import styled from '@emotion/styled'
 import { PropsWithChildren } from 'react'
-import { CheckIcon } from 'ui'
+import { CheckIcon, theme } from 'ui'
 
 export const Root = styled.ul({
   padding: 0,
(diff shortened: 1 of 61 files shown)


apps/store/src/components/CheckList/CheckList.tsx
```diff
@@ -17,11 +17,11 @@ export const Item = ({ children }: PropsWithChildren<unknown>) => {
   )
 }
 
-const StyledItem = styled.li(({ theme }) => ({
+const StyledItem = styled.li({
   display: 'flex',
   alignItems: 'center',
-  gap: theme.space[2],
-}))
+  gap: theme.space.xs,
```
amandaflod, line 23: sneaky 🕵️‍♀️

amandaflod (approved):
Looks good to me!

apps/store/src/components/CheckList/CheckList.tsx
KarnellSchultz: Tried not to changes those 👀 

Maybe the next PR though???

gustaveen (approved):
Ouff! MVP of maintenance day right there! ⭐️

PR #1509, opened 2023-02-01
Add global product metadata to layout header

robinandeer:
## Describe your changes

Introduce the concept of global product metadata that can be used in places like the layout header/navigation.

Currently implemented as React context around the layout component.

Add pillow image option to nav item in Storyblok to be able to override the default image.

To be able to render product info in global places like the layout header.

I don't want to fetch this on every render but rather fetch onces during build and just pass it down to the layout component like other static CMS-driven data.

### DIFF
--- a/apps/store/src/blocks/HeaderBlock.tsx
+++ b/apps/store/src/blocks/HeaderBlock.tsx
@@ -22,7 +22,13 @@ import {
 } from '@/components/Header/NavigationLink'
 import { TopMenuDesktop } from '@/components/Header/TopMenuDesktop/TopMenuDesktop'
 import { TopMenuMobile } from '@/components/Header/TopMenuMobile/TopMenuMobile'
-import { ExpectedBlockType, LinkField, SbBaseBlockProps } from '@/services/storyblok/storyblok'
+import { useProductMetadata } from '@/components/LayoutWithMenu/ProductMetadataContext'
+import {
+  ExpectedBlockType,
+  LinkField,
+  SbBaseBlockProps,
+  StoryblokAsset,
+} from '@/services/storyblok/storyblok'
 import {
   checkBlockType,
   filterByBlockType,
@@ -34,6 +40,7 @@ import { ButtonBlockProps } from './ButtonBlock'
 type NavItemBlockProps = SbBaseBlockProps<{
   name: string
   link: LinkField
+  pillowImage?: StoryblokAsset
 }>
 
 export const NavItemBlock = ({ blok }: NavItemBlockProps) => {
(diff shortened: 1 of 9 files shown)


robinandeer:
* main
  * **PR #1509** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1509" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1509?utm_source=stack-comment).

apps/store/src/blocks/HeaderBlock.tsx
```diff
@@ -90,10 +97,31 @@ type ProductNavContainerBlockProps = SbBaseBlockProps<{
...
+      productNavItems.push({
+        name: product.displayNameShort,
+        url: product.pageLink,
+        image: product.pillowImage.src,
+      })
+    } else {
+      productNavItems.push({
+        name: item.name,
+        url: getLinkFieldURL(item.link, item.name),
+        image: item.pillowImage?.filename,
```
gustaveen, line 120: What's the use case for the fallback? If we want to link to "coming soon" product pages that are not yet in the API?

apps/store/src/graphql/ProductMetadata.graphql
```diff
@@ -0,0 +1,13 @@
...
+    id
+    name
+    displayNameShort
+    pageLink
+    pillowImage {
+      id
+      alt
+      src
+    }
+  }
```
gustaveen, line 12: Just so I understand what the data comes from :) Is this all Product Meta data from Storyblok or a mix of Storyblok/PCMS/API content? If it's all from Storyblok we could just get it from there instead right?

gustaveen (approved):
Nice work! 🙌
Just had some questions :D

apps/store/src/graphql/ProductMetadata.graphql
robinandeer: This is all product data which is not stored in PCMS

However, I like the idea of a single source of truth. We get this data from the API. We shouldn't care where it's originally stored.

I did do some investigation to get it from the CMS but didn't find a simple way using the Storyblok API so still think this is the simplest.

apps/store/src/blocks/HeaderBlock.tsx
robinandeer: This is to support linking to non-product pages like "home insurance" which represents multiple product pages.

gustaveen: Ah right, makes sense! 🙌

PR #1516, opened 2023-02-02
Update confirmation page to fetch CMS content

robinandeer:
## Describe your changes

Integrate CMS on confirmation page.

Get all content from the CMS.

This is 90% a content page so the source should be the CMS.

This is a first implementation, we will need to update this to support rearranging the content blocks in the future.

### DIFF
--- a/apps/store/src/components/ConfirmationPage/ConfirmationPage.stories.tsx
+++ b/apps/store/src/components/ConfirmationPage/ConfirmationPage.stories.tsx
@@ -1,0 +1,117 @@
+import { ComponentMeta } from '@storybook/react'
+import { CurrencyCode, ExternalInsuranceCancellationOption } from '@/services/apollo/generated'
+import { ConfirmationPage } from './ConfirmationPage'
+
+export default {
+  title: 'ConfirmationPage',
+  component: ConfirmationPage,
+  parameters: {
+    layout: 'fullscreen',
+  },
+} as ComponentMeta<typeof ConfirmationPage>
+
+export const Default = () => {
+  return (
+    <ConfirmationPage
+      globalStory={{
+        ...storyblokStory,
+        content: {
+          header: [],
+          footer: [],
+        },
+      }}
+      currency="SEK"
+      platform="apple"
+      cart={{
+        id: '123',
+        cost: {
+          gross: { amount: 225, currencyCode: CurrencyCode.Sek },
+          net: { amount: 215, currencyCode: CurrencyCode.Sek },
+          discount: { amount: 10, currencyCode: CurrencyCode.Sek },
+        },
+        redeemedCampaigns: [],
+        entries: [
+          {
+            id: '123',
+            startDate: '2023-02-24',
+            priceIntentData: {},
+            variant: {
+              typeOfContract: 'SE_APARTMENT_BRF',
+              displayName: 'Hedvig BRF',
+              product: {
+                name: 'SE_APARTMENT_BRF',
+                displayNameFull: 'Home Insurance Homeowner',
+                pillowImage: {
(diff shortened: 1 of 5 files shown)


robinandeer:
* main
  * **PR #1516** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1516" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1516?utm_source=stack-comment).

apps/store/src/components/ConfirmationPage/ConfirmationPage.tsx
```diff
@@ -19,70 +21,109 @@ export const ConfirmationPage = (props: Props) => {
   const { locale } = useCurrentLocale()
   const { platform, cart, story } = props
 
+  const appStoreButton = platform ? (
```
gustaveen, line 24: Aha, so we do have a platform check 👍

apps/store/src/components/ConfirmationPage/ConfirmationPage.stories.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+export default {
+  title: 'ConfirmationPage',
+  component: ConfirmationPage,
+  parameters: {
+    layout: 'fullscreen',
+  },
+} as ComponentMeta<typeof ConfirmationPage>
+
+export const Default = () => {
```
alebedev, line 13: Is story data just an export from Storyblok or it needs hand-editing?

If it can be exported as is and put here, let's leave a comment about it - would simplify maintenance

apps/store/src/components/ConfirmationPage/FooterSection.tsx
```diff
@@ -0,0 +1,47 @@
...
+  [mq.sm]: {
+    minHeight: '40rem',
+  },
+
+  [mq.lg]: {
+    minHeight: '50rem',
+  },
+})
+
+const FooterImage = styled(Image)({
```
gustaveen, line 37: nit: tricky as always to align these images 🤪 looks like it could be pushed more to the left, but maybe easier to just crop the original image. let's merge and adjust later

alebedev (approved):
Great improvement!

apps/store/src/services/storyblok/storyblok.ts
```diff
@@ -130,6 +130,16 @@ export type ReusableStory = ISbStoryData & {
...
+    title: string
+    subtitle: string
+    footerTitle: string
+    footerSubtitle: string
+    footerImage: StoryblokAsset
+    checklistTitle: string
+    checklistSubtitle: string
+    checklist: string
+    faqTitle: string
+    faqSubtitle: string
```
gustaveen, line 142: Nice and simple 👌

gustaveen (approved):
Awesome! 🎉

apps/store/src/components/ConfirmationPage/FooterSection.tsx
robinandeer: right yeah it doesn't look great on all viewports but the image is also cropped so it really should stick to the right for as long as possible

but this solution is very specific to this exact image which is not so great

apps/store/src/components/ConfirmationPage/ConfirmationPage.stories.tsx
robinandeer: right... yeah I guess it could, perhaps we need some sort of fixtures?

although most of the time I feel like these cases highlight a general need to simplify props - we should transform the data before passing it down to visual components

but this is hard for CMS stuff :P

PR #1523, opened 2023-02-03
Implement shared 12 column grid layout

robinandeer:
## Describe your changes

Implement shared component for the 12 column grid.

Include a content wrapper that makes sure main content is centered and takes up the correct amount of horizontal space.

We are starting to see this pattern across the site. Having a shared component will lead to a more consistent experience.

Didn't want to go too far with this PR. However, let's start to thing about what other utility layout components we might need.

### DIFF
--- a/apps/store/src/components/CartPage/CartPage.tsx
+++ b/apps/store/src/components/CartPage/CartPage.tsx
@@ -9,6 +9,7 @@ import { CampaignCodeList } from '@/components/CartInventory/CampaignCodeList'
 import { CartEntryItem } from '@/components/CartInventory/CartEntryItem'
 import { CartEntryList } from '@/components/CartInventory/CartEntryList'
 import { CostSummary } from '@/components/CartInventory/CostSummary'
+import { GridLayout } from '@/components/GridLayout/GridLayout'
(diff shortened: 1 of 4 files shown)


robinandeer:
* main
  * **PR #1523** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1523" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1523?utm_source=stack-comment).

gustaveen (approved):
Nice! 🙌 The`Root` we can use for all grid layouts but I wonder if it's easier in cases of multiple item layouts to keep it local to the component rather than adding props to `GridLayout.Content`

robinandeer:
@gustaveen yes `Root` is always used :)

`Content` was meant to only be used in cases where you want to use the centered layout which takes up ⅓ of the screen on larger viewports. Perhaps it's not the best name though. Still thinks it's something worth sharing across pages etc.

Don't see a big need to add props to `Content` -- we would rather identify another common layout pattern and create an additional `Root`-child component to use instead -- what do you think about that?

PR #1528, opened 2023-02-03
GRW-2165 / Navigation-design-comments

amandaflod:
## Describe your changes

Tweak paddings on navigation menu after feedback from Design. 

Also improved ability to hover across to secondary menu without the menu closing. 

Screenshot 2023-02-03 at 14 31 44

### DIFF
--- a/apps/store/src/components/Header/HeaderStyles.tsx
+++ b/apps/store/src/components/Header/HeaderStyles.tsx
@@ -75,13 +75,15 @@ export const NavigationMenuProductItem = styled(NavigationMenuPrimitive.Item)()
 export const NavigationMenuPrimitiveContent = styled(NavigationMenuPrimitive.Content)({
   [mq.lg]: {
     position: 'absolute',
-    top: `calc(${MENU_BAR_HEIGHT_DESKTOP} + ${theme.space.xs})`,
+    paddingTop: `calc(${theme.space.sm} + ${theme.space.xs})`,
     left: '50%',
     transform: 'translate(-50%, 0)',
   },
 })
 
 export const NavigationMenuListWrapper = styled.div({
+  paddingBottom: theme.space.lg,
+
   [mq.lg]: {
     backgroundColor: theme.colors.light,
     boxShadow: '0px 1px 2px rgba(0, 0, 0, 0.15)',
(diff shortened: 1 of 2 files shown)


apps/store/src/components/Header/HeaderStyles.tsx
```diff
@@ -105,12 +107,13 @@ export const NavigationPrimaryList = styled(NavigationMenuPrimitive.List)({
     flexDirection: 'row',
     alignItems: 'center',
     height: MENU_BAR_HEIGHT_DESKTOP,
+    padding: 0,
```
guilhermespopolin, line 110: nit + question:
Should we be using the following instead? 🤔 
 
```suggestion
    padding: theme.space.none,
```

apps/store/src/components/Header/HeaderStyles.tsx
```diff
@@ -75,13 +75,15 @@ export const NavigationMenuProductItem = styled(NavigationMenuPrimitive.Item)()
 export const NavigationMenuPrimitiveContent = styled(NavigationMenuPrimitive.Content)({
   [mq.lg]: {
     position: 'absolute',
-    top: `calc(${MENU_BAR_HEIGHT_DESKTOP} + ${theme.space.xs})`,
+    paddingTop: `calc(${theme.space.sm} + ${theme.space.xs})`,
```
guilhermespopolin, line 78: extremely nit:
If I understood correctly we needed `1.25rem ~20px` here and since we don't have that specific value in our design system you've achieved it by adding values that we have. 

I think one of the benefits of having a design system with pre defined _spacing_ values is that we can space things in a consistent way while simplifying our lives by reducing the amount of possible values for spacing. I know that we'll have some exceptions but if we add them like this we'll be getting _magic numbers_ for spacing again.

With that said I can see 3 possible solutions:
1. Identify the real necessity for that value and add it into our `theme` in case we actually need it;
2. Use an approximate value from our design system and move on with our lives without a _pixel perfect_ design on some cases: I believe the closest value would be `theme.space.lg --> 1.5rem`
3. Consider this an edge case and leave it as it is.

In general I'd choose option 2. Let me know what you think 😄

guilhermespopolin (approved):
Looking good
Added some comments. All of them are _nits_ tho :)

apps/store/src/components/Header/HeaderStyles.tsx
amandaflod: In this case we wrote it like this to signify that the padding is made up of two different components, the padding of the header (sm) and the space between the header and the sub-menu, if that makes sense? :P

apps/store/src/components/Header/HeaderStyles.tsx
amandaflod: I didn't know we had that one, will update!

PR #1545, opened 2023-02-06
Integrate new renewal date API in cancellation form

robinandeer:
## Describe your changes

Integrate new API for renewal date in cancellation form.

Update date input to handle rendering without a date.

Update chevron icon to match figma.

We should use the renewal date for cases where the user is selecting the expiry date of their current insurance.

We need to make some API updates so "bank signering" options are based on these dates rather than start date.


robinandeer:
* main
  * **PR #1545** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1545" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1545?utm_source=stack-comment).

apps/store/src/components/ProductPage/PurchaseForm/OfferPresenter.tsx
```diff
@@ -23,6 +23,7 @@ import { CancellationForm, CancellationOption } from './CancellationForm/Cancell
 import * as ComparisonTable from './ComparisonTable/ComparisonTable'
```
alebedev, line 23: Nit: This module keeps growing, is it time to split?

alebedev (approved):
Looks good

apps/store/src/components/ProductPage/PurchaseForm/OfferPresenter.tsx
robinandeer: Yes indeed -- just wanna let it grow until it stabilizes and then refactor

PR #1547, opened 2023-02-06
GRW-2120 - feat(VideoListBlock): play only one video at a time

guilhermespopolin:
## Describe your changes

* Update `Slideshow` so it can get a ref;
* Add an event handler for `play` event to each video of `VideoListBlock` that make sure only one video gets played at a time.

Requested by design.

## Jira issue(s): [GRW-2120](https://hedvig.atlassian.net/browse/GRW-2120)

https://user-images.githubusercontent.com/19200662/216939688-cc8f2800-2bd4-4c3d-8783-a97cce962385.mov

[GRW-2120]: https://hedvig.atlassian.net/browse/GRW-2120?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/store/src/blocks/VideoListBlock.tsx
+++ b/apps/store/src/blocks/VideoListBlock.tsx
@@ -5,13 +5,35 @@ import { Slideshow } from '@/components/Slideshow/Slideshow'
 import { ExpectedBlockType, SbBaseBlockProps } from '@/services/storyblok/storyblok'
 import { VideoBlock, VideoBlockProps } from './VideoBlock'
 
+const playVideoExclusively = (
+  videoList: Array<HTMLVideoElement>,
+  videoToBePlayed: HTMLVideoElement,
+) => {
+  videoList.forEach((videoNode) => {
+    if (videoNode === videoToBePlayed) return
+    if (videoNode.played.length > 0 && !videoNode.paused) {
+      videoNode.pause()
+    }
+  })
+}
+
 type VideoListBlockProps = SbBaseBlockProps<{
   videos: ExpectedBlockType<VideoBlockProps>
 }>
 
 export const VideoListBlock = ({ blok }: VideoListBlockProps) => {
   return (
-    <Slideshow alignment="center" {...storyblokEditable(blok)}>
+    <Slideshow
+      ref={(node) => {
+        const videos = Array.from(node?.querySelectorAll('video') ?? [])
+
+        videos.forEach((video) => {
+          video.addEventListener('play', () => playVideoExclusively(videos, video))
+        })
+      }}
+      alignment="center"
+      {...storyblokEditable(blok)}
+    >
       {blok.videos.map((nestedVideoBlock) => (
         <StyledVideoBlock key={nestedVideoBlock._uid} blok={nestedVideoBlock} />
       ))}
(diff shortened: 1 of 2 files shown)


guilhermespopolin:
* main
  * **PR #1547** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1547" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1547?utm_source=stack-comment).

apps/store/src/blocks/VideoListBlock.tsx
```diff
@@ -5,13 +5,35 @@ import { Slideshow } from '@/components/Slideshow/Slideshow'
...
 type VideoListBlockProps = SbBaseBlockProps<{
   videos: ExpectedBlockType<VideoBlockProps>
 }>
 
 export const VideoListBlock = ({ blok }: VideoListBlockProps) => {
   return (
-    <Slideshow alignment="center" {...storyblokEditable(blok)}>
+    <Slideshow
+      ref={(node) => {
+        const videos = Array.from(node?.querySelectorAll('video') ?? [])
```
robinandeer, line 28: I wonder if there's a more react-y way to accomplish this 🤔 

maybe we can mob program it later?

robinandeer (approved):
Nice solution! Let's see if we can come together and make it more React-y later

gustaveen (approved):
Great one! Works like a charm 🙌

apps/store/src/blocks/VideoListBlock.tsx
guilhermespopolin: Yes, please. The more react-y way to accomplish this that I had came up with looked more complicated. But while I was writing you this I've came up with a another one that might be promising:

* Update `VideoBlock` so it can receive a `ref` that will then be passed to `Video` component. `VideoBlock` would also need to be updated so it can take in more `Video` related props as `onPlay` event handler;
* `Video` component already keeps a `ref` for the `<video>` element so we can control it (play/pause) so we'd need to update `Video` component by getting both `ref`s merged (`ui/merge-refs`);
* Update `VideoListBlock` so we keep a `ref` for each video. Since the number of videos is dynamic I though about storing that list as a ref. Not sure if there's a better way  🤔:
```
export const VideoListBlock = ({ blok }: VideoListBlockProps) => {
  const videos = useRef<HTMLVideoElement>([])
  
  const  playVideoExclusively = useCallback((event) => {
    // need to check this but I assume I cant the target video from `event` object here
    const videosThatShouldBePaused = videos.current.filter(video => video !== event.target && video.played.length > 0 && !video.paused)
    videosThatShouldBePaused.forEach(video => video.pause())
  }, [])
  
  return (
    <Slideshow alignment="center" {...storyblokEditable(blok)}>
      {blok.videos.map((nestedVideoBlock) => (
        <StyledVideoBlock ref={(node) => node && videos.current.push(node)} key={nestedVideoBlock._uid} blok={nestedVideoBlock} onPlay={} />
      ))}
    </Slideshow>
  )
}
```

What do you think?!

robinandeer: haha yeah using refs is also advanced -- I would say merge this now since it's working and we can look into it again later

PR #1556, opened 2023-02-06
GRW-2161 / Cart notification mobile design

amandaflod:
## Describe your changes

To match design, I added a prop `isToast` to Dialog that lets the modal's overlay start below the header in Mobile. 

Screenshot 2023-02-06 at 16 07 43

[GRW-2161]: https://hedvig.atlassian.net/browse/GRW-2161?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/store/src/components/CartNotification/CartToast.tsx
```diff
@@ -38,7 +38,7 @@ export const CartNotificationContent = ({ name, price, startDate, onClose }: Pro
   const { t } = useTranslation('purchase-form')
 
   return (
-    <DialogContent onClose={onClose}>
+    <DialogContent onClose={onClose} isToast>
```
robinandeer, line 41: nit: learned in the "Joy of React" course that this syntax is not favoured

We should probably be more explicit and write `isToast={true}`

packages/ui/src/components/Dialog/Dialog.tsx
```diff
@@ -33,6 +34,13 @@ const StyledOverlay = styled(DialogPrimitive.Overlay, {
   }),
 }))
 
+const ToastOverlay = styled(StyledOverlay)({
+  top: '3rem',
```
robinandeer, line 38: this will be quite hard to maintain... we would really like a reference to the menu bar height, don't you agree?

is there some way to handle this logic inside `store` app instead? I would hope we can keep the dialog component as generic as possible

packages/ui/src/components/Dialog/Dialog.tsx
```diff
@@ -33,6 +34,13 @@ const StyledOverlay = styled(DialogPrimitive.Overlay, {
...
+  [mq.lg]: {
+    top: 0,
```
robinandeer, line 40: How does this look on desktop? Won't it overlap with the menu?

robinandeer (commented):
Left some comments -- is there someway we can add this functionality only to "store" app and not "UI" package?

apps/store/src/components/CartNotification/CartToast.tsx
amandaflod: Ohh right! Will change that :))

packages/ui/src/components/Dialog/Dialog.tsx
amandaflod: Yes, I was thinking the same but the problem is that the overlay is just inside a Portal element so it's quite hard to reach 🤔 Maybe I should instead handle it the same way that we do the frosted overlay prop? Or maybe should I create a custom dialog (toast) component from radix? WDYT?

packages/ui/src/components/Dialog/Dialog.tsx
amandaflod: It should cover the menu in desktop! :)

packages/ui/src/components/Dialog/Dialog.tsx
guilhermespopolin: I guess the issue is that our `ui/Dialog` is currently not easy to extend. Almost all of the "markup" necessary for building a Dialog is composed inside `Content` component which renders the `Portal`, `Overlay` and `Content`. It's a tradeoff: It's easy to create a Dialog but it's hard to customise it like, for example, change some styles of the Overlay.

I've tried to export `StyledOverlay` from `ui/Dialog` and change its style under `CartToast` like so:

```
const DialogContent = styled(Dialog.Content)({
  marginTop: MENU_BAR_HEIGHT_MOBILE,
  [mq.lg]: {
    marginTop: MENU_BAR_HEIGHT_DESKTOP,
    display: 'flex',
    justifyContent: 'flex-end',
  },

  [`& ${Dialog.StyledOverlay}`]: {
    top: MENU_BAR_HEIGHT_MOBILE,
    [mq.md]: {
      top: MENU_BAR_HEIGHT_DESKTOP,
    },
  },
})
```

But it doesn't work because I can't came up with a CSS selector that targets the overlay due the DOM structure on which it is being rendered:

```
  return (
    <DialogPrimitive.Portal>
      <StyledOverlay frosted={frostedOverlay} />
      <StyledContentWrapper>
        <DialogPrimitive.Content
          className={className}
          onEscapeKeyDown={handleClose}
          onInteractOutside={handleClose}
        >
          {children}
        </DialogPrimitive.Content>
      </StyledContentWrapper>
    </DialogPrimitive.Portal>
  )
```

`~` and `+` sibling selectors also doesn't work 😬 
Should we construct `CartToast` from Radix Dialog instead? 🤔

amandaflod: Thank you for the input!! 🙏 Yes, I think building it directly from radix is the way we have to go about it 🤔

apps/store/src/components/CartNotification/CartToast.tsx
```diff
@@ -64,6 +73,29 @@ const DialogContent = styled(Dialog.Content)({
...
+const overlayShow = keyframes({
+  '0%': { opacity: 0 },
+  '100%': { opacity: 1 },
+})
+
+const StyledOverlay = styled(DialogPrimitive.Overlay)({
+  backgroundColor: 'rgba(0, 0, 0, 0.15)',
+  position: 'fixed',
+  inset: 0,
+  top: '3rem',
```
robinandeer, line 85: what is this value? Can we use `MENU_BAR_HEIGHT_MOBILE`?

robinandeer (approved):
Nice, I like this solution much better :D

And looks good in the preview app

apps/store/src/components/CartNotification/CartToast.tsx
amandaflod: Good catch, will update to that!

PR #1557, opened 2023-02-06
GRW-2120 refact(VideoBlock): play one video at a time

guilhermespopolin:
## Describe your changes

This is a working draft for an alternative more _react-y_ version of the _Play one video at a time_ originally implemented [here](https://github.com/HedvigInsurance/racoon/pull/1547).

Basically the idea here was:

* Update `VideoBlock` so it can receive a `ref` that will then be passed to `Video` component. `VideoBlock` would also need to be updated so it can take in more `Video` related props as `onPlay` event handler;
* `Video` component already keeps a `ref` for the `<video>` element so we can control it (play/pause) so we'd need to update `Video` component by getting both `ref`s merged (`ui/merge-refs`);
* Update `VideoListBlock` so we keep a `ref` for each video. Since the number of videos is dynamic I though about storing that list as a ref. Not sure if there's a better way  🤔:

We can use this as an starting point of even discard it and start from scratch. No strings attached.

It's just some exploration based on some comments I got from the other PR.

## Jira issue(s): [GRW-2120](https://hedvig.atlassian.net/browse/GRW-2120)

[GRW-2120]: https://hedvig.atlassian.net/browse/GRW-2120?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

FYI @robinandeer

### DIFF
--- a/apps/store/src/blocks/VideoBlock.tsx
+++ b/apps/store/src/blocks/VideoBlock.tsx
@@ -1,37 +1,56 @@
 import styled from '@emotion/styled'
+import { forwardRef } from 'react'
 import { mq, theme } from 'ui'
 import { Video, VideoProps } from '@/components/Video/Video'
 import { SbBaseBlockProps, StoryblokAsset } from '@/services/storyblok/storyblok'
 
-export type VideoBlockProps = SbBaseBlockProps<
-  {
-    video: StoryblokAsset
-    poster?: StoryblokAsset
-  } & Pick<
+type SbRelatedProps = {
+  video: StoryblokAsset
+  poster?: StoryblokAsset
+} & Pick<
+  VideoProps,
+  | 'autoPlay'
+  | 'aspectRatioLandscape'
+  | 'aspectRatioPortrait'
+  | 'maxHeightLandscape'
+  | 'maxHeightPortrait'
+>
+
+export type VideoBlockProps = SbBaseBlockProps<SbRelatedProps> &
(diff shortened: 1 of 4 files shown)


guilhermespopolin:
* main
  * **PR #1557** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1557" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1557?utm_source=stack-comment).

apps/store/src/blocks/VideoBlock.tsx
```diff
@@ -1,37 +1,56 @@
...
+          maxHeightLandscape={blok.maxHeightLandscape}
+          maxHeightPortrait={blok.maxHeightPortrait}
+          {...delegated}
+        />
+      </Wrapper>
+    )
+  },
+)
+VideoBlock.displayName = 'VideoBlock'
 VideoBlock.blockName = 'videoBlock'
```
guilhermespopolin, line 54: Here's something that I couldn't solve:
I'm getting an TS error because I can't set that `blockName` field since the type returned by `forwardRef` doesn't allow it 😢

guilhermespopolin:
It was decided to close this one and use as reference when we cycle back on this during a maintenance day.

PR #1560, opened 2023-02-07
Update text contnent block

gustaveen:
## Describe your changes
New block that we will use for text content
The block can render `Heading`, `HeadingLabel` and `Text`.

`Content` block is still left but this will replace that block once migrated

https://user-images.githubusercontent.com/6661511/217306036-b6fcd659-84cc-48ea-b8ca-87a6f526ed33.mov

Next steps: Add settings for controlling column layout/width and spacing between text.
The plan is to align the content to the 12 column grid.

To work with text heavy pages we need more options to render text

### DIFF
--- a/apps/store/src/blocks/HeadingBlock.tsx
+++ b/apps/store/src/blocks/HeadingBlock.tsx
@@ -1,6 +1,6 @@
 import styled from '@emotion/styled'
 import { storyblokEditable } from '@storyblok/react'
-import { Heading, HeadingProps, PossibleHeadingVariant, theme } from 'ui'
+import { ConditionalWrapper, Heading, HeadingProps, PossibleHeadingVariant, theme } from 'ui'
 import { SbBaseBlockProps } from '@/services/storyblok/storyblok'
 
 export const Wrapper = styled.div({
@@ -15,11 +15,15 @@ export type HeadingBlockProps = SbBaseBlockProps<{
   variant?: PossibleHeadingVariant
   variantDesktop?: PossibleHeadingVariant
   textAlignment?: HeadingProps['align']
+  nested?: boolean
 }>
 
 export const HeadingBlock = ({ blok }: HeadingBlockProps) => {
   return (
-    <Wrapper>
+    <ConditionalWrapper
+      condition={!blok.nested}
+      wrapWith={(children) => <Wrapper>{children}</Wrapper>}
+    >
       <Heading
         as={blok.as}
         variant={{ _: blok.variant ?? 'standard.32', md: blok.variantDesktop ?? 'standard.40' }}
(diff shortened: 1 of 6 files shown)


gustaveen:
* main
  * **PR #1560** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1560" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1560?utm_source=stack-comment).

apps/store/src/blocks/HeadingBlock.tsx
```diff
@@ -15,11 +15,15 @@ export type HeadingBlockProps = SbBaseBlockProps<{
...
+  nested?: boolean
 }>
 
 export const HeadingBlock = ({ blok }: HeadingBlockProps) => {
   return (
-    <Wrapper>
+    <ConditionalWrapper
+      condition={!blok.nested}
+      wrapWith={(children) => <Wrapper>{children}</Wrapper>}
+    >
```
gustaveen, line 26: When using heading as a nested block it should render without wrapper with padding. This is a pattern we can use for primitive blocks that can be both nested or standalone

apps/store/src/blocks/HeadingBlock.tsx
```diff
@@ -15,11 +15,15 @@ export type HeadingBlockProps = SbBaseBlockProps<{
...
   variantDesktop?: PossibleHeadingVariant
   textAlignment?: HeadingProps['align']
+  nested?: boolean
 }>
 
 export const HeadingBlock = ({ blok }: HeadingBlockProps) => {
   return (
-    <Wrapper>
+    <ConditionalWrapper
+      condition={!blok.nested}
```
robinandeer, line 24: is `blok.nested` provided by Storyblok or do we have to add it to all blocks?

apps/store/src/blocks/TextContentBlock.tsx
```diff
@@ -0,0 +1,50 @@
...
+  )
+}
+TextContentBlock.blockName = 'textContent'
+
+const Wrapper = styled(GridLayout.Root)({
+  paddingLeft: theme.space.md,
+  paddingRight: theme.space.md,
+})
+
+const TextWrapper = styled(Space)<{ alignment: Alignment }>(({ alignment }) => ({
```
robinandeer, line 39: Let's see if we should co-locate this later with the `GridLayout` component

robinandeer (approved):
Looks really good to me :D

apps/store/src/blocks/TextContentBlock.tsx
gustaveen: Yes, let's see if it will be reusable or more specific for this block 👍

apps/store/src/blocks/HeadingBlock.tsx
gustaveen: Nope, not provided by Storyblok. I thought we could use this to determine if a block is nested and hence shouldn't have paddings, but pass the prop in React. But maybe better place it outside of the `blok` object

PR #1564, opened 2023-02-07
TierSelector: add product variant descriptions

robinandeer:
## Describe your changes

`TierSelector`: add product variant descriptions

This should ideally come from the API but that's not possible in the short term. Added comment to remember.


robinandeer:
* main
  * **PR #1564** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1564" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1564?utm_source=stack-comment).

apps/store/src/components/TierSelector/TierSelector.tsx
```diff
@@ -43,7 +43,7 @@ const TierItem = ({
           <PriceText isSelected={isSelected}>{price}</PriceText>
         </TitleItem>
       </TitleContainer>
-      <SecondaryText>{description}</SecondaryText>
+      {description && <SecondaryText>{description}</SecondaryText>}
```
alebedev, line 46: Nit: How would suggested item + description looks like?  Would there be any overflow / overlap / other visual artifacts?

alebedev (approved):
Nice!

apps/store/src/components/TierSelector/TierSelector.tsx
robinandeer: Hmm no don't think so but good point -- the design assumes we always have description so we are a little on our own now

Also "suggested text" is an Insurely feature :P

PR #1558, opened 2023-02-07
Fix ApolloClient authorization header during SSR

alebedev:
## Describe your changes

SSR ApolloClient never sent authentication header, fixed this

## Checklist before requesting a review

- [ ] I have performed a self-review of my code

PR #1554, opened 2023-02-06
Remove authentication step from cart page

alebedev:
## Describe your changes

- Remove returning member authentication before proceeding to checkout
- Implement signing by returning member (get token -> authenticate session -> sign).  Very basic support, UI would be updated in following PRs

## Justify why they are needed

Allow returning members to sign if they skipped login in price calculator

## Checklist before requesting a review

- [ ] I have performed a self-review of my code

PR #1549, opened 2023-02-06
Improve BankID login design

alebedev:
## Describe your changes

- Expose BankID login status through API
- Reflect every BankId status in UI
- Apply designs

## Justify why they are needed

Intermediate step, to be followed by more improvements

## Checklist before requesting a review

- [ ] I have performed a self-review of my code

PR #1529, opened 2023-02-03
Improve returning member authentication UI

alebedev:
## Describe your changes

Incremental improvements for returning member authentication
- Apply design to BankID login prompt
- Localize affected components

## Justify why they are needed

Make returning member flow look good. To be followed by other PRs with the same goal

## Checklist before requesting a review

- [ ] I have performed a self-review of my code

PR #1565, opened 2023-02-07
Extract BankId login from SsnSeSection into app-wide service available through context

alebedev:
## Describe your changes

- Create BankIdContext to handle authentication as app-global concern
- Add global BankIdDialog with authentication UI

- Ensure BankId auth is always on top of any modal dialogs
- Prepare to handle signing the same way


alebedev:
* main
  * **PR #1525** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1525" target="_blank">Graphite</a> 
    * **PR #1529** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1529" target="_blank">Graphite</a> 
      * **PR #1558** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1558" target="_blank">Graphite</a> 
        * **PR #1549** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1549" target="_blank">Graphite</a> 
          * **PR #1554** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1554" target="_blank">Graphite</a> 
            * **PR #1565** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1565" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1565?utm_source=stack-comment).

robinandeer (commented):
Interesting approach -- I can see the need for reuse here

What I'm concerned about it to create a global concept out of something that we know is a country-specific feature

I would like to discuss whether we should abstract away the country-specific implementations/names if we make this switch. Otherwise it seems like we will sprinkle a lot of "sweden" logic all over the code base which is what we want to avoid

apps/store/src/services/bankId/BankIdContext.tsx
```diff
@@ -0,0 +1,96 @@
...
+export const BankIdContextProvider = ({ children }: PropsWithChildren) => {
+  const { shopSession } = useShopSession()
+  const [currentOperation, setCurrentOperation] =
+    useState<BankIdContextValue['currentOperation']>(null)
+
+  // TODO: Expose and handle errors
+  const [authenticateShopSession] = useShopSessionAuthenticateMutation()
+
+  const showLoginPrompt = useCallback(() => {
+    console.log('init login')
```
robinandeer, line 34: do you mean to log with datadog?

alebedev: Yes, plan to clear this out in the next PR

apps/store/src/services/bankId/BankIdContext.tsx
```diff
@@ -0,0 +1,102 @@
...
+export const BankIdContextProvider = ({ children }: PropsWithChildren) => {
+  const { shopSession } = useShopSession()
+  const [currentOperation, setCurrentOperation] =
+    useState<BankIdContextValue['currentOperation']>(null)
+
+  // TODO: Expose and handle errors
+  const [authenticateShopSession] = useShopSessionAuthenticateMutation()
+
+  const showLoginPrompt: BankIdContextValue['showLoginPrompt'] = useCallback(({ onCompleted }) => {
+    console.log('show login', { onCompleted })
```
robinandeer, line 37: Datadog?

apps/store/src/services/bankId/BankIdContext.tsx
```diff
@@ -0,0 +1,102 @@
...
+    setCurrentOperation({
```
robinandeer, line 38: Perhaps we should experiment a little with `useReducer` in these more complicated cases -- future improvement

robinandeer (approved):
A lot to take it but I believe it's moving us in the right direction ⭕

apps/store/src/services/bankId/BankIdContext.tsx
alebedev: Next PR 😄

PR #1569, opened 2023-02-08
GRW-2185 - styles(ImageTextBlock): updates responsive styles

guilhermespopolin:
## Describe your changes

Updated styles as requested by design

## Jira issue(s): [GRW-2185](https://hedvig.atlassian.net/browse/GRW-2185)

Screenshot 2023-02-09 at 10 01 15

[GRW-2185]: https://hedvig.atlassian.net/browse/GRW-2185?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/store/src/blocks/ButtonBlock.tsx
+++ b/apps/store/src/blocks/ButtonBlock.tsx
@@ -5,6 +5,7 @@ import { Button, theme } from 'ui'
 import { ButtonNextLink } from '@/components/ButtonNextLink'
 import { LinkField, SbBaseBlockProps } from '@/services/storyblok/storyblok'
 import { getLinkFieldURL } from '@/services/storyblok/Storyblok.helpers'
+import { VerticalBodyWrapper, FluidBodyWrapper } from './ImageTextBlock'
 
 export type ButtonBlockProps = SbBaseBlockProps<{
   text: string
@@ -28,9 +29,14 @@ export const ButtonBlock = ({ blok }: ButtonBlockProps) => {
 }
 ButtonBlock.blockName = 'button'
 
-export const Wrapper = styled.div({
+const Wrapper = styled.div({
   display: 'flex',
   justifyContent: 'center',
   paddingLeft: theme.space.md,
   paddingRight: theme.space.md,
+
+  [`${VerticalBodyWrapper} &, ${FluidBodyWrapper} &`]: {
+    display: 'inline-block',
+    padding: 0,
+  },
 })
--- a/apps/store/src/blocks/ImageTextBlock.tsx
+++ b/apps/store/src/blocks/ImageTextBlock.tsx
@@ -2,8 +2,6 @@ import styled from '@emotion/styled'
 import { SbBlokData, StoryblokComponent } from '@storyblok/react'
 import { default as NextImage } from 'next/image'
 import { mq, theme } from 'ui'
-import { Wrapper as ButtonBlockWrapper } from '@/blocks/ButtonBlock'
-import { Wrapper as HeadingBlockWrapper } from '@/blocks/HeadingBlock'
 import { GridLayout } from '@/components/GridLayout/GridLayout'
 import { SbBaseBlockProps, StoryblokAsset } from '@/services/storyblok/storyblok'
 
@@ -71,18 +69,10 @@ const VerticalImageTextBlockWrapper = styled.div({
   },
 })
 
-const VerticalBodyWrapper = styled.div({
+export const VerticalBodyWrapper = styled.div({
   padding: theme.space.xs,
   [mq.md]: {
     paddingBlock: theme.space.md,
-  },
-
-  [`${HeadingBlockWrapper}`]: {
-    padding: 0,
-  },
-  [`${ButtonBlockWrapper}`]: {
-    display: 'inline-block',
-    padding: 0,
   },
 })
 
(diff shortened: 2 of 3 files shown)


guilhermespopolin:
* main
  * **PR #1569** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1569" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1569?utm_source=stack-comment).

apps/store/src/blocks/ImageTextBlock.tsx
```diff
@@ -20,86 +21,142 @@ type ImageTextBlockProps = SbBaseBlockProps<{
...
+          image={blok.image}
+          body={blok.body}
+          textAlignment={blok.textAlignment}
+          imagePlacement={blok.imagePlacement}
+        />
+      )
+    case 'vertical':
+      return <VerticalImageTextBlock image={blok.image} body={blok.body} />
+    default:
+      console.warn(`orientation type '${orientation}' is not valid`)
```
robinandeer, line 42: Hmm perhaps we should discuss logging during rendering -- it's technically a side effect so doesn't belong there I think

Also this only shows up in local console instead of datadog -- but ok to leave for now

apps/store/src/blocks/ImageTextBlock.tsx
```diff
@@ -20,86 +21,142 @@ type ImageTextBlockProps = SbBaseBlockProps<{
...
-  [mq.lg]: {
-    paddingInline: theme.space.md,
+  [`${HeadingBlockWrapper}`]: {
+    padding: 0,
   },
-}))
+  [`${ButtonBlockWrapper}`]: {
+    display: 'inline-block',
+    padding: 0,
+  },
```
robinandeer, line 86: You might want to consider reversing this:

```
const HeadingBlockWrapper = styled.div({
  [`${VerticalBodyWrapper} &`]: {
    padding: 0,
  }
})
```

That way we collect all styles for a styled component in once place 🙂 

But feel free to leave as is for now -- we can discuss this more in the team

apps/store/src/blocks/ImageTextBlock.tsx
```diff
@@ -20,86 +21,142 @@ type ImageTextBlockProps = SbBaseBlockProps<{
...
+  paddingBlock: theme.space.md,
+  paddingInline: theme.space.xs,
+  gridColumn: '1 / -1',
+  [mq.lg]: {
+    display: 'flex',
+    flexDirection: 'column',
+    justifyContent: getTextAlignmentStyles(textAlignment),
+    gridColumn: 'span 6',
+    order: imagePlacement === 'right' ? -1 : 'initial',
+    maxWidth: '37.5rem',
```
robinandeer, line 155: do we need grid-column and max-width?

robinandeer (approved):
Looks good I think

Left some comments but we can discuss in team more about it

apps/store/src/blocks/ImageTextBlock.tsx
guilhermespopolin: In this case yes because otherwise it would be very hard to constrain the text to the maximum on 600px:

* If I only set `gridColumn` it would be a guessing game the amount of columns I'd need to take to match that requirement. Actually, I guess it would be impossible since the columns don't have fixed dimensions.
* If I only set `maxWidth` by default the image will occupy another row. I guess I could tweak the grid to avoid that but at the end `gridColumn` + `maxWidth` seems more declarative.

Screenshot 2023-02-09 at 11 05 20

apps/store/src/blocks/ImageTextBlock.tsx
guilhermespopolin: No that's ok. I like try new things. Gonna give it a try.

guilhermespopolin: Have updated the code with the style changes.

apps/store/src/blocks/ImageTextBlock.tsx
gustaveen: We discussed this yesterday in in some cases where we have texts, we will most likely need to set some max width to make the text readable in terms of row lengths 🙂

apps/store/src/blocks/ImageTextBlock.tsx
gustaveen: Can't we just set a default option in Storyblok? Or always fallback to on of the options instead? :) @guilhermespopolin

apps/store/src/blocks/ButtonBlock.tsx
```diff
@@ -28,9 +29,14 @@ export const ButtonBlock = ({ blok }: ButtonBlockProps) => {
...
+const Wrapper = styled.div({
   display: 'flex',
   justifyContent: 'center',
   paddingLeft: theme.space.md,
   paddingRight: theme.space.md,
+
+  [`${VerticalBodyWrapper} &, ${FluidBodyWrapper} &`]: {
+    display: 'inline-block',
+    padding: 0,
+  },
```
gustaveen, line 41: If this would be the case in more places, let's add a setting for it in Sotryblok

gustaveen (approved):
great work! 🙌

apps/store/src/blocks/ImageTextBlock.tsx
guilhermespopolin: Yes, both are in place. Just added this here to cover the scenario when, for some reason, the value provided by storyblok is not covered in the code.

PR #1603, opened 2023-02-09
Accordion Animations

KarnellSchultz:
### [Easings.co](https://easings.co/)

Added open/close keyframe animations to the accordion.

Tried several ways to do this with some issues around the padding the `<Content />` component had. Decided to remove the padding since the block itself has padding. However, going forward I'd like to move the padding into the accordion rather than having it on the block.

[Before demo fixes](https://www.notion.so/hedviginsurance/Before-Demo-Fixes-0f8755117f6e4092b16ae0743dc7bcec?pvs=4)

### DIFF
--- a/apps/store/src/components/Accordion/Accordion.tsx
+++ b/apps/store/src/components/Accordion/Accordion.tsx
@@ -89,10 +89,10 @@ export const Content = styled(AccordionPrimitives.Content)({
   overflow: 'hidden',
 
   '[data-state=open] &': {
-    animation: `${slideDown} 640ms cubic-bezier(0.65,0.05,0.36,1)`,
+    animation: `${slideDown} 400ms cubic-bezier(0.65,0.05,0.36,1)`,
   },
 
   '[data-state=closed] &': {
-    animation: `${slideUp} 640ms cubic-bezier(0.65,0.05,0.36,1)`,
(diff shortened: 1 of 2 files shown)


apps/store/src/components/Accordion/Accordion.tsx
```diff
@@ -64,9 +65,34 @@ const CloseIcon = styled(MinusIcon)({
   '[data-state=open] &': { display: 'block' },
 })
 
+const slideDown = keyframes({
+  from: {
+    height: 0,
+  },
+  to: {
+    height: 'var(--radix-accordion-content-height)',
```
gustaveen, line 73: Where is this variable defined?

gustaveen (approved):
Nice, very smoooth! 🙌

Maybe it can speed up a bit but I'll let Zak and Wille be the judge of that :D

apps/store/src/components/Accordion/Accordion.tsx
KarnellSchultz: it's coming from [radix](https://www.radix-ui.com/docs/primitives/components/accordion) 

Should I drop a comment on that?

KarnellSchultz:
slack convo: https://hedviginsurance.slack.com/archives/C041SD67G82/p1675955904078479

PR #1612, opened 2023-02-10
WIP: Make bankId login request cancellable by returning Observable

alebedev:
## Describe your changes

Allow to cancel BankID login.  This is an API experiment, will continue with Sign cancellation if we're happy with the approach


alebedev:
* main
  * **PR #1612** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1612" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1612?utm_source=stack-comment).

apps/store/src/services/bankId/bankIdReducer.ts
```diff
@@ -64,13 +64,24 @@ export const bankIdReducer = (
           ...action.options!,
         },
       }
-    case 'success':
+    case 'success': {
+      state.currentOperation?.onSuccess()
```
alebedev, line 68: This is bad practice, I'll try to refactor and move side effects out of reducer

robinandeer (commented):
Okey this looks interesting! Learning something new :)

I have two questions/thoughts:

1. Wouldn't it be possible to use `AbortController` to accomplish the same thing? That's a built-in API so worth considering I think. More here: https://stackoverflow.com/a/59999584.

2. I would consider this code slightly more advanced/harder to understand for people without experience with Observables. Should we consider this an isolated case or do we wanna aim at using it more across the code base? How is it to debug observables?

apps/store/src/services/authApi/login.ts
```diff
@@ -1,18 +1,40 @@
...
+          pollTimeoutId = window.setTimeout(() => poll(statusUrl), 1000)
+        }
+      } catch (error) {
+        subscriber.error(error)
+      }
+    }
+
+    memberLoginCreateSE(ssn)
+      .then(({ statusUrl }) => poll(AuthEndpoint.LOGIN_STATUS(statusUrl)))
+      .catch((err) => subscriber.error(err))
```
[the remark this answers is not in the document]
alebedev: Just ergonomics, `Observable` constructor does not accept async functions, so I'd have to wrap it

apps/store/src/services/authApi/login.ts
```diff
@@ -1,18 +1,40 @@
...
 enum MemberLoginMethod {
   SE_BANKID = 'SE_BANKID',
 }
 
-type StatusChangeCallback = (status: MemberLoginStatusResponse['status']) => void
+export const loginMemberSeBankId = (ssn: string): Observable<MemberLoginStatusResponse> => {
+  return new Observable((subscriber) => {
+    let pollTimeoutId: number
+    const poll = async (statusUrl: string) => {
+      if (subscriber.closed) return
```
alebedev, line 13: > The docs explicitly say that an ignore flag is better than an AbortController, because the AbortController can be unreliable if there are asynchronous steps chained after the fetch.

From https://sebastiancarlos.com/react-js-best-practices-from-the-new-docs-1c65570e785d

apps/store/src/services/bankId/bankIdReducer.ts
alebedev: To be done in GRW-2235

apps/store/src/components/BankIdDialog.tsx
```diff
@@ -54,7 +51,9 @@ export const BankIdDialog = () => {
   let content: ReactElement | null = null
   let footer: ReactElement | null = null
 
-  if (currentOperation !== null) {
+  const shopSessionId = shopSession?.id
+  const ssn = shopSession?.customer?.ssn ?? ''
```
robinandeer, line 55: why empty string? would it be more explicit with null/undefined?

apps/store/src/services/bankId/useBankIdCheckoutSign.ts
```diff
@@ -1,60 +1,97 @@
...
+          },
+        })
+      }).subscribe({
+        next(value) {
+          console.log('next', value)
+          dispatch({ type: 'operationStateChange', nextOperationState: value })
+        },
+        complete() {
+          console.log('complete')
+          onSuccess()
```
robinandeer, line 79: remove console log

robinandeer (approved):
It's a lot to take it but I think it looks good :) 

Will be interesting to learn about this new concept

apps/store/src/components/BankIdDialog.tsx
alebedev: It's a string, so shorter type can be used.  Matter of preference, I guess

alebedev:
Job ID: VJXQW2wDUmDAlQgTIHJL.
See details on [graphite.dev](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1612)

PR #1616, opened 2023-02-10
Fix / Start date format cart toast

amandaflod:
## Describe your changes

Update the start date description in the cart toast to include cancellation. 

Screenshot 2023-02-10 at 15 39 38


apps/store/src/components/CartNotification/ProductItem.tsx
```diff
@@ -1,25 +1,22 @@
...
-import { useTranslation } from 'react-i18next'
 import { Space, Text } from 'ui'
 import { Pillow } from '@/components/Pillow/Pillow'
 import { SpaceFlex } from '@/components/SpaceFlex/SpaceFlex'
 
 export type ProductItemProps = {
   name: string
   price: string
-  startDate?: string
+  startDateDescription?: string
```
robinandeer, line 9: I would consider calling it just "description" to make it more flexible

apps/store/src/components/ProductPage/PurchaseForm/PurchaseForm.tsx
```diff
@@ -102,8 +102,12 @@ export const PurchaseForm = () => {
             notifyProductAdded({
               name: productData.displayNameFull,
               price: formatter.monthlyPrice(item.price),
-              /* @TODO: display "automatically switches" if cancellation is requested" */
-              startDate: formatter.fromNow(new Date(item.startDate)),
+              startDateDescription: item.cancellation.requested
+                ? t('CART_ENTRY_AUTO_SWITCH', { ns: 'cart' })
```
robinandeer, line 106: nit: consider moving this translation key from "cart" to "purchase-form" namespace

robinandeer (approved):
Very nice! Only thing I would consider changing is updating the new prop name

guilhermespopolin (approved):
Well done :)

apps/store/src/components/CartNotification/ProductItem.tsx
amandaflod: okidoki, will do!

apps/store/src/components/ProductPage/PurchaseForm/PurchaseForm.tsx
amandaflod: The thing is that we're using it in the CartEntryItem as well, which definitely belongs to cart 🤔 Should I create a duplicate maybe?

PR #1642, opened 2023-02-14
ProductCard title link

KarnellSchultz:
## Describe your changes

Wrapped the title and subtitle text in a link.

Improved the Product card story.

[Notion page](https://www.notion.so/hedviginsurance/Make-the-headline-text-in-Product-grid-as-the-link-text-inside-a-a-not-L-s-mer-OK-if-the-d0e2b5f78a6a473693834d73a43b9c79?pvs=4)

### DIFF
--- a/apps/store/src/components/ProductCard/ProductCard.tsx
+++ b/apps/store/src/components/ProductCard/ProductCard.tsx
@@ -29,7 +29,7 @@ export const ProductCard = ({
 }: ProductCardProps) => {
   const { t } = useTranslation('common')
   return (
-    <Wrapper y={1.5}>
+    <Space y={1.5}>
       <Link href={link} tabIndex={-1} aria-hidden={true}>
         <ImageWrapper aspectRatio={aspectRatio}>
           <Image {...imageProps} alt={alt} fill sizes="100vw" />
@@ -44,13 +44,9 @@ export const ProductCard = ({
           {t('READ_MORE')}
         </ButtonNextLink>
       </CallToAction>
-    </Wrapper>
+    </Space>
   )
 }
-
-const Wrapper = styled(Space)({
-  height: '100%',
-})
 
 const ImageWrapper = styled.div<ImageSize>(({ aspectRatio }) => ({
   display: 'block',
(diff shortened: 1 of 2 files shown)


robinandeer (changes requested):
Now this component is getting really confusing with three clickable elements that all do the same thing 😅 

I would not consider If website the best standard - they use some image with text inside it 😛 

I think we should do what apple does: https://www.apple.com/

1. Wrap the WHOLE component in a `<Link>` with `tabIndex={-1} aria-hidden={true}`
2. Use a `<ButtonNextLink>` for "read more"

robinandeer:
@KarnellSchultz sorry realized just now that Apple has a more advanced setup - "a" should never be nested with "a"

But same goes for nesting "button" inside "a" -- we need another solution for that

KarnellSchultz:
@robinandeer Thanks for the tips. Going to play around with the component for a bit to see if I can get

apps/store/src/components/ProductCard/ProductCard.tsx
```diff
@@ -62,6 +66,7 @@ const ImageWrapper = styled.a<ImageSize>(({ aspectRatio }) => ({
 
   ':hover, :active': {
     opacity: 0.95,
+    transition: 'opacity 0.1s ease-out',
```
KarnellSchultz, line 69: matching the buttons transition

KarnellSchultz:
Screenshot 2023-02-14 at 12 21 49

apps/store/src/components/ProductCard/ProductCard.tsx
```diff
@@ -29,26 +29,30 @@ export const ProductCard = ({
...
+          {t('READ_MORE')}
+        </ButtonNextLink>
+      </CallToAction>
+    </Wrapper>
   )
 }
 
-const ImageWrapper = styled.a<ImageSize>(({ aspectRatio }) => ({
+const Wrapper = styled(Space)({
+  height: '100%',
```
robinandeer, line 52: I think I removed this to fix an issue in Safari 😅

robinandeer (approved):
Very nice!

Just make sure the design is not broken in safari with "height = 100%"

apps/store/src/components/ProductCard/ProductCard.tsx
KarnellSchultz: Removed it and (re)fixed the layout 👍

PR #1644, opened 2023-02-14
Wrap navigation items w/ links

KarnellSchultz:
## Describe your changes

Wrapped the nav items with links to make the whole item clickable. 

Screenshot 2023-02-14 at 11 28 37

[Test the issue in the staging env](https://www.store.dev.hedvigit.com/en-se)

Right now the clickable area is only the text. We'd want to give the user more space to click what they want. 


robinandeer (approved):
Very nice!

PR #1646, opened 2023-02-14
GRW-2241 - feat(StoryblokPage): added SEO info

guilhermespopolin:
## Describe your changes

* Updates `HeadSeoInfo` so it handles more SEO related tags
* Refactor `ProductPageBlock` by leveraging `HeadSeoInfo` in order to add seo related tags into documents `<head>`

image

Requested by Peter.

## Jira issue(s): [GRW-2241](https://hedvig.atlassian.net/browse/GRW-2241)

[GRW-2241]: https://hedvig.atlassian.net/browse/GRW-2241?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/store/src/components/HeadSeoInfo/HeadSeoInfo.tsx
+++ b/apps/store/src/components/HeadSeoInfo/HeadSeoInfo.tsx
@@ -1,17 +1,34 @@
 import { ISbStoryData } from '@storyblok/react'
 import Head from 'next/head'
+import { SEOData } from '@/services/storyblok/storyblok'
 import { isRoutingLocale, toIsoLocale } from '@/utils/l10n/localeUtils'
 import { ORIGIN_URL } from '@/utils/PageLink'
 
-export const HeadSeoInfo = ({ story }: { story: ISbStoryData }) => {
+export const HeadSeoInfo = ({ story }: { story: ISbStoryData<SEOData> }) => {
   // AB testing
-  const { canonicalUrl } = story.content
+  const { canonicalUrl, robots, seoMetaTitle, seoMetaDescription, seoMetaOgImage } = story.content
   // Translations and other alternates
   const { alternates } = story
 
   return (
     <>
-      <Head>{canonicalUrl && <link rel="canonical" href={canonicalUrl} />}</Head>
+      <Head>
+        {canonicalUrl && <link rel="canonical" href={canonicalUrl} />}
+        <meta name="robots" content={robots} />
+        {seoMetaTitle && (
+          <>
+            <meta name="title" content={seoMetaTitle} />
+            <meta property="og:title" content={seoMetaTitle} />
+          </>
+        )}
+        {seoMetaDescription && (
+          <>
+            <meta name="description" content={seoMetaDescription} />
+            <meta property="og:description" content={seoMetaDescription} />
+          </>
+        )}
+        {seoMetaOgImage?.filename && <meta property="og:image" content={seoMetaOgImage.filename} />}
+      </Head>
       {/* Must include link to self along with other variants */}
 
       <AlternateLink fullSlug={story.full_slug} />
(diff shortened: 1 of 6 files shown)


guilhermespopolin:
* main
  * **PR #1646** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1646" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1646?utm_source=stack-comment).

apps/store/src/pages/_document.tsx
```diff
@@ -14,7 +14,6 @@ export default class MyDocument extends Document {
       <Html lang={this.lang()} dir="ltr">
         <Head>
           {/*TODO: Allow indexing right before going live*/}
-          <meta name="robots" content="none" />
```
guilhermespopolin, line 17: I wasn't sure if I can remove this line. The reasoning behind it is to avoid adding index meta tag twice. Let me know what you think.

alebedev: Looks safe

apps/store/src/pages/[[...slug]].tsx
```diff
@@ -77,7 +79,7 @@ export const getStaticProps: GetStaticProps<
   const apolloClient = initializeApollo()
   console.time('getStoryblokData')
   const [story, globalStory, translations, productMetadata] = await Promise.all([
-    getStoryBySlug(slug, { version, locale }),
+    getStoryBySlug<PageStory | ProductStory>(slug, { version, locale }),
```
Comment (robinandeer) [inline apps/store/src/pages/[[...slug]].tsx:82]: why needed?

apps/store/src/components/HeadSeoInfo/HeadSeoInfo.tsx
```diff
@@ -1,17 +1,34 @@
...
+  const { canonicalUrl, robots, seoMetaTitle, seoMetaDescription, seoMetaOgImage } = story.content
   // Translations and other alternates
   const { alternates } = story
 
   return (
     <>
-      <Head>{canonicalUrl && <link rel="canonical" href={canonicalUrl} />}</Head>
+      <Head>
+        {canonicalUrl && <link rel="canonical" href={canonicalUrl} />}
+        <meta name="robots" content={robots} />
```
alebedev, line 17: Do we need to provide default value or Storyblok guarantees that even old pages will page some value here?

alebedev (approved):
Looks good, some questions inside

apps/store/src/pages/_document.tsx
robinandeer: hmm we don't want to expose any of these pages for now do - should editors have possibility to change this on a page by page basis?

apps/store/src/components/HeadSeoInfo/HeadSeoInfo.tsx
guilhermespopolin: No you're right. Storyblok provides default index value for every _Page type_ page. However, one would need to publish each page in order to get the change applied.

robinandeer (approved):
Looks good, just left some general comments

apps/store/src/pages/_document.tsx
guilhermespopolin: Not sure if I've followed but they can change that page by page with SEO tab.

apps/store/src/pages/[[...slug]].tsx
Comment (guilhermespopolin) [reply apps/store/src/pages/[[...slug]].tsx:82]: The pages pre-rendered by this file - StorbyokPage and ProductPage - define some SEO data fields (SEOData type defined inside storyblok service). HeadSeoInfo component then uses those fields to properly add meta tags. If we don't tell TS that the story we're getting have those fields it will complains when we try o pass down that story to HeadSeoInfo

PR #1658, opened 2023-02-14
refact(ProductPageBlock): disable active section dectection for mobile

guilhermespopolin:
## Describe your changes

* Refactor ProductPageBlock between it's two distinct layouts: _Mobile_ and _Desktop_
* Instantiate _active section detection_ only for _Desktop_ layout

Despite having a lot of things in common, it doesn't make sense to instantiate _active section detection_ for both layout modes as we had before since on mobile we display the content using tabs.


guilhermespopolin:
* main
  * **PR #1658** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1658" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1658?utm_source=stack-comment).

apps/store/src/blocks/ProductPageBlock.tsx
```diff
@@ -275,9 +275,14 @@ function useActiveSectionChangeListener(
 ) {
   // Used to determine if user is scrolling up
   const scrollPositionRef = useRef(0)
+  const isLarge = useBreakpoint('lg')
 
   useEffect(
     function identifyActiveSection() {
+      // We only care about updating active section based on scroll while on
+      // desktop layout
+      if (!isLarge) return
```
guilhermespopolin, line 284: [Input Wanted]
So what I want is to "disable" active section detection on mobile since it doesn't make sense having it on that scenario. I can see two options:

1. Do it inside the hook as I'm doing here. It simple but not so declarative.
2. Split Desktop/Mobile layout between their own components. By including the hook call inside DesktopLayout only we'd be preventing it to be execute for mobile layout. That's more declarative, but both components share a lot of stuff, so I can see some level of jumping between files while maintaining them.

Any ideas/suggestions?

apps/store/src/blocks/ProductPageBlock/ProductPageBlock.tsx
```diff
@@ -0,0 +1,44 @@
...
+export type ProductPageBlockProps = SbBaseBlockProps<{
+  overviewLabel: string
+  coverageLabel: string
+  overview: SbBlokData[]
+  coverage: SbBlokData[]
+  body: SbBlokData[]
+}>
+
+export const ProductPageBlock = ({ blok }: ProductPageBlockProps) => {
+  const isLarge = useBreakpoint('lg')
```
robinandeer, line 19: let's hope this doesn't give us too many SSR issues

apps/store/src/blocks/ProductPageBlock/MobileLayout.tsx
```diff
@@ -0,0 +1,68 @@
...
+  return (
+    <>
+      <PurchaseForm />
+      <Content>
+        <RadixTabs.Tabs
+          value={activeSection}
+          onValueChange={(value) => {
+            setActiveSection(value as PageSection)
+            window?.scrollTo({ top: 0 })
+          }}
```
robinandeer, line 35: nit: extract as function

apps/store/src/blocks/ProductPageBlock/DesktopLayout.tsx
```diff
@@ -0,0 +1,144 @@
...
+
+const useActiveSectionChangeListener = (
+  sectionsId: Array<string>,
+  callback: (activeSectionId: string) => void,
+) => {
+  // Used to determine if user is scrolling up
+  const scrollPositionRef = useRef(0)
+
+  useEffect(
+    function identifyActiveSection() {
```
robinandeer, line 104: hmm why do we name this function? that's not React convention - just use arrow function I would say

robinandeer (approved):
Great work working on this -- I wonder if we really need such a complicated solution but let's worry about that post-launch I guess

guilhermespopolin:
> Great work working on this -- I wonder if we really need such a complicated solution but let's worry about that post-launch I guess

With complicated solution you mean the chosen design or the implementation of it? 🤔

PR #1667, opened 2023-02-15
Add custom 500 error page

robinandeer:
## Describe your changes

- Add custom error 500 page
- Refactor 404 error page and create reusable error page component

They are basically the same page.

Caveat: it's not 100% clear what should happen when user clicks "try again".

### DIFF
--- a/apps/store/public/locales/en/common.json
+++ b/apps/store/public/locales/en/common.json
@@ -1,6 +1,8 @@
 {
   "404_PAGE_BUTTON": "Go to start page",
   "404_PAGE_MESSAGE": "The page could not be found",
+  "500_PAGE_BUTTON": "Try again",
+  "500_PAGE_MESSAGE": "Something went wrong, try again",
   "CALCULATE_PRICE": "Calculate price",
   "COUNTRY_LABEL_DK": "Denmark",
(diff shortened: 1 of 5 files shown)


robinandeer:
* main
  * **PR #1667** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1667" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1667?utm_source=stack-comment).

apps/store/public/locales/en/common.json
```diff
@@ -1,13 +1,16 @@
 {
   "404_PAGE_BUTTON": "Go to start page",
   "404_PAGE_MESSAGE": "The page could not be found",
+  "500_PAGE_BUTTON": "Try again",
```
alebedev, line 4: I think this should be something like "Go to home page".  500 errors that we didn't catch are probably unrecoverable without at least full reload

alebedev (approved):
Looks good

robinandeer:
@alebedev This is what would happen currently, the rest is just a copy question so let's merge and ask design

PR #1670, opened 2023-02-15
Predefine different page/block layouts

robinandeer:
## Describe your changes

Setup pre-defined layouts based on the 12 column grid.

We have 4 pre-defined widths:

- full (1)
- two-thirds (2/3)
- half (1/2)
- third (1/3)

We have 3 pre-defined alignments:

- left
- center
- right

This will make it easier to create layouts that are consistent with the design system. They will also behave the same across breakpoints.

The idea is to control these settings from the CMS on a block level.


robinandeer:
* main
  * **PR #1670** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1670" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1670?utm_source=stack-comment).

apps/store/src/components/GridLayout/GridLayout.tsx
```diff
@@ -17,23 +17,68 @@ const Root = styled.div({
   },
 })
 
-const Content = styled.div({
-  gridColumn: '1 / span 12',
-
-  [mq.md]: {
-    gridColumn: '3 / span 8',
-  },
+type ContentWidth = '1' | '2/3' | '1/2' | '1/3'
```
alebedev, line 20: Nit: Can you sort the values?

apps/store/src/components/GridLayout/GridLayout.tsx
```diff
@@ -17,23 +17,68 @@ const Root = styled.div({
...
+}
+
+const thirdCenterStyles: CSSObject = {
+  ...baseStyles,
+  [mq.md]: { gridColumn: '3 / span 8' },
+  [mq.lg]: { gridColumn: '4 / span 6' },
+  [mq.xl]: { gridColumn: '5 / span 4' },
+}
+
+const STYLES: Record<ContentWidth, Record<ContentAlignment, CSSObject>> = {
```
alebedev, line 67: Nit: From the look of it, you can omit every `smth: baseStyles` rule and get the same end result

apps/store/src/components/GridLayout/GridLayout.tsx
```diff
@@ -17,23 +17,68 @@ const Root = styled.div({
...
+  width: ContentWidth
+  align?: ContentAlignment
+}
 
-  [mq.xl]: {
-    gridColumn: '5 / span 4',
-  },
-})
+const Content = styled.div<ContentProps>(({ width, align }) => ({
+  gridColumn: '1 / span 12',
```
alebedev, line 29: If it must be same rule as `baseStyles`, can we deduplicate it or remove one of the versions?

alebedev (approved):
Nice!  Some ideas that may simplify implementation inside

apps/store/src/components/GridLayout/GridLayout.tsx
robinandeer: how do you mean? they are sorted by size already 🤔

apps/store/src/components/GridLayout/GridLayout.tsx
robinandeer: oh sorry, this must've slipped through

apps/store/src/components/GridLayout/GridLayout.tsx
alebedev: My bad, expected increasing order.  It's fine 😃

alebedev (approved):

PR #1689, opened 2023-02-17
Add React-wrap-balancer to `<Heading />`

KarnellSchultz:
## Describe your changes

- Added react-wrap-balancer
- Updated the heading component to use the balancer
- updated heading story to show better how the balancer works

I see several places where the balancer could be used. It's up to us to decide if/how much we'd like to use it.

### Without Balancer

https://user-images.githubusercontent.com/50870173/219662083-c403f6bd-ea5b-411b-ad49-27168108afda.mov

### With Balancer

https://user-images.githubusercontent.com/50870173/219661432-9ed6f1bd-9924-4fd1-abb9-9eb8c263a8e4.mov

### Heading Story

Screenshot 2023-02-17 at 14 21 28

Help make our titles more readable.

### DIFF
--- a/apps/store/package.json
+++ b/apps/store/package.json
@@ -48,6 +48,7 @@
     "react": "18.2.0",
     "react-dom": "18.2.0",
     "react-i18next": "12.1.5",
+    "react-wrap-balancer": "0.4.0",
     "ui": "workspace:",
     "uuid": "9.0.0",
     "zen-observable-ts": "1.2.5"
--- a/apps/store/src/pages/_app.tsx
+++ b/apps/store/src/pages/_app.tsx
@@ -29,6 +29,7 @@ import { useDebugTranslationKeys } from '@/utils/l10n/useDebugTranslationKeys'
 import { useAllowActiveStylesInSafari } from '@/utils/useAllowActiveStylesInSafari'
 import { useReloadOnCountryChange } from '@/utils/useReloadOnCountryChange'
 
+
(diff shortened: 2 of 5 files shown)


alebedev (approved):
Very nice!

gustaveen (approved):
Nice one! @KarnellSchultz ⭐️ This is a subject close at heart for designers to balance headings and I think it looks much better. I think this solution is nice that it automagically fix it for us. Since the alternative is that we will need to add settings for max-width in headings or support line-breaks. So I think it comes down if Zak an co thinks it does the trick. 

One alternative we discussed is to work with `max-width` and `ch` (characters) unit. So you can limit the width based on characters and the text would then "break" the same regardless of font-size. Not a perfect solution but gives some more more control. 


https://user-images.githubusercontent.com/6661511/220083020-57b04fb9-bfaf-44c6-a526-70b4ca0dc4df.mov

KarnellSchultz:
@gustaveen Thanks for checking it out. I wanted to get some of y'alls input on this before going forward with this pr. 

One thing I wanted to point out is that you're able to customize the "balance ratio" With Balancer as well. Might be worth checking out (or showing to the designer folks) as well. If you haven't already see it in the balancer docs they have a little demo user ["Custom Balance Ratio"](https://react-wrap-balancer.vercel.app/)

[Another interesting blog post I found on the topic ](https://www.ctrl.blog/entry/text-wrap-balance.html)

PR #1690, opened 2023-02-17
fix / add switching to house

amandaflod:
## Describe your changes

Updated the se-house form to include currentinsurer field and added a section to keep the form less cluttered after syncing with Zak. 


alebedev (approved):
Good work!

PR #1723, opened 2023-02-21
Pause/resume videos when FullscreenDialog opens or closes

alebedev:
## Describe your changes

Send open/close events from FullscreenDialog to pause/resume videos in background

We use FullscreenDialog in Change SSN warning.  Tried to make this solution generic so that it works for BankId dialog,  potential error alerts, etc

### DIFF
--- a/apps/store/src/components/FullscreenDialog/FullscreenDialog.tsx
+++ b/apps/store/src/components/FullscreenDialog/FullscreenDialog.tsx
@@ -1,5 +1,8 @@
 import styled from '@emotion/styled'
+import { DialogProps } from '@radix-ui/react-dialog'
+import { useEffect, useRef } from 'react'
 import { CrossIcon, Dialog, mq, theme } from 'ui'
+import { sendDialogEvent } from '@/utils/dialogEvent'
 
 type Props = {
   children: React.ReactNode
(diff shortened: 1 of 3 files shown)


alebedev:
* main
  * **PR #1723** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1723" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1723?utm_source=stack-comment).

apps/store/src/components/Video/Video.tsx
```diff
@@ -107,7 +107,10 @@ export const Video = ({
   useDialogEvent('open', handleDialogOpen)
 
   const handleDialogClose = useCallback(() => {
-    if (wasPlaying) {
+    // We may have dialogs on top of dialogs.  Need to check that all are closed to resume video
+    // Since we don't have anything stable like data attribute for it, let's check styles
+    const hasOpenDialog = window.getComputedStyle(document.body).pointerEvents === 'none'
+    if (wasPlaying && !hasOpenDialog) {
```
robinandeer, line 113: nit: maybe it would be clearer to return early if `hasOpenDialog`

apps/store/src/components/FullscreenDialog/FullscreenDialog.tsx
```diff
@@ -96,5 +99,26 @@ const FooterWrapper = styled.footer({
   },
 })
 
-export const Root = Dialog.Root
+export const Root = (props: DialogProps) => {
+  useSendDialogEvents(!!(props.open || props.defaultOpen))
```
robinandeer, line 103: Would this still work for uncontrolled dialogs? Then the props won't change...

We also have a race condition if "defaultOpen" changes from one value to another -- that shouldn't trigger a dialog event if the component is already mounted, right? Not a big problem but annoying :P

robinandeer (commented):
Appreciate you looking into this -- don't have a great solution in mind

Still, I would consider investigating how bad it would be to send these events in whatever event handler which causes a dialog to open/close.

The second part I think it pretty decent though -- avoiding to resume video if we can guess there's an active modal!

apps/store/src/components/FullscreenDialog/FullscreenDialog.tsx
alebedev: It works, made SSN change alert uncontrolled to verify.
Uncontrolled dialogs should set `defaultOpen=true` and never change it, see https://www.radix-ui.com/docs/primitives/components/dialog#root

alebedev:
OK, let's discuss it tomorrow in the office - does not look urgent

apps/store/src/components/FullscreenDialog/FullscreenDialog.tsx
robinandeer: Hmm why `defaultOpen=true`? Could also render a button on the page that opens the modal, right?

But I guess then, `onOpenChange` would run

PR #1724, opened 2023-02-21
Make balancing the header optional

KarnellSchultz:
## Describe your changes

- Added a `balance` prop to the `<Heading /> ` component to allow for the balancer to be adopted in a case by case basis.
- removed the balancer provider from the `_app` dir
- Updated the Heading block in Storyblok to allow the balancer to be toggled. 

## notes
- If we decide to use react-wrap-balancer more around the app we'd probably want to use a provider to share the balancing script with all the components. 
- If we start using server components there are some optimisations available as well.

**(Note the 'balance' option in the bottom right of the screenshots)**
Screenshot 2023-02-21 at 10 24 51

Screenshot 2023-02-21 at 10 24 58

Adding the balancer to every heading component caused issues with some of the headings around the app. [Check out this thread from slack](https://hedviginsurance.slack.com/archives/G01MARCFUUW/p1676966424962669)

(screenshot of the issue)
image

### DIFF
--- a/apps/store/src/pages/_app.tsx
+++ b/apps/store/src/pages/_app.tsx
@@ -4,7 +4,6 @@ import { appWithTranslation } from 'next-i18next'
 import type { AppPropsWithLayout } from 'next/app'
 import Head from 'next/head'
 import Router from 'next/router'
-import { Provider as BalancerProvider } from 'react-wrap-balancer'
 import { globalStyles, theme } from 'ui'
 import { BankIdDialog } from '@/components/BankIdDialog'
 import { useApollo } from '@/services/apollo/client'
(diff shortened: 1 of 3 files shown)


apps/store/src/pages/_app.tsx
```diff
@@ -82,9 +80,7 @@ const MyApp = ({ Component, pageProps }: AppPropsWithLayout) => {
           <ShopSessionProvider shopSessionId={pageProps[SHOP_SESSION_PROP_NAME]}>
             <TrackingProvider value={tracking}>
               <BankIdContextProvider>
-                <BalancerProvider>
```
alebedev, line 85: Try keeping the provider - maybe the fact that non-CMS headers are not balanced by default will fix the problems Robin reported

KarnellSchultz: Sounds like a plan 👍

PR #1754, opened 2023-02-22
GRW-2289 - refact(ImageTextBlock): leverage ImageBlock for image rendering

guilhermespopolin:
## Describe your changes

* Update `ImageTextBlock` so it leverages an `ImageBlock` to get an image rendered;
* General updates: Export _wrapper_ components defined inside `ButtonBlock`, `HeadingBlock` and `ImageBlock` components so I can use then to override spacing related styles (E.g: padding) while rendering those blocks inside an `ImageTextBlock`. Due to cycle dependencies, I couldn't achieve the same by doing the inverse as we've done in some places in the codebase; That would mean, import parent container inside children and use it to get children's styles overridden.

What _design team_ requested was a way for editors to configure `ImageTextBlock`'s image aspect ratios based on screen orientation - something that's already supported by `ImageBlock`. So instead of updating `ImageTextBlock` to receive the same controls `ImageBlock` does to achieve that goal, I'm composing both blocks.

## Next steps

As one can see, I've added a new field for `ImageTextBlock` - `imageBlock`. My idea is to get this reviewed and merged first. Then I'm going to remove old `image` field and rename ~`imageBlock`~ --> `image`.

image

## Jira issue(s): [GRW-2289](https://hedvig.atlassian.net/browse/GRW-2289)

[GRW-2289]: https://hedvig.atlassian.net/browse/GRW-2289?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/store/src/blocks/HeadingBlock.tsx
+++ b/apps/store/src/blocks/HeadingBlock.tsx
@@ -2,15 +2,10 @@ import styled from '@emotion/styled'
 import { storyblokEditable } from '@storyblok/react'
 import { ConditionalWrapper, Heading, HeadingProps, PossibleHeadingVariant, theme } from 'ui'
 import { SbBaseBlockProps } from '@/services/storyblok/storyblok'
-import { VerticalBodyWrapper, FluidBodyWrapper } from './ImageTextBlock'
 
 const Wrapper = styled.div({
   paddingLeft: theme.space.md,
   paddingRight: theme.space.md,
-
-  [`${VerticalBodyWrapper} &, ${FluidBodyWrapper} &`]: {
-    padding: 0,
-  },
 })
 
 export type HeadingBlockProps = SbBaseBlockProps<{
--- a/apps/store/src/blocks/ButtonBlock.tsx
+++ b/apps/store/src/blocks/ButtonBlock.tsx
@@ -1,11 +1,10 @@
 import styled from '@emotion/styled'
 import { storyblokEditable } from '@storyblok/react'
 import { ComponentProps } from 'react'
-import { Button, theme } from 'ui'
+import { Button, ConditionalWrapper, theme } from 'ui'
 import { ButtonNextLink } from '@/components/ButtonNextLink'
 import { LinkField, SbBaseBlockProps } from '@/services/storyblok/storyblok'
 import { getLinkFieldURL } from '@/services/storyblok/Storyblok.helpers'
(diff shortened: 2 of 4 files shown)


guilhermespopolin:
* main
  * **PR #1754** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1754" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1754?utm_source=stack-comment).

apps/store/src/blocks/HeadingBlock.tsx
```diff
@@ -44,3 +34,8 @@ export const HeadingBlock = ({ blok }: HeadingBlockProps) => {
   )
 }
 HeadingBlock.blockName = 'heading'
+
+export const Wrapper = styled.div({
+  paddingLeft: theme.space.md,
+  paddingRight: theme.space.md,
```
robinandeer, line 40: nit: padding-inline

robinandeer (approved):
Nice work!

It's starting to get a little hard to follow but okey for now

I think generally if these are intended as reusable styles / styled components we should move them out of the blocks folder. That way we can import and avoid circular imports.

apps/store/src/blocks/HeadingBlock.tsx
```diff
@@ -44,3 +34,8 @@ export const HeadingBlock = ({ blok }: HeadingBlockProps) => {
...
+})
```
gustaveen, line 41: Hm, I leaning towards using the `ConditionalWrapper` pattern for using blocks inside other blocks and thus removing padding, instead of exporting Wrapper from every component. If you look at `HeadingBlock` that's already implemented with the `nested` prop. However I realise it's not the best to put it inside `blok` object, but we should rather make it a separate prop like so `<StoryblokComponent nested={true}>`

gustaveen: See this PR and add your thoughts https://github.com/HedvigInsurance/racoon/pull/1767 :)

guilhermespopolin: Have updated this one to take advantage of `nested`

PR #1767, opened 2023-02-23
Put nested prop outside of Storyblok data

gustaveen:
## Describe your changes
Put `nested` prop outside Storyblok data

When using blocks inside other blocks, we don't want to render any padding (meaning no need for a `Wrapper`). So instead of overriding with nested CSS, exporting Wrapper from every component, I'm leaning towards the `ConditionalWrapper` pattern that we can reuse across blocks. 

### DIFF
--- a/apps/store/src/blocks/TextContentBlock.tsx
+++ b/apps/store/src/blocks/TextContentBlock.tsx
@@ -25,10 +25,7 @@ export const TextContentBlock = ({ blok }: Props) => {
       >
         <Space y={1}>
           {blok.body?.map((nestedBlock) => (
-            <StoryblokComponent
-              key={nestedBlock._uid}
-              blok={{ ...nestedBlock, ...{ nested: true } }}
-            />
+            <StoryblokComponent key={nestedBlock._uid} blok={nestedBlock} nested={true} />
           ))}
         </Space>
       </GridLayout.Content>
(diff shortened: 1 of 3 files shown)


apps/store/src/blocks/TextContentBlock.tsx
```diff
@@ -25,10 +25,7 @@ export const TextContentBlock = ({ blok }: Props) => {
       >
         <Space y={1}>
           {blok.body?.map((nestedBlock) => (
-            <StoryblokComponent
-              key={nestedBlock._uid}
-              blok={{ ...nestedBlock, ...{ nested: true } }}
-            />
+            <StoryblokComponent key={nestedBlock._uid} blok={nestedBlock} nested={true} />
```
guilhermespopolin, line 28: It's definitely a tricky thing these mechanics with nested blocks. I mean, ideally I'd solve the issue of wanting to override some styles based on the context a component is rendered with composition. As an example, if I have a `Title` component that should have some styles overridden when rendered as child of a `Container` component I would, inside `Container` component, create a `StyledTitle` --> `const StyledTitle = styled(Title)({...})`. But that's not possible with nested blocks because we don't know what we would end up rendering. And I definitely agree that overriding with CSS selectors is became hard to follow 😬 

I think that could be a good solution for the issue at hand :)
Definitely gonna tried it out on this [PR](https://github.com/HedvigInsurance/racoon/pull/1754)

My only consideration though, is that it would be nice if `nested` was proper typed for `StoryblokComponent`. I mean, its type allows us pass down anything but it would be nice to have that level of awareness while working with it.

```
interface StoryblokComponentProps {
    blok: SbBlokData;
    [key: string]: unknown;
}
```

guilhermespopolin (approved):

apps/store/src/blocks/TextContentBlock.tsx
gustaveen: True, we could improve the typing for sure! 🙌

PR #1771, opened 2023-02-23
Fix various text field bugs

robinandeer:
## Describe your changes

- Fix delete button on Safari by not using `display: none`
- Only run warning animation 1 time (instead of 2)
- Fix label color warning animation
- Make campaign code input required

Safari doesn't like setting display to none on a button. It seems like it just doesn't register the event handlers. Instead I just show/hide the button with opacity which I think works fine. Let me know if you think there are some accessibility issues with this.

### DIFF
--- a/apps/store/src/components/TextField/TextField.tsx
+++ b/apps/store/src/components/TextField/TextField.tsx
@@ -77,7 +77,7 @@ export const TextField = (props: Props) => {
             (inputProps.disabled ? (
               <LockIcon size="1rem" color={theme.colors.textSecondary} />
             ) : (
-              <DeleteButton type="button" onClick={handleClickDelete}>
+              <DeleteButton type="button" onClick={handleClickDelete} aria-hidden={true}>
                 <CrossIcon size="1rem" />
               </DeleteButton>
             ))}
@@ -104,7 +104,7 @@ const warningColorAnimation = keyframes({
     color: theme.colors.amberText,
   },
   '100%': {
-    color: theme.colors.textPrimary,
+    color: theme.colors.textSecondary,
   },
 })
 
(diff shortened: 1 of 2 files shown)


robinandeer:
* main
  * **PR #1771** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1771" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1771?utm_source=stack-comment).

apps/store/src/components/TextField/TextField.tsx
```diff
@@ -199,10 +199,10 @@ const SmallInput = styled(LargeInput)({ fontSize: theme.fontSizes.lg })
 const MessageText = styled(Text)({ paddingLeft: theme.space.md })
 
 const DeleteButton = styled.button({
-  display: 'none',
   cursor: 'pointer',
+  opacity: 0,
```
gustaveen, line 203: Would it need to have an `aria-hidden` attribute? Just wondering myself :)

gustaveen (approved):
Sweet! 📱

apps/store/src/components/TextField/TextField.tsx
guilhermespopolin: I guess that might be a good idea.

guilhermespopolin (approved):

apps/store/src/components/TextField/TextField.tsx
robinandeer: no... that's not needed, right? The input device would first focus the input field and then we want the delete button to be focusable

Hmm but perhaps this is not really a useful feature for assistive devices in the first place

We should start adding aria labels tho for icons buttons

robinandeer:
Job ID: jzn3ukcQStl012j9oZlr.
See details on [graphite.dev](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1771)

PR #1802, opened 2023-02-28
Remove renewal date stuff

amandaflod:


gustaveen (approved):
Don't have the best overview on what it is that should be removed 😄 But this fixes the builds so I'm happy, good job Amanda! 🎉

PR #1806, opened 2023-02-28
Refactor `<ContactSupport />` component

KarnellSchultz:
## Describe your changes

- 🔧 Tightened up the types. Specifically on the `icon=` prop so it now takes a limited number of Icons. Before it took a react component as a prop.
- General cleanup of code.
- Moved the contact card into the contact support file.

Screenshot 2023-02-28 at 14 12 18

### Figma for reference
the lottie icons are a bit smaller in the design then they are IRL
Screenshot 2023-02-28 at 14 14 06

Make this component easier to use / easier to change going forward.


apps/store/src/components/ContactSupport/ContactCard.tsx
```diff
@@ -1,5 +1,38 @@
...
+type ContactCardProps = {
+  icon: (typeof IconOptions)[keyof typeof IconOptions]
+  children: React.ReactNode
+}
+
+export const ContactCard = ({ icon, children }: ContactCardProps) => {
+  const IconComponent = {
+    message: Message,
+    numberPad: NumberPad,
+  }[icon]
```
guilhermespopolin, line 20: nit
Maybe we could have an constant _map_ object outside the component that register an icon for each possible icon option, and them just use that constant inside your component:

```
// not sure about how to type that Map object though 🤔 
// based on the other types, I guess it would need to be
// const ICON_MAP: Record<(typeof IconOptions)[keyof typeof IconOptions], React.ReactNode>
// nah, too complicated. Let's leave it as it is :)
const ICON_MAP = {
  message: Message,
  numberPad: NumberPad
}

export const Contractcard = ({ icon, children }: ContractCardProps) => {
   const IconCoponent = ICON_MAP[icon]
}
```

guilhermespopolin (approved):

apps/store/src/components/ContactSupport/ContactCard.tsx
KarnellSchultz: This works 👍 no fancy types needed. 

```
const ICON_MAP = {
  message: Message,
  numberPad: NumberPad,
} as const


export const ContactCard = ({ icon, children }: ContactCardProps) => {
  const IconComponent = ICON_MAP[icon]
  return (
    <Card>
      <Space y={1.5}>
        <IconComponent />
        <div>{children}</div>
      </Space>
    </Card>
  )
}
```

Although I'm Interested in why you might prefer that way? 

Guessing it's due to some performance impact? Having the ICON_MAP inside the component as it's rerendered might have?

guilhermespopolin (approved):

apps/store/src/components/ContactSupport/ContactCard.tsx
KarnellSchultz: cool cool cool, thanks for the insights 👍

PR #1856, opened 2023-03-02
Intercom - lazy loading 🐡

KarnellSchultz:
## Describe your changes

Refactored the intercom button to allow it to be lazy loaded. 

Intercom is now loaded after the user clicks on the support button.

Allow the user to click the button to open and close the intercom window.

### note
since we're loading the scrips onclick we may want to have a loading state for this component

Intercom is hurting our page performance and should only be loaded when the user needs it.

### could use a hand 

When testing this component (sometimes) there is now a while flash the first time the user clicks on the button to open intercom. I tried several ways to fix this but could not get it sorted.

### DIFF
--- a/apps/store/src/components/ContactSupport/IntercomChatButton.tsx
+++ b/apps/store/src/components/ContactSupport/IntercomChatButton.tsx
@@ -37,7 +37,7 @@ export const IntercomChatButton = ({ children }: Props) => {
     }
   }, [appId])
 
-  // if (!appId || !INTERCOM_ENABLED) return <>{children}</>
+  if (!appId || !INTERCOM_ENABLED) return <>{children}</>
 
   if (!isLoaded)
     return <IntercomButton onClick={() => setIsLoaded(true)}>{children}</IntercomButton>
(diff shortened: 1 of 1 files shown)


apps/store/src/components/ContactSupport/IntercomChatButton.tsx
```diff
@@ -1,50 +1,54 @@
...
-  const { show } = useIntercom()
+  const { show, hide, isOpen } = useIntercom()
 
-  return <IntercomButton onClick={show}>{children}</IntercomButton>
+  useEffect(() => {
+    show()
+    return () => {
+      hide()
+    }
+  }, [hide, show])
```
robinandeer, line 25: Do we really need this? Can't we tell intercom to show on load -- or expose this function so we can "show()" in the on-click handler?

robinandeer (approved):
Nice work!

I think the load time seems fine, not sure we need more of a loading state

apps/store/src/components/ContactSupport/IntercomChatButton.tsx
KarnellSchultz: This works too
```
const WithIntercom = ({ children }: Props) => {
  const { show } = useIntercom()

  useEffect(() => {
    show()
  }, [show])

  return <IntercomButton onClick={show}>{children}</IntercomButton>
}
```

I attempted to avoid using this useEffect but intercom does not have a "openOnLoad" feature at the moment.

Without the useEffect the user would need to click on this button twice for intercom to open. 

Open to other ideas as well.

PR #1857, opened 2023-03-02
fix / date picker move chevron

amandaflod:
## Describe your changes

Fix alignment inside date picker so that the chevron icon is level with the value text.

### DIFF
--- a/apps/store/src/components/ProductPage/PurchaseForm/CancellationForm/DateInput.tsx
+++ b/apps/store/src/components/ProductPage/PurchaseForm/CancellationForm/DateInput.tsx
@@ -2,7 +2,7 @@ import styled from '@emotion/styled'
 import { motion } from 'framer-motion'
 import { useTranslation } from 'next-i18next'
 import { ChangeEventHandler, InputHTMLAttributes, useId, useState } from 'react'
-import { ChevronIcon, Space, theme, Text } from 'ui'
+import { ChevronIcon, theme, Text } from 'ui'
 import { convertToDate } from '@/utils/date'
 import { useFormatter } from '@/utils/useFormatter'
 import { useHighlightAnimation } from '@/utils/useHighlightAnimation'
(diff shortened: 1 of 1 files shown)


apps/store/src/components/ProductPage/PurchaseForm/CancellationForm/DateInput.tsx
```diff
@@ -32,46 +32,50 @@ export const DateInput = (props: Props) => {
...
   cursor: 'pointer',
 })
 
+const InnerWrapper = styled.div({
+  width: '100%',
+  display: 'flex',
+  flexDirection: 'row',
+  alignItems: 'center',
+  justifyContent: 'space-between',
+  padding: `${theme.space.xl} 1.25rem ${theme.space.sm} ${theme.space.md}`,
```
robinandeer, line 76: nit: break into multiple attributes

apps/store/src/components/ProductPage/PurchaseForm/CancellationForm/DateInput.tsx
```diff
@@ -32,46 +32,50 @@ export const DateInput = (props: Props) => {
...
-
-        <StyledChevronIcon size="1rem" />
-      </Wrapper>
-    </Space>
+    <Wrapper {...animationProps} data-active={!!inputValue}>
+      <Label htmlFor={inputProps.id} data-disabled={inputProps.disabled}>
+        {label}
+      </Label>
+      <InnerWrapper>
+        {displayValue} <StyledChevronIcon size="1rem" />
```
robinandeer, line 48: nit: I think the default size is 1rem, no?

robinandeer (approved):
Looks good (tested in chrome desktop)

Added via Giphy

apps/store/src/components/ProductPage/PurchaseForm/CancellationForm/DateInput.tsx
amandaflod: The Chevron for some reason takes size 1,5 as a default!

apps/store/src/components/ProductPage/PurchaseForm/CancellationForm/DateInput.tsx
amandaflod: will do :)

PR #1865, opened 2023-03-03
Add gray placeholder while loading images

gustaveen:
## Describe your changes
Show gray placeholder while loading images.

Tested with adding a blurDataURL with a simple gray image but it automatically adds filter blur which looks weird.
So easier to add a background color, but since images can have alpha layers (opacity) we should remove the background color after the image is loaded.

You can test in [preview app](https://hedvig-dot-8wufb0tih-hedvig.vercel.app/se) by using network throttling 

https://user-images.githubusercontent.com/6661511/222730364-fb4368cf-d439-4ff0-b0c3-db62c6c37cb7.mov

Should show a placeholder when loading takes a long time. 

### DIFF
--- a/apps/store/src/blocks/BannerBlock.tsx
+++ b/apps/store/src/blocks/BannerBlock.tsx
@@ -1,7 +1,7 @@
 import styled from '@emotion/styled'
 import { storyblokEditable } from '@storyblok/react'
-import { default as NextImage } from 'next/image'
 import { Heading, Text, mq, theme } from 'ui'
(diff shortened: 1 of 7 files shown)


robinandeer (approved):
Looks great! ✨ 

Would love to add this also to the pillows in the top menu -- or just load them in the background 🤔

gustaveen:
> Looks great! ✨
> 
> Would love to add this also to the pillows in the top menu -- or just load them in the background 🤔

Yes, didn't test that in this PR but I'll see how it looks 🙌

PR #1887, opened 2023-03-06
update csp rules

KarnellSchultz:
## Describe your changes

- update csp rules

- removed `https://js.intercomcdn.com` from scripts to see if it's actually being used there. I think it was misconfigured from the beginning.


alebedev (approved):
Nice!

PR #1230, opened 2023-01-03
Enable overflow scroll in offer presenter

robinandeer:
## Describe your changes

See title.

[Screen Recording 2023-01-03 at 10.32.38.mov](https://graphite-user-uploaded-assets-prod.s3.amazonaws.com/OgXegTwxM9IeuXHZyw5h/f32ee03d-55ad-425b-8768-5c3c622a4e59/Screen%20Recording%202023-01-03%20at%2010.32.38.mov)

## Justify why they are needed

When displaying tier selector, the content is overflowing so you can't scroll down to the "add to cart" button.

## Checklist before requesting a review

- [ ] I have performed a self-review of my code

PR #1894, opened 2023-03-07
styles(ProductDetailsPage): remove double scroll

guilhermespopolin:
## Describe your changes

See title ☝🏻 

It was added [here](https://github.com/HedvigInsurance/racoon/pull/1230/files). Not sure the problem it was trying to fix when it got added. Maybe @robinandeer remembers it?!
Let me know if I'm missing something.

**Before**

Screenshot 2023-03-07 at 11 06 43

**After**
Screenshot 2023-03-07 at 11 10 47

### DIFF
--- a/apps/store/src/blocks/ProductPageBlock/DesktopLayout.tsx
+++ b/apps/store/src/blocks/ProductPageBlock/DesktopLayout.tsx
@@ -89,9 +89,6 @@ const ContentNavigationTrigger = styled.a(triggerStyles)
 const PurchaseFormWrapper = styled.div({
   position: 'sticky',
   top: 0,
-  // Scroll independently if content is too long
-  maxHeight: '100vh',
-  overflow: 'auto',
   paddingBlock: theme.space.xl,
 
   paddingTop: '3vw',


guilhermespopolin:
* main
  * **PR #1894** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1894" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1894?utm_source=stack-comment).

robinandeer (commented):
@guilhermespopolin the issue is that when you try to scroll in car PDP you can't get to the bottom of the right hand content unless you scroll very far down 🤔 

But I would show this to design and ask which version is preferrable!

guilhermespopolin:
> @guilhermespopolin the issue is that when you try to scroll in car PDP you can't get to the bottom of the right hand content unless you scroll very far down 🤔
> 
> But I would show this to design and ask which version is preferrable!

Ohh I see now. Thanks for the explanation. I thought it was solving an issue that might not be still present, but that's not the case.

robinandeer:
@guilhermespopolin hmm yeah neither solution is very good :P

PR #1895, opened 2023-03-07
csp update

KarnellSchultz:
## Describe your changes

- add new csp records

security

### DIFF
--- a/apps/store/next-csp.config.js
+++ b/apps/store/next-csp.config.js
@@ -51,6 +51,8 @@ const fontSrc = [
   'https://fonts.gstatic.com',
   // Vercel
   'https://*.vercel.com',
+
+  'https://at.alicdn.com',
   "'self'",
 ]
 const imgSrc = [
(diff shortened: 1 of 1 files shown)


apps/store/next-csp.config.js
```diff
@@ -51,6 +51,8 @@ const fontSrc = [
   'https://fonts.gstatic.com',
   // Vercel
   'https://*.vercel.com',
+
+  'https://at.alicdn.com',
```
KarnellSchultz, line 55: not totally sure what this, guessing it's from intercom

robinandeer, line 55: Do we know what this even is for? Not sure if we should whitelist in that case, or??

Let's investigate and comment before adding :)

apps/store/next-csp.config.js
```diff
@@ -110,6 +112,7 @@ const frameSrc = [
   'https://dc.insurely.com',
   'https://player.vimeo.com',
   'https://vercel.live', // Vercel Live
+  'https://tpc.googlesyndication.com',
```
robinandeer, line 115: Same here 🤔  what do we need this for?

apps/store/next-csp.config.js
KarnellSchultz: it only came up once. Thinking we should not add it / investigate it a bit more.

KarnellSchultz:
going to hold off on both of these changes

PR #1898, opened 2023-03-07
GRW-2348 / Exclude pages from google indexing

amandaflod:
## Describe your changes

Added a robot.txt file with the pages to exclude. I have never done this before and am not sure how to test it so please feedback if something looks weird! 🤗

[GRW-2348]: https://hedvig.atlassian.net/browse/GRW-2348?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


alebedev (approved):
Nice!  Let's. add meta noindex to all these pages as well, to be 100% sure

amandaflod:
> Nice! Let's. add meta noindex to all these pages as well, to be 100% sure

Ok! Will that then act as some kind of fallback? @alebedev

apps/store/src/pages/_document.tsx
```diff
@@ -22,6 +22,7 @@ export default class MyDocument extends Document {
           <link rel="manifest" href="/site.webmanifest" />
           <link rel="mask-icon" href="/safari-pinned-tab.svg" color={theme.colors.gray1000} />
           <meta name="msapplication-TileColor" content={theme.colors.gray1000} />
+          <meta name="robots" content="noindex" />
```
robinandeer, line 25: hmm should this be added -- I thought it would be controlled from the CMS 🤔

apps/store/public/robots.txt
```diff
@@ -0,0 +1,7 @@
+//robots.txt for hedvig.com
+
+User-agent: *
+Disallow: /cart
+Disallow: /checkout
+Disallow: /confirmation
```
robinandeer, line 6: Does this include all sub pages?

amandaflod: I think so - https://developers.google.com/search/docs/crawling-indexing/robots/robots_txt#url-matching-based-on-path-values

apps/store/src/pages/_document.tsx
amandaflod: I didn't have it to start, what do you think @alebedev?

alebedev:
It's CMS controlled only on CMS pages.  Out of ones listed here only confirmation is a CMS page

robinandeer:
@alebedev @amandaflod yes but if we add a global noindex in `_document` we override the CMS settings 🤔 

Or am I missing something?

apps/store/src/pages/_document.tsx
gustaveen: I wonder if `HeadSeoInfo` with setting from Storyblok would override this, in that case it would be alright.

amandaflod:
Ok, I did not know that! @robinandeer do you think I should remove it completely or add it some other place?

apps/store/src/pages/_document.tsx
guilhermespopolin: I was doing some tests and it seems `meta` tag added by `_document` gets added later. In that case, I assume this change would override the setting for it that cames from CSM, right?!

Screenshot 2023-03-09 at 13 18 44

robinandeer:
* main
  * **PR #1898** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1898" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1898?utm_source=stack-comment).

PR #1903, opened 2023-03-08
GRW-2275 / Lazy load autoplaying videos

gustaveen:
## Describe your changes
Lazy load autoplaying videos

To do this we add the video src to a  `data-src` attribute, observe when it enters the viewport and then load the video.

Review app: https://hedvig-dot-gb1pjdijv-hedvig.vercel.app/se
### Before
**16.9 MB transferred** 🐌

Screenshot 2023-03-08 at 11 10 31 (2)

### After
**66 kB transferred** 🚀

Screenshot 2023-03-08 at 11 10 58 (2)
 

Possible improvement:
We treat autoplaying videos above fold the same (like the product pages). It loads slightly slower but barely noticeable. Let's see if we want to add similar functionality as images for above the fold media to preload the video instead.

Since the `autoplay` attribute on videos override the `preload` attribute, we fetch all autoplaying videos on page load. 

### DIFF
--- a/apps/store/src/components/Video/Video.tsx
+++ b/apps/store/src/components/Video/Video.tsx
@@ -154,15 +154,17 @@ export const Video = ({
         {...autoplayAttributes}
         {...delegated}
       >
-        {sources.map((source) => (
-          // TODO: its adviced to provide the media format type ('type' attribute)
-          // More info http://bitly.ws/y4Jf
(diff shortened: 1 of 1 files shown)


robinandeer (approved):
Very nice! Works great when I tried it

Added via Giphy

PR #1912, opened 2023-03-09
Add combobox component based on downshift

robinandeer:
## Describe your changes

Add combobox component (autocomplete).

This will be needed for breed selector to launch PET.

Based on downshift lib. We don't want to implement all the accessibility stuff ourselves.

I investigated Radix but they don't offer a combobox so we would have to do a lot of wiring ourselves. I also tried out Reach UI but it wasn't working 100% they way we wanted. It has has also not been updated in 6 months which is a red flag.

### DIFF
--- a/apps/store/src/components/Combobox/Combobox.stories.tsx
+++ b/apps/store/src/components/Combobox/Combobox.stories.tsx
@@ -1,0 +1,44 @@
+import { ComponentMeta, StoryFn } from '@storybook/react'
+import { useState } from 'react'
+import { Combobox } from './Combobox'
+
+export default {
+  title: 'Combobox',
+  component: Combobox,
+  argTypes: {},
+} as ComponentMeta<typeof Combobox>
+
+const FRUIT = [
+  {
+    id: '1',
+    name: 'Apple',
+  },
+  {
+    id: '2',
+    name: 'Pear',
+  },
+  {
+    id: '3',
+    name: 'Banana',
+  },
+]
+
+type Fruit = (typeof FRUIT)[number]
+
+export const Default: StoryFn = () => {
+  const [value, setValue] = useState<Fruit | null>(FRUIT[1])
+
+  return (
+    <Combobox
+      placeholder="Search fruit..."
+      items={FRUIT}
+      defaultValue={value}
+      onSelect={setValue}
+      displayValue={(fruit) => fruit.name}
+    />
+  )
+}
+
+export const NoOptionFound: StoryFn = () => {
+  return <Combobox items={['Apple', 'Pear', 'Banana']} defaultValue="Plu" />
+}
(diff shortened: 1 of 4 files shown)


robinandeer:
* main
  * **PR #1912** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1912" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1912?utm_source=stack-comment).

apps/store/src/components/Combobox/Combobox.tsx
```diff
@@ -0,0 +1,209 @@
...
+    onSelectedItemChange({ selectedItem }) {
+      if (selectedItem) {
+        highlight()
+      }
+
+      onSelect?.(selectedItem ?? null)
+    },
+  })
+
+  const filteredItems = items
```
alebedev, line 45: Let's check how we want to search.  Most common options that come to mind
- anywhere in the string (like we have now)
- only at string start
- only at word start

apps/store/src/components/Combobox/Combobox.tsx
```diff
@@ -0,0 +1,209 @@
...
+    .filter((item) => item.toLowerCase().includes(inputValue.toLowerCase()))
+    .slice(0, 5)
```
alebedev, line 47: Nit: should 5 be a constant?

apps/store/src/components/Combobox/Combobox.tsx
```diff
@@ -0,0 +1,209 @@
...
+      if (selectedItem) {
+        highlight()
+      }
+
+      onSelect?.(selectedItem ?? null)
+    },
+  })
+
+  const filteredItems = items
+    .filter((item) => item.toLowerCase().includes(inputValue.toLowerCase()))
```
alebedev, line 46: Nit: depending on performance, it may make sense to use a loop and stop when we get target # of matches.  Filter will always check every element

alebedev (approved):
Very nice that it's library based.  I imagine we'd have to write 10x as much code for our own implementation

Some questions and ideas in the comments

apps/store/src/components/Combobox/Combobox.stories.tsx
```diff
@@ -0,0 +1,26 @@
...
+  argTypes: {},
+} as ComponentMeta<typeof Combobox>
+
+export const Default: StoryFn = () => {
+  const [value, setValue] = useState<string | null>('Pear')
+
+  return (
+    <Combobox
+      placeholder="Search fruit..."
+      items={['Apple', 'Pear', 'Banana']}
```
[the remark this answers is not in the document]
robinandeer: With `useDeferredValue` it seems like performance will not be an issue

PR #1919, opened 2023-03-10
GRW-2346 - feat(Button): support 'open in new window' storyblok feature

guilhermespopolin:
## Describe your changes

* Updates `ButtonBlock` so it supports _open in new window_ storyblok feature

Screenshot 2023-03-10 at 12 31 30

Requested by Peter

## Jira issue(s): [GRW-2346](https://hedvig.atlassian.net/browse/GRW-2346)

[GRW-2346]: https://hedvig.atlassian.net/browse/GRW-2346?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/store/src/blocks/ButtonBlock.tsx
+++ b/apps/store/src/blocks/ButtonBlock.tsx
@@ -21,6 +21,7 @@ export const ButtonBlock = ({ blok, nested }: ButtonBlockProps) => {
         href={getLinkFieldURL(blok.link, blok.text)}
         variant={blok.variant ?? 'primary'}
         size={blok.size ?? 'medium'}
+        target={blok.link.target}
       >
         {blok.text}
       </ButtonNextLink>
--- a/apps/store/src/components/ButtonNextLink.tsx
+++ b/apps/store/src/components/ButtonNextLink.tsx
@@ -5,10 +5,11 @@ import { Button } from 'ui'
 type ButtonProps = ComponentProps<typeof Button>
 
 type Props = LinkProps &
+  Partial<Pick<HTMLAnchorElement, 'target'>> &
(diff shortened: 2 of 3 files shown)


guilhermespopolin:
* main
  * **PR #1919** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1919" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1919?utm_source=stack-comment).

robinandeer (approved):
Nice update!

- Seems like `rel=noopener` is default in modern browsers (https://mathiasbynens.github.io/rel-noopener/) -- but we should keep it

- Never understand if we also need to add `rel=noreferrer` (older browsers) 🤔

guilhermespopolin:
> > Nice update!
> > 
> > * Seems like `rel=noopener` is default in modern browsers (https://mathiasbynens.github.io/rel-noopener/) -- but we should keep it
> > * Never understand if we also need to add `rel=noreferrer` (older browsers) 🤔
> 
> It [seems](https://developer.chrome.com/docs/lighthouse/best-practices/external-anchors-use-rel-noopener/) one can use one **or** another. If I understand correctly, `noreferrer` does what `noopener` does + some extra. It's just about figure it out if we need that extra in order to make a decision between them. I guess in our case, what we want is `noopener` but I'm going to double check with Peter.
> 
> I guess it's fine to keep `noopener` for legacy browsers. But good to know that property is being implicitly set by browsers nowadays :)

It turned out `noreferrer` is what we want.

PR #1950, opened 2023-03-15
Add inline link styles to RichText

gustaveen:
## Describe your changes
- Compose styles a bit cleaner
- Add inline link styling
This styling will be used in more places. So I figure we can import it to whatever blocks/components that need it rather than making it a global style

https://user-images.githubusercontent.com/6661511/225369402-c0ca10bd-f1e9-4319-acf7-5fa1c4f575ae.mov

## Justify why they are needed
Highlight links in text

## Checklist before requesting a review

- [ ] I have performed a self-review of my code

PR #1951, opened 2023-03-15
Randomize hover color on link

gustaveen:
## Describe your changes
Randomize hover color on links. A new color on every page load but the same for all links on the page

https://user-images.githubusercontent.com/6661511/226832633-3a8c3839-649b-485b-82ac-f63eb1bae3bd.mov

[Preview app](https://hedvig-dot-3t8gu3gex-hedvig.vercel.app/api/preview?secret=DWafxGrY6M8&slug=se/hedvig/legal-information-copy&_storyblok=266110806&_storyblok_c=page&_storyblok_version=&_storyblok_lang=default&_storyblok_release=0&_storyblok_rl=1679470530315&_storyblok_tk[space_id]=165473&_storyblok_tk[timestamp]=1679470530&_storyblok_tk[token]=b2221ff34357f849bc92b527a80b33d9f9bdb387)

Nice touch, see figma [spec ](https://www.figma.com/file/qUhLjrKl98PAzHov9ilaDH/New-Web-UI-Kit?node-id=2035%3A12007&t=dOuORAt9hzZlM6Xe-1)

### DIFF
--- a/apps/store/src/blocks/RichTextBlock/RichTextBlock.styles.tsx
+++ b/apps/store/src/blocks/RichTextBlock/RichTextBlock.styles.tsx
@@ -1,6 +1,47 @@
-import { css } from '@emotion/react'
+import { css, Global } from '@emotion/react'
 import { getHeadingVariantStyles } from 'ui/src/components/Heading/Heading.helpers'
 import { theme, mq } from 'ui'
+
+const hoverColors = [
+  theme.colors.blue600,
+  theme.colors.green600,
+  theme.colors.amber600,
+  theme.colors.red600,
+  theme.colors.pink600,
+  theme.colors.purple600,
+  theme.colors.teal600,
+]
+
+const getRandomHoverColor = () => {
+  // Random number between 0-6
+  const randomNumber = Math.floor(Math.random() * hoverColors.length)
+  return hoverColors[randomNumber]
+}
+
+export const GlobalLinkStyles = () => {
+  const color = getRandomHoverColor()
+
+  return <Global styles={{ ':root': { '--random-hover-color': color } }} />
+}
+
+export const linkStyles = css({
+  a: {
+    position: 'relative',
+    '&::after': {
+      content: '""',
+      position: 'absolute',
+      bottom: -2,
+      left: 0,
+      width: '100%',
+      height: 2,
+      backgroundColor: theme.colors.borderOpaque,
+    },
+
+    '&:hover::after': {
+      backgroundColor: 'var(--random-hover-color)',
+    },
+  },
+})
 
 export const listStyles = css({
   ul: {
@@ -80,25 +121,6 @@ export const listStyles = css({
 
     ol: {
       marginLeft: 0,
-    },
-  },
-})
-
-export const linkStyles = css({
-  a: {
-    position: 'relative',
-    '&::after': {
-      content: '""',
-      position: 'absolute',
-      bottom: -1,
-      left: 0,
-      width: '100%',
-      height: 2,
-      backgroundColor: theme.colors.borderOpaque,
-    },
-
-    '&:hover::after': {
-      backgroundColor: theme.colors.blue600,
     },
   },
 })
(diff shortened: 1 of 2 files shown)


gustaveen:
* main
  * **PR #1950** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1950" target="_blank">Graphite</a> 
    * **PR #1951** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1951" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1951?utm_source=stack-comment).

apps/store/src/blocks/RichTextBlock/RichTextBlock.styles.tsx
```diff
@@ -2,6 +2,22 @@ import { css } from '@emotion/react'
...
+  theme.colors.blue600,
+  theme.colors.green600,
+  theme.colors.amber600,
+  theme.colors.red600,
+  theme.colors.pink600,
+  theme.colors.purple600,
+  theme.colors.teal600,
+]
+
+const getRandomHoverColor = () => {
```
alebedev, line 15: Let's try to find a way to solve hydration warning, looks annoying and potentially unsafe.  As an idea, try getting pseudo-random value based on `useId` hook

robinandeer (commented):
Not that we are aiming for the best accessibility but do we know if this impacts it in any meaningful way?

apps/store/src/blocks/RichTextBlock/RichTextBlock.styles.tsx
gustaveen: Makes sense, I'll I'll have a look at that 👍 @alebedev

gustaveen:
> Not that we are aiming for the best accessibility but do we know if this impacts it in any meaningful way?

Do you mean the link styling in general or the color choices? @robinandeer Removed one color that had bad contrast at least

robinandeer:
> > Not that we are aiming for the best accessibility but do we know if this impacts it in any meaningful way?
> 
> Do you mean the link styling in general or the color choices? @robinandeer Removed one color that had bad contrast at least

Mostly I meant the random colors, it's not really a pattern I've seen before. It's a good idea to stick to behaviors that users expect unless we have good reason to deviate of course :)

gustaveen:
> Mostly I meant the random colors, it's not really a pattern I've seen before. It's a good idea to stick to behaviors that users expect unless we have good reason to deviate of course :)

I see. As long as the underline color is the same (in the non-hover state) I think user will recognise it as links. And it will be the same link color on a page so it won't differ between links on a page. But let's evaluate for sure @robinandeer

apps/store/src/blocks/RichTextBlock/RichTextBlock.styles.tsx
```diff
@@ -1,7 +1,52 @@
...
+  // Random number between 0-6
+  const randomNumber = Math.floor(Math.random() * hoverColors.length)
+  return hoverColors[randomNumber]
+}
+
+export const GlobalLinkStyles = () => {
+  const [color, setColor] = useState(getRandomHoverColor)
+  useEffect(() => {
+    setColor(getRandomHoverColor())
+  }, [])
```
robinandeer, line 26: What are we trying to accomplish here?

We first set the initial state and then update it once after we mount this component.

What's the practical difference of skipping the `useEffect`?

robinandeer (approved):
Let's go :)

Have we verified that all colors have enough contrast?

apps/store/src/blocks/RichTextBlock/RichTextBlock.styles.tsx
gustaveen: > Let's go :)
> 
> Have we verified that all colors have enough contrast?

I removed the yellow one since it had poor contrast. I contrast checked the others and some fail for text on background checks. I don't think that applies the same on a 2px border though. But I suppose it all comes down to if we want to be fully WCAG compatible 🤷‍♂️ @robinandeer

PR #1959, opened 2023-03-16
Add livingSpace field to pet form

gustaveen:
Needed to get a price


apps/store/src/services/PriceCalculator/data/SE_PET_CAT.ts
```diff
@@ -102,6 +102,6 @@ export const SE_PET_CAT: Template = {
         },
       ],
     },
-    yourAddressSection,
+    yourAddressSectionWithlivingSpace,
```
gustaveen, line 105: Not sure what you feel about these formFractions, `yourAddressSection`  will only be used by Car now

robinandeer: hmm in that case I would inline it but feel free to do later

robinandeer (approved):
neat

PR #1961, opened 2023-03-16
GRW-2354-feat: added PetBreedField

guilhermespopolin:
## Describe your changes

* Adds loading UI form `Combobox`
* Added PetBreedInput

https://user-images.githubusercontent.com/19200662/225735575-6a08a023-aaa0-48a9-a3be-02f67655f7e5.mov

<br/>

Testing Links:

* [🐕  Dog Insurance](https://hedvig-dot-com-git-grw-2354-featpet-breed-field-hedvig.vercel.app/se/forsakringar/djurforsakring/hundforsakring)
* [🐱 Cat Insurance](https://hedvig-dot-com-git-grw-2354-featpet-breed-field-hedvig.vercel.app/se/forsakringar/djurforsakring/kattforsakring) 

**Disclaimer 1**: Please note this doesn't support mixed breed for dogs yet.
**Disclaimer 2**: It's not possible to create a offer with this setup yet because BE expects a different field for breeds: `breed` instead of `breeds`. However, this will change. More info [here](https://hedviginsurance.slack.com/archives/C04SUH8E61K/p1678982160603139).

Requerid for _Pet Insurance_

## Jira issue(s): [GRW-2354](https://hedvig.atlassian.net/browse/GRW-2354)

[GRW-2354]: https://hedvig.atlassian.net/browse/GRW-2354?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/store/src/components/Combobox/Combobox.tsx
+++ b/apps/store/src/components/Combobox/Combobox.tsx
@@ -27,14 +27,14 @@ export const Combobox = <Item,>({
   selectedItem,
   onSelectedItemChange,
   defaultSelectedItem,
-  displayValue,
+  displayValue = (item) => String(item),
   ...externalInputProps
 }: Props<Item>) => {
   const { highlight, animationProps } = useHighlightAnimation()
 
   const [inputValue, setInputValue] = useState(() => {
     if (defaultSelectedItem) {
-      return displayValue?.(defaultSelectedItem) ?? String(defaultSelectedItem)
+      return displayValue(defaultSelectedItem)
     }
 
     return ''
@@ -85,7 +85,7 @@ export const Combobox = <Item,>({
             // Select on [Enter] when only one item is available for selection
             return {
               ...changes,
-              inputValue: displayValue?.(filteredItems[0]) ?? String(filteredItems[0]) ?? '',
+              inputValue: displayValue(filteredItems[0]),
               selectedItem: filteredItems[0],
             }
           }
@@ -95,7 +95,7 @@ export const Combobox = <Item,>({
       }
     },
     itemToString(item) {
-      return item ? displayValue?.(item) ?? String(item) : ''
+      return item ? displayValue(item) : ''
     },
   })
 
@@ -136,7 +136,7 @@ export const Combobox = <Item,>({
                 {...getItemProps({ item, index })}
                 data-highlighted={highlightedIndex === index}
               >
-                {displayValue?.(item) ?? String(item)}
+                {displayValue(item)}
               </ComboboxOption>
             </Fragment>
           ))}
--- a/apps/store/src/components/Combobox/Combobox.tsx
+++ b/apps/store/src/components/Combobox/Combobox.tsx
@@ -2,6 +2,7 @@ import styled from '@emotion/styled'
 import { useCombobox } from 'downshift'
 import { motion } from 'framer-motion'
 import { Fragment, useState, useMemo, useDeferredValue } from 'react'
+import { DotPulse } from 'ui/src/components/Button/DotPulse'
 import { ChevronIcon, CrossIconSmall, Text, theme, WarningTriangleIcon } from 'ui'
 import { useHighlightAnimation } from '@/utils/useHighlightAnimation'
 
@@ -15,6 +16,7 @@ type Props<Item> = {
   displayValue?: (item: Item) => string
   disabled?: boolean
   required?: boolean
+  loading?: boolean
 }
 
 /**
@@ -26,6 +28,7 @@ export const Combobox = <Item,>({
   selectedItem,
   onSelectedItemChange,
   displayValue,
+  loading,
   ...externalInputProps
 }: Props<Item>) => {
   const { highlight, animationProps } = useHighlightAnimation()
@@ -107,14 +110,25 @@ export const Combobox = <Item,>({
           {...externalInputProps}
           data-expanded={isOpen}
           data-warning={noOptions}
+          disabled={loading}
         />
         <Actions>
-          <DeleteButton onClick={handleClickDelete} hidden={inputValue.length === 0}>
-            <CrossIconSmall />
-          </DeleteButton>
-          <ToggleButton {...getToggleButtonProps()} data-warning={noOptions}>
-            <ChevronIcon size="1rem" />
-          </ToggleButton>
+          {loading ? (
+            <DotPulse />
+          ) : (
+            <>
+              <DeleteButton
+                type="button"
+                onClick={handleClickDelete}
+                hidden={inputValue.length === 0}
+              >
+                <CrossIconSmall />
+              </DeleteButton>
+              <ToggleButton type="button" {...getToggleButtonProps()} data-warning={noOptions}>
+                <ChevronIcon size="1rem" />
+              </ToggleButton>
+            </>
+          )}
         </Actions>
       </InputWrapper>
 
--- a/apps/store/src/components/Combobox/Combobox.tsx
+++ b/apps/store/src/components/Combobox/Combobox.tsx
@@ -9,8 +9,9 @@ const ITEMS_TO_SHOW = 5
 
 type Props<Item> = {
   items: Array<Item>
-  selectedItem: Item | null
-  onSelectedItemChange: (item: Item | null) => void
+  selectedItem?: Item | null
+  onSelectedItemChange?: (item: Item | null) => void
+  defaultSelectedItem?: Item | null
   placeholder?: string
   displayValue?: (item: Item) => string
   disabled?: boolean
@@ -25,12 +26,19 @@ export const Combobox = <Item,>({
   items,
   selectedItem,
   onSelectedItemChange,
+  defaultSelectedItem,
   displayValue,
   ...externalInputProps
 }: Props<Item>) => {
   const { highlight, animationProps } = useHighlightAnimation()
 
-  const [inputValue, setInputValue] = useState('')
+  const [inputValue, setInputValue] = useState(() => {
+    if (defaultSelectedItem) {
+      return displayValue?.(defaultSelectedItem) ?? String(defaultSelectedItem)
+    }
+
+    return ''
+  })
   const deferredInputValue = useDeferredValue(inputValue)
   const filteredItems = useMemo(() => {
     return filterItems(items, deferredInputValue, displayValue)
@@ -50,8 +58,9 @@ export const Combobox = <Item,>({
   } = useCombobox({
     items: filteredItems,
     selectedItem,
+    defaultSelectedItem,
     onSelectedItemChange({ selectedItem }) {
-      onSelectedItemChange(selectedItem ?? null)
+      onSelectedItemChange?.(selectedItem ?? null)
 
       if (selectedItem) {
         highlight()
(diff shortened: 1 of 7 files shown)


guilhermespopolin:
* main
  * **PR #1961** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1961" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1961?utm_source=stack-comment).

apps/store/src/components/Combobox/Combobox.tsx
```diff
@@ -107,14 +110,25 @@ export const Combobox = <Item,>({
...
         />
         <Actions>
-          <DeleteButton onClick={handleClickDelete} hidden={inputValue.length === 0}>
-            <CrossIconSmall />
-          </DeleteButton>
-          <ToggleButton {...getToggleButtonProps()} data-warning={noOptions}>
-            <ChevronIcon size="1rem" />
-          </ToggleButton>
+          {loading ? (
+            <DotPulse />
```
gustaveen, line 117: nit: I wonder if we should have the loader centered like on buttons but maybe that interferes with the label

gustaveen (approved):
Great stuff! 🙌

apps/store/src/components/Combobox/Combobox.tsx
guilhermespopolin: Yeah, I forgot to write that this is kind of a placeholder solution while I check with design team what do they think loading UI should be,

apps/store/src/components/PriceCalculator/PetBreedField.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+type PetBreedFieldProps = {
+  field: InputFieldPetBreed
+  loading?: boolean
+}
+
+export const PetBreedField = ({ field }: PetBreedFieldProps) => {
+  const translateLabel = useTranslateFieldLabel()
+
+  const { data, loading: isLoadingAvailableBreeds } = usePriceIntentAvailableBreedsQuery({
```
guilhermespopolin, line 20: In a perfect would that data could be fetched server side, however I couldn't think in a way to make it work. Any ideas?

apps/store/src/components/PriceCalculator/PetBreedField.tsx
```diff
@@ -0,0 +1,55 @@
...
+      (data?.priceIntentAvailableBreeds ?? []).map(({ id, displayName }) => ({
+        id,
+        displayName,
+      })),
+    [data?.priceIntentAvailableBreeds],
+  )
+
+  const [selectedBreed, setSelectedBreed] = useState<Breed | null>(
+    () => availableBreeds.find((breed) => breed.id === field.value) ?? null,
+  )
```
guilhermespopolin, line 37: 🤔 this is not going to work. I need to make sure to call that _get initial state_ function when `availableBreeds` list is populated...

apps/store/src/components/PriceCalculator/PetBreedField.tsx
robinandeer: Hmm not sure we really need a loading state 🤔 just show an error in case we are unable to fetch the data maybe?

I was considering running this query during build time and just passing the data in context but I think we also should start with something simple like this 🙂

apps/store/src/components/Combobox/Combobox.tsx
```diff
@@ -2,6 +2,7 @@ import styled from '@emotion/styled'
 import { useCombobox } from 'downshift'
 import { motion } from 'framer-motion'
 import { Fragment, useState, useMemo, useDeferredValue } from 'react'
+import { DotPulse } from 'ui/src/components/Button/DotPulse'
```
robinandeer, line 5: strange import -- should be from "ui" no?

apps/store/src/components/Combobox/Combobox.tsx
robinandeer: I would skip this loader -- perhaps just disable the field until we have data

apps/store/src/components/PriceCalculator/AutomaticField.tsx
```diff
@@ -149,6 +150,9 @@ export const AutomaticField = ({ field, priceIntent, onSubmit, loading, autoFocu
         />
       )
 
+    case 'pet-breed':
```
robinandeer, line 153: hmm... however, cat and dog will work differently, no?

I would consider naming them:

- breeds-dog
- breeds-cat (I think it's still "breeds" for cat)

apps/store/src/components/PriceCalculator/PetBreedField.tsx
robinandeer: hmm why do we need to keep this state?? We should just send the request to the API, no?

It needs to be formatted as an array anyways so this needs to be a custom field type 🙂

apps/store/src/services/PriceCalculator/data/SE_PET_CAT.ts
```diff
@@ -100,6 +91,16 @@ export const SE_PET_CAT: Template = {
           },
           layout: LAYOUT.FULL_WIDTH,
         },
+        {
+          field: {
+            type: 'pet-breed',
+            name: 'breed',
```
robinandeer, line 97: this should be "breeds", no?

apps/store/src/services/PriceCalculator/data/SE_PET_DOG.ts
```diff
@@ -85,6 +76,16 @@ export const SE_PET_DOG: Template = {
           },
           layout: LAYOUT.FULL_WIDTH,
         },
+        {
+          field: {
+            type: 'pet-breed',
+            name: 'breed',
```
robinandeer, line 82: "breeds"

apps/store/src/services/PriceCalculator/data/SE_PET_CAT.ts
guilhermespopolin: Yes, correct.

apps/store/src/components/PriceCalculator/AutomaticField.tsx
guilhermespopolin: Hum...the way I understood it based on [this](https://hedviginsurance.slack.com/archives/C04SUH8E61K/p1678982160603139) cat and dogs should send a list of breeds, the difference is that we'd be eventually sending more than one item for dogs. Isn't that the case?

apps/store/src/components/PriceCalculator/PetBreedField.tsx
guilhermespopolin: You mean get `onSubmit` function `PetBreedField` and call it with a cb given to `Combobox`'s `onSelectedItemChange`?

apps/store/src/components/PriceCalculator/AutomaticField.tsx
guilhermespopolin: Oh now I get it. What I thought was keep like that for this iteration and differentiate then when adding support for dogs mixed breed.

apps/store/src/components/PriceCalculator/PetBreedField.tsx
```diff
@@ -0,0 +1,60 @@
...
+  const handleSelectBreed = async (selectedBreed: Breed | null) => {
+    await onSubmit({
+      [field.name]: selectedBreed ? [selectedBreed.id] : [],
+    })
+  }
+
+  if (error) {
+    // TODO better handle error
+    return <span>Unable to fetch breeds.</span>
+  }
```
guilhermespopolin, line 45: Any suggestions on what we should here?

robinandeer: I think this is a good use case for error boundary -- so we should just throw an error in this component

then we add an error boundary around the whole purchase form!

apps/store/src/services/PriceCalculator/data/SE_PET_DOG.ts
```diff
@@ -86,6 +77,16 @@ export const SE_PET_DOG: Template = {
           },
           layout: LAYOUT.FULL_WIDTH,
         },
+        {
```
robinandeer, line 80: According to design -- this field should be rendered under the "name", no?

apps/store/src/components/PriceCalculator/PetBreedField.tsx
```diff
@@ -0,0 +1,60 @@
...
+        displayName,
+      })),
+    [data?.priceIntentAvailableBreeds],
+  )
+  let defaultSelectedItem: Breed | null = null
+  if (field.value) {
+    defaultSelectedItem = availableBreeds.find((breed) => breed.id === field.value?.[0]) ?? null
+  }
+
+  const handleSelectBreed = async (selectedBreed: Breed | null) => {
```
robinandeer, line 36: nit: skip making async/await unless we plan to catch error etc.

apps/store/src/components/PriceCalculator/PetBreedField.tsx
```diff
@@ -0,0 +1,60 @@
...
+      (data?.priceIntentAvailableBreeds ?? []).map(({ id, displayName }) => ({
+        id,
+        displayName,
+      })),
+    [data?.priceIntentAvailableBreeds],
+  )
+  let defaultSelectedItem: Breed | null = null
+  if (field.value) {
+    defaultSelectedItem = availableBreeds.find((breed) => breed.id === field.value?.[0]) ?? null
+  }
```
robinandeer, line 34: nit: comment that this will be removed when we handle multiple breeds

apps/store/src/components/PriceCalculator/PetBreedField.tsx
```diff
@@ -0,0 +1,60 @@
...
+  })
+
+  const availableBreeds: Array<Breed> = useMemo(
+    () =>
+      (data?.priceIntentAvailableBreeds ?? []).map(({ id, displayName }) => ({
+        id,
+        displayName,
+      })),
+    [data?.priceIntentAvailableBreeds],
+  )
```
robinandeer, line 30: nit: not needed, just do `availableBreeds = data?.priceIntentAvailableBreeds ?? []`, no?

apps/store/src/components/Combobox/Combobox.tsx
```diff
@@ -9,8 +9,9 @@ const ITEMS_TO_SHOW = 5
 
 type Props<Item> = {
   items: Array<Item>
-  selectedItem: Item | null
-  onSelectedItemChange: (item: Item | null) => void
+  selectedItem?: Item | null
+  onSelectedItemChange?: (item: Item | null) => void
```
robinandeer, line 13: why not required?

robinandeer (approved):
I think it looks good -- just some refactor ideas in comments

apps/store/src/components/PriceCalculator/PetBreedField.tsx
guilhermespopolin: Yeah, `PrinceIntentAnimalBreed` can be assigned to `Breed`. Just thought it would be good to keep the exact same type. Gonna change it.

apps/store/src/components/Combobox/Combobox.tsx
guilhermespopolin: Because if they become required it would be impossible to use `defaultSelectedItem` since it would also be required to provide `selectedItem` which then, would always override `defaultSelectedItem`,

robinandeer: hmm good point, although for our use case both those would be required - well well

robinandeer (approved):
Great work! Let's try it out :D

PR #1991, opened 2023-03-20
refact(InsuranceCard): general updates on onboarding's Landing Page (old platform)

guilhermespopolin:
## Describe your changes

* General
  * Removes `merge-refs` implementation
* `ui/Checkox`
  * Removes usage of `merge-refs`.
  * Expose an imperative API that can be used to perform some actions via `ref` - an imperative way of triggering `click`.
  * Removes invalidation related code: that doesn't belong here.
  * Move styled components to the end of the file
* `LandingPage/InsuranceCard` 
  * Removes unecessary `isSelected` state. The issue it solves is now being handle strictly on CSS with `has` selector. That might be the most controversial update.
  * Removes `moreDetailsButtonRef` and make sure each interactive control inside `InsuranceCard` doesn't affect each other while clicking them with `stopPropagation()`
* `LandingPage/LandingPage`
  * Implement custom form validation  

This is not strictly needed and it will be removed when we fully migrate to the new platform. 
The motivation behind it is that I was studying some stuff this weekend and I used this PR as an exercise. We can merge it if we agree it's worth it.


guilhermespopolin:
* main
  * **PR #1991** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1991" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/1991?utm_source=stack-comment).

robinandeer (approved):
Seems to work well 🙂

packages/ui/src/components/Checkbox/Checkbox.tsx
```diff
@@ -1,16 +1,65 @@
...
+    useImperativeHandle(
+      ref,
+      () => ({
+        click: () => inputRef.current?.click(),
+      }),
+      [],
+    )
+
+    let LabelComponent: ReactNode = null
+    if (label) {
```
alebedev, line 37: If you're extracting condition from JSX, you can move `prependLabel` check here as well

alebedev (approved):
Looks good to me

packages/ui/src/components/Checkbox/Checkbox.tsx
guilhermespopolin: 🤔 but they're conditions that doesn't correlate. This one is about "building" or not the `LabelComponent`. The `prependLabel` one is about putting the label after of before the checkbox.

packages/ui/src/components/Checkbox/Checkbox.tsx
```diff
@@ -1,16 +1,65 @@
...
   circle?: boolean
-  errorMessage?: string
 }
 
+export const Checkbox = forwardRef<CheckboxImperativeAPI, CheckboxProps>(
+  ({ circle, prependLabel, label, ...delegated }, ref) => {
+    const inputRef = useRef<HTMLInputElement>(null)
+    const id = useId()
+
+    useImperativeHandle(
```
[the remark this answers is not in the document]
guilhermespopolin: You're right. It made sense before when I also needed to keep a ref to the input internally because of this [`effect`](https://github.com/HedvigInsurance/racoon/pull/1991/files#diff-e318e4aaceae2e338ef857d2f05afdfcba891a2d7bf49250d860f46d5181a60eL111). But since now no validation code is placed inside `Checkbox` there's no need to also keep a reference to it internally, and therefore, make usage of `useImperativeHandle`. Nice catch.

PR #2010, opened 2023-03-21
Bug: keep track of latest selected offer on PDP

robinandeer:
## Describe your changes

Refresh selected offer after editing price intent/offer (start date or re-confirm).

We have a bug now where you don't see the updated start date after you edit. This is because we store the selected offer in local state and don't update it after editing. This PR fixes that.

### DIFF
--- a/apps/store/src/components/ProductPage/PurchaseForm/OfferPresenter.tsx
+++ b/apps/store/src/components/ProductPage/PurchaseForm/OfferPresenter.tsx
@@ -2,7 +2,7 @@ import { datadogLogs } from '@datadog/browser-logs'
 import styled from '@emotion/styled'
 import { useInView } from 'framer-motion'
 import { useTranslation } from 'next-i18next'
-import { RefObject, useEffect, useMemo, useRef, useState } from 'react'
+import { RefObject, useEffect, useMemo, useRef } from 'react'
 import { Button, Space, Text, theme } from 'ui'
 import { useUpdateCancellation } from '@/components/ProductPage/PurchaseForm/useUpdateCancellation'
 import { useUpdateStartDate } from '@/components/ProductPage/PurchaseForm/useUpdateStartDate'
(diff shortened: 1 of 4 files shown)


robinandeer:
* main
  * **PR #2010** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/2010" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/2010?utm_source=stack-comment).

apps/store/src/components/ProductPage/PurchaseForm/OfferPresenter.tsx
```diff
@@ -37,9 +39,10 @@ type Props = {
 
 export const OfferPresenter = (props: Props) => {
   const { priceIntent, shopSession, scrollPastRef, onAddedToCart, onClickEdit } = props
+  const { selectedOffer } = props
```
alebedev, line 42: Maybe we should read it from atom instead of passing via props?

alebedev (approved):
Looks good

apps/store/src/components/ProductPage/PurchaseForm/OfferPresenter.tsx
robinandeer: yeah but then we need to deal with a nullable value

the offer presenter should never be rendered with without a selected offer so I chose to pass it in instead

makes sense?

apps/store/src/components/ProductPage/PurchaseForm/OfferPresenter.tsx
```diff
@@ -37,9 +39,10 @@ type Props = {
...
+  const [, setSelectedOffer] = useSelectedOffer()
```
guilhermespopolin, line 43: Woudn't be the case of using Jotai's `useAtomValue` here?
I understand we're using Jotai via a custom hook..maybe we should export a read only option from `useSelectedOffer` as well? 🤔

guilhermespopolin (approved):

PR #2013, opened 2023-03-21
Add page transition progress indicator

robinandeer:
## Describe your changes

Show the progress indicator on the top of the page during page transitions.

To make the page transitions more smooth.


alebedev (approved):
Great visual improvement!

apps/store/src/services/nprogress/nprogress.tsx
```diff
@@ -0,0 +1,57 @@
...
+    router.events.on('routeChangeError', handleRouteChangeComplete)
+
+    return () => {
+      router.events.off('routeChangeStart', handleRouteStart)
+      router.events.off('routeChangeComplete', handleRouteChangeComplete)
+      router.events.off('routeChangeError', handleRouteChangeComplete)
+    }
+  }, [router.events])
+
+  return <Global styles={[styles, { ':root': { '--nprogress-color': color } }]} />
```
gustaveen, line 34: Oh smart to set it as a css variable to avoid hydration errors! 💡 Will use it for the links 🙌

gustaveen (approved):
Stylish! 💅

apps/store/src/services/nprogress/nprogress.tsx
robinandeer: Yeah, it's too bad that emotion doesn't make them easier to use but this does seem like a good implementation :)

PR #2055, opened 2023-03-28
Add API route to exchange authorization code

robinandeer:
## Describe your changes

Add API route to exchange authorization code for access token and create authenticated session.

Refactor "reset session" API route and add utility function to clear session.

We will use this new endpoint to support the Payment Link feature in HOPE.


guilhermespopolin (approved):

apps/store/src/pages/api/session/reset.ts
```diff
@@ -1,38 +1,19 @@
...
-  const { query } = req
+  await resetSessionServerSide(req, res)
 
+  const nextQueryParam = req.query['next']
   const nextURL = new URL(ORIGIN_URL)
-  nextURL.pathname = PageLink.home()
-
-  const nextQueryParam = query['next']
   if (typeof nextQueryParam === 'string') {
     nextURL.pathname = nextQueryParam
```
guilhermespopolin, line 14: I've experienced a bug recently while assign a string to `URL.pathname`: If `nextQueryParam` end up being a url that contains query parameters like `se/insurance?openPriceCalculator`, that url string will be encoded when assigned to `pathname`. Eg:

```javascript
let url1 = new URL('/api/campaign?next=/se?openPriceCalculator=1', 'http://localhost:8080')
let url2 = new URL('http://localhost:8080')

url2.pathname = url.searchParams.get('next')
console.log(url2.pathname) // '/se%3FopenPriceCalculator=1'
```

Not exactly part of this PR but maybe it's worth to handle that case here as well in order to avoid malformed urls

guilhermespopolin (approved):
Well done. Just left a comment!

apps/store/src/pages/api/session/reset.ts
robinandeer: Instead of nesting query params we should just forward any query params from the original URL 🙂

guilhermespopolin (approved):

PR #2058, opened 2023-03-28
Avoid resize error in media query hook

robinandeer:
## Describe your changes

Wrap resize callback in request animation frame.

Source: https://stackoverflow.com/questions/49384120/resizeobserver-loop-limit-exceeded/58701523#58701523

Prevent resize error:

```
ResizeObserver loop limit exceeded
```


guilhermespopolin (approved):
Awesome 🙌🏻

PR #2101, opened 2023-04-03
Improve ProductCard SEO and accessibility

gustaveen:
## Describe your changes
Currently we have two links in the same ProductCard which we should avoid if possible for SEO reasons. Having the "Read more" button as the main link is also bad practise for screenreaders and SEO since it's the same on all cards.

- Make the title of the Card the main link to get a descriptive link text
- Add an `:after` element to the link to cover the card and make it clickable
- Add a focus state to "Read More" button when the main link is focused
- Navigate with router to the product page when "Read More" button is clicked

https://user-images.githubusercontent.com/6661511/229752222-35f818b8-d9aa-41a2-ba36-030989ec85b9.mov

Improve ProductCard SEO and accessibility

Good resource I found on the topic https://inclusive-components.design/cards/

### DIFF
--- a/apps/store/src/components/ProductCard/ProductCard.tsx
+++ b/apps/store/src/components/ProductCard/ProductCard.tsx
@@ -1,8 +1,9 @@
 import { UrlObject } from 'url'
 import styled from '@emotion/styled'
 import Link from 'next/link'
+import { useRouter } from 'next/router'
 import { useTranslation } from 'react-i18next'
-import { mq, Space, theme } from 'ui'
+import { Button, mq, Space, theme } from 'ui'
 import { ImageSize } from '@/blocks/ProductCardBlock'
 import { ButtonNextLink } from '@/components/ButtonNextLink'
 import { ImageWithPlaceholder } from '@/components/ImageWithPlaceholder/ImageWithPlaceholder'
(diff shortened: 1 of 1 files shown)


gustaveen:
* main
  * **PR #2101** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/2101" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/2101?utm_source=stack-comment).

apps/store/src/components/ProductCard/ProductCard.tsx
```diff
@@ -32,30 +33,35 @@ export const ProductCard = ({
...
+            aria-hidden={true}
+            size="medium"
+            variant="secondary"
+          >
+            {t('READ_MORE')}
+          </Button>
+          {priceLink && (
+            <ButtonNextLink href={priceLink} size="medium" variant="primary-alt">
+              {t('GET_PRICE_LINK')}
+            </ButtonNextLink>
```
robinandeer, line 60: Is it a problem that this link is also nested?

apps/store/src/components/ProductCard/ProductCard.tsx
```diff
@@ -32,30 +33,35 @@ export const ProductCard = ({
...
+        <Image {...imageProps} alt={alt} fill sizes="100vw" />
+      </ImageWrapper>
+      <ContentWrapper>
+        <MainLink href={link.url}>{title}</MainLink>
+        <Subtitle>{subtitle}</Subtitle>
+        <CallToAction>
+          <Button
+            onClick={() => router.push(link.url)}
+            tabIndex={-1}
+            aria-hidden={true}
```
robinandeer, line 51: So we can't just add aria-hidden to the link component? That would solve the a11y concerns tho, right? But not SEO issues...?

robinandeer (approved):
Nice research and clean up -- I learned something new 🆕

apps/store/src/components/ProductCard/ProductCard.tsx
gustaveen: Right! That's my understanding at least. @robinandeer  Peter recommended we keep duplicated links to a minimum if possible. But as always with SEO it's no clear right or wrong 🤣 So let's see how this works or if we change it back ro a regular link, this is the part where I'm a bit unsure.

apps/store/src/components/ProductCard/ProductCard.tsx
gustaveen: It should be fine as long as it isn't nested inside another link 👍

PR #2141, opened 2023-04-11
Integrate C1 (customer support) script

robinandeer:
## Describe your changes

Integrate customer first (C1) script globally on the site.

Only show widget on-demand.

Remove intercom stuff.

We wanna test this out in staging to begin with.


alebedev (approved):
Looks good.  I don't think we need separate feature flag when we can just set script URL to empty

guilhermespopolin (approved):

apps/store/src/services/CustomerFirst.tsx
```diff
@@ -0,0 +1,43 @@
...
+  )
+}
+
+export const useCustomerFirst = () => {
+  const setOpen = useSetAtom(OPEN_ATOM)
+
+  useEffect(() => {
+    const handleMessage = ({ data }: MessageEvent<MessageData>) => {
+      if (data === 'showWidget') setOpen(true)
+      else if (data === 'hideWidget') setOpen(false)
```
alebedev, line 29: Do we get `hideWidget` from C1?

gustaveen (approved):
Nice! 🙌 
Should we try to build our own floating button instead to make it easier to control? :) 
Saw that clicking the "Telefonnummer" button in the widget goes to this [page](https://new.hedvig.com/se/support/inte-medlem), is that intentional?

robinandeer:
> Nice! 🙌 
> Should we try to build our own floating button instead to make it easier to control? :) 
> Saw that clicking the "Telefonnummer" button in the widget goes to this [page](https://new.hedvig.com/se/support/inte-medlem), is that intentional? 

Hmm that's strange --- not sure, but hopefully Malin is on top of that

We could maybe build our own button but I think we can influence how it looks but discussing with them to begin with.

PR #2152, opened 2023-04-13
Add ability to replace a cart item

robinandeer:
## Describe your changes

Allow users to replace an entry in the cart.

Pick up the offer ID from the URL and use it to replace the entry in the cart.

Also use it to prefill some data in the form.

To be able to add an "edit" button in the cart.

### DIFF
--- a/apps/store/src/components/CartPage/useEditProductOffer.ts
+++ b/apps/store/src/components/CartPage/useEditProductOffer.ts
@@ -1,0 +1,70 @@
+import { useApolloClient } from '@apollo/client'
+import { datadogLogs } from '@datadog/browser-logs'
+import { useRouter } from 'next/router'
+import { useState } from 'react'
+import { useProductMetadata } from '@/components/LayoutWithMenu/ProductMetadataContext'
+import { OPEN_PRICE_CALCULATOR_QUERY_PARAM } from '@/components/ProductPage/PurchaseForm/useOpenPriceCalculatorQueryParam'
+import { fetchPriceTemplate } from '@/services/PriceCalculator/PriceCalculator.helpers'
+import { priceIntentServiceInitClientSide } from '@/services/priceIntent/PriceIntentService'
+
+type State = 'idle' | 'loading' | 'error'
+
+type Params = {
+  shopSessionId: string
+  offerId: string
+  data: any
+  productName: string
+}
+
+export const useEditProductOffer = () => {
+  const [state, setState] = useState<State>('idle')
+  const apolloClient = useApolloClient()
+  const products = useProductMetadata()
+  const router = useRouter()
+
+  const editProductOffer = async (params: Params) => {
+    setState('loading')
+    // TODO: Check if product offer is connected to price intent -- if not, create new one
+
+    const product = products?.find((item) => item.name === params.productName)
+    if (!product) throw new Error('Product not found')
+
+    const priceIntentService = priceIntentServiceInitClientSide(apolloClient)
+    const priceTemplate = fetchPriceTemplate(params.productName)
+    if (!priceTemplate) {
+      setState('error')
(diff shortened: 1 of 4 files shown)


apps/store/src/components/CartPage/useEditProductOffer.ts
```diff
@@ -0,0 +1,77 @@
...
+  const apolloClient = useApolloClient()
+  const products = useProductMetadata()
+  const router = useRouter()
+
+  const editProductOffer = async (params: Params) => {
+    setState('loading')
+    // TODO: Check if product offer is connected to price intent -- if not, create new one
+
+    const product = products?.find((item) => item.name === params.productName)
+    if (!product) {
```
alebedev, line 30: Nit: if you throw here instead, control flow would be more robust.  Currently if programmer error happens before `try` on line 47, it won't be handled

apps/store/src/components/ProductPage/PriceIntentContext.tsx
```diff
@@ -40,19 +42,26 @@ const usePriceIntentContextValue = () => {
...
-        )
+        if (prev) {
+          const matchingOffer = data.priceIntent.offers.find((item) => compareOffer(item, prev))
+          if (matchingOffer) return matchingOffer
+        }
 
-        if (matchingOffer) return matchingOffer
+        if (entryToReplace) {
+          const matchingReplaceOffer = data.priceIntent.offers.find(
+            (item) =>
```
alebedev, line 59: Nit: can we use `compareOffer` or similar helper here?

apps/store/src/components/ProductPage/ProductPage.tsx
```diff
@@ -22,3 +24,16 @@ export const ProductPage = ({ story, ...props }: ProductPageProps) => {
...
+  const router = useRouter()
+  const { shopSession } = useShopSession()
+
+  const cartEntryId = router.query.replace as string | undefined
+  const productOffer = useMemo(
+    () => shopSession?.cart.entries.find((item) => item.id === cartEntryId),
+    [shopSession?.cart.entries, cartEntryId],
+  )
+
+  return productOffer ?? null
```
alebedev, line 38: Let's just return memo result, why convent undefined to nul?

alebedev (approved):
Good one!  Some comments inside

apps/store/src/components/ProductPage/PriceIntentContext.tsx
```diff
@@ -40,19 +42,24 @@ const usePriceIntentContextValue = () => {
...
+          if (matchingOffer) return matchingOffer
+        }
 
-        if (matchingOffer) return matchingOffer
+        if (entryToReplace) {
+          const matchingReplaceOffer = data.priceIntent.offers.find((item) =>
+            compareOffer(item, entryToReplace),
+          )
+          if (matchingReplaceOffer) return matchingReplaceOffer
+        }
```
guilhermespopolin, line 62: I guess I didn't get this part: Why the selected offer should be the one corresponding with a cart entry that should be replaced, when we have one? 🤔

guilhermespopolin (approved):

apps/store/src/components/ProductPage/PurchaseForm/useHandleSubmitAddToCart.ts
```diff
@@ -35,7 +40,11 @@ export const useHandleSubmitAddToCart = ({ cartId, onSuccess }: Params) => {
     const productOfferId = getOrThrowFormValue(formData, FormElement.ProductOfferId)
     const nextUrl = event.nativeEvent.submitter?.getAttribute('value') ?? undefined
 
-    addEntry({
+    if (entryToReplace) {
```
robinandeer, line 43: Do we need to compare this with the entry we are adding?

PR #2157, opened 2023-04-14
Set transactionTotal to 0

gustaveen:
## Describe your changes
Our Adtraction program isn't based on order amount, total should be set to 0

Request by Adtraction


alebedev (approved):
Nice that it does not look random with the comment!

PR #2151, opened 2023-04-13
Support attributed_to query param, use it to create new shopSession with given attribution

alebedev:
## Describe your changes

Handle attributed_to query param
- initalize new shop session with proper attribution
- remove it from query to make page reload safe

We're not reading back the value from shopSession since it's not part of the schema

Will have undefined behavior if used on pages that rely on server-side created session - cart, checkout, etc.  We don't plan to use any of those in cross-sell link from our app

## Justify why they are needed

Track results of cross-selling through the app

## Checklist before requesting a review

- [ ] I have performed a self-review of my code

PR #2159, opened 2023-04-14
Add /api/redriect/recommend route to force session creation with given attributed_to value

alebedev:
## Describe your changes

Alternative approach to setting attributed_to - make a special API route to force session creation

More stable solution than relying on query param when starting the app (https://github.com/HedvigInsurance/racoon/pull/2151?no-redirect=1) 

### DIFF
--- a/apps/store/src/pages/api/redirect/recommend.ts
+++ b/apps/store/src/pages/api/redirect/recommend.ts
@@ -1,0 +1,40 @@
+import { NextApiRequest, NextApiResponse } from 'next'
+import { initializeApolloServerSide } from '@/services/apollo/client'
+import { setupShopSessionServiceServerSide } from '@/services/shopSession/ShopSession.helpers'
+import { getCountryByLocale } from '@/utils/l10n/countryUtils'
+import { getUrlLocale } from '@/utils/l10n/localeUtils'
+import { ORIGIN_URL, PageLink } from '@/utils/PageLink'
+
+const recommendHandler = async (req: NextApiRequest, res: NextApiResponse) => {
+  try {
+    const { query = {} } = req
+    const { attributed_to: attributedTo, next = '' } = query
+
+    if (typeof attributedTo !== 'string') {
+      return res.status(400).send('Expected attributed_to query param with single value')
+    }
+    if (typeof next !== 'string') {
+      return res.status(400).send('Multiple next values not supported')
+    }
+
+    const routingLocale = getUrlLocale(next)
+    const { countryCode } = getCountryByLocale(routingLocale)
+
+    const apolloClient = await initializeApolloServerSide({ req, res })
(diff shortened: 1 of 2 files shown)


alebedev:
* main
  * **PR #2159** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/2159" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/2159?utm_source=stack-comment).

apps/store/src/pages/api/redirect/recommend.ts
```diff
@@ -0,0 +1,35 @@
...
+import { setupShopSessionServiceServerSide } from '@/services/shopSession/ShopSession.helpers'
+import { getCountryByLocale } from '@/utils/l10n/countryUtils'
+import { getUrlLocale } from '@/utils/l10n/localeUtils'
+import { ORIGIN_URL, PageLink } from '@/utils/PageLink'
+
+const recommendHandler = async (req: NextApiRequest, res: NextApiResponse) => {
+  const { query = {} } = req
+  const { attributed_to: attributedTo, next = '' } = query
+
+  if (!attributedTo || Array.isArray(attributedTo)) {
```
robinandeer, line 12: I would just do:

```
if (typeof attributedTo !== 'string') {
```

apps/store/src/pages/api/redirect/recommend.ts
```diff
@@ -0,0 +1,35 @@
...
+    return res.status(400).send('Expected attributed_to query param with single value')
+  }
+  if (Array.isArray(next)) {
```
robinandeer, line 15: same here probably

apps/store/src/pages/api/redirect/recommend.ts
```diff
@@ -0,0 +1,35 @@
...
+    return res.status(400).send('Multiple next values not supported')
+  }
+
+  const routingLocale = getUrlLocale(next)
+  const { countryCode } = getCountryByLocale(routingLocale)
+
+  const apolloClient = await initializeApolloServerSide({ req, res })
+  const shopSessionService = setupShopSessionServiceServerSide({ apolloClient, req, res })
+  const shopSession = await shopSessionService.create({ countryCode, attributedTo })
+  shopSessionService.saveId(shopSession.id)
```
robinandeer, line 25: We might want to try/catch here so we make sure to redirect the user somewhere meaningful even if they encounter an error

robinandeer (approved):
Neat!

apps/store/src/pages/api/redirect/recommend.ts
alebedev: Good idea.  There's also empty string, but let's ignore it

PR #2154, opened 2023-04-13
Add toggle for campaign section in cart

robinandeer:
## Describe your changes

Add animation to "Switch" component

Make Switch component global

Add toggle for campaign section in cart

## Justify why they are needed

Design updates for the cart from Wille

## Checklist before requesting a review

- [ ] I have performed a self-review of my code

PR #2164, opened 2023-04-14
Add reveal animation for campaign section

robinandeer:
## Describe your changes

Add animation when expanding and collapsing the campaign section.

Checkout the preview app.

Testing it out with designers.

### DIFF
--- a/apps/store/src/components/CartInventory/CampaignsSection.tsx
+++ b/apps/store/src/components/CartInventory/CampaignsSection.tsx
@@ -1,8 +1,9 @@
+import { keyframes } from '@emotion/react'
 import styled from '@emotion/styled'
 import * as Collapsible from '@radix-ui/react-collapsible'
 import { useTranslation } from 'next-i18next'
 import { FormEventHandler, useState } from 'react'
-import { Button, CrossIconSmall, Heading, Space, Text, theme } from 'ui'
+import { Button, CrossIconSmall, Heading, Text, theme } from 'ui'
 import { Switch } from '@/components/Switch'
 import { TextField } from '@/components/TextField/TextField'
 import { CartCampaign } from './CartInventory.types'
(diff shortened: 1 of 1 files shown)


robinandeer:
* main
  * **PR #2154** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/2154" target="_blank">Graphite</a> 
    * **PR #2164** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/2164" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/2164?utm_source=stack-comment).

apps/store/src/components/CartInventory/CampaignsSection.tsx
```diff
@@ -144,3 +143,44 @@ const ChipButton = styled.button({
...
+    height: 'var(--radix-collapsible-content-height)',
+    opacity: 0,
+  },
+  '100%': {
+    height: 0,
+    opacity: 0,
+  },
+})
+
+const CollapsibleContent = styled(Collapsible.Content)({
```
alebedev, line 177: Is it easier than using framer-motion for animation?

alebedev (approved):

guilhermespopolin (approved):
https://user-images.githubusercontent.com/19200662/232456685-dd3d4e93-3760-435b-aa23-12a53c1b4e8c.mov

<br/>
The animation doesn't run so smoothly because of the spacing added/removed between the trigger and the form.

image

That spacing is added/removed outside the animation so that's why we can see that something "odd" is happening there. It's easier to spot it when content get's collapsed: The following happens: 

Spacing gets added/removed immediately --> content height/opacity gets animated into their final result;

Two ways I can see for solving it is making that spacing contribute to _Content_ height (by using a [Spacer component](https://www.joshwcomeau.com/react/modern-spacer-gif/)) or the spacing should always be there, regardless _Collapsable_ state.

apps/store/src/components/CartInventory/CampaignsSection.tsx
robinandeer: I think so, it's closer to pure CSS so nice to stick to if possible

It's also the official way from Radix 🙂

PR #2185, opened 2023-04-19
Make Insurely optional in Current Insurance picker

robinandeer:
## Describe your changes

Make it possible to edit current insurer without an Insurely client ID.

We currently just skip that section if there is no Insurely client ID. However, it should still be possible to get help with cancellation even if we don't perform price matching.

Now we always let the user pick their insurer, and if there is an Insurely client ID we will perform price matching.


robinandeer:
* main
  * **PR #2185** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/2185" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/2185?utm_source=stack-comment).

alebedev (approved):
Looks good

gustaveen (approved):
Good catch! 🕺

PR #2191, opened 2023-04-19
Add edit button to `CartEntryItem`

robinandeer:
## Describe your changes

- Allow users to edit cart items by pressing a button in the cart

## Justify why they are needed

- Makes it easier to edit items in the cart

## Checklist before requesting a review

- [ ] I have performed a self-review of my code

[GRW-2451]: https://hedvig.atlassian.net/browse/GRW-2451?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2190, opened 2023-04-19
Accept any Error in App Error dialog

robinandeer:
## Describe your changes

- Allow any Error to be passed to the AppError dialog

## Justify why they are needed

- Makes it more useful for cases outside of failed GraphQL requests

## Checklist before requesting a review

- [ ] I have performed a self-review of my code

PR #2192, opened 2023-04-20
Avoid flicker when navigating to PDP with open price calculator

robinandeer:
## Describe your changes

- Preload a given price intent from the URL when loading a PDP
- Derive the initial state of the price calculator (when possible)

- Avoid flickering between different price calculator states (during client side navigation)

- When we load the page from scratch (e.g. when we refresh the page) we can't really avoid the initial "loading" state while we fetch the price intent.


robinandeer:
* main
  * **PR #2190** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/2190" target="_blank">Graphite</a> 
    * **PR #2191** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/2191" target="_blank">Graphite</a> 
      * **PR #2192** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/2192" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/2192?utm_source=stack-comment).

apps/store/src/components/ProductPage/PurchaseForm/PurchaseForm.tsx
```diff
@@ -45,23 +47,30 @@ export const PurchaseForm = () => {
...
-  }
+  }, [isLarge, setFormState])
+
+  const handleQueryParamDetected = useCallback(() => {
+    tracking.reportOpenPriceCalculator({
+      id: productData.id,
+      displayNameFull: productData.displayNameFull,
+    })
+
+    editForm()
```
robinandeer, line 68: This also fixes a problem where we don't stop videos playing in the background

apps/store/src/components/ProductPage/PurchaseForm/useOpenPriceCalculatorQueryParam.ts
```diff
@@ -10,14 +10,22 @@ type Params = {
...
       // GOTCHA: We've tried removing query param via shallow router.replace, but it triggers some weird
       // window.location change, and this problem only ever happens in production
       triggeredRef.current = true
       onQueryParamDetected()
     }
-  }, [onQueryParamDetected, router])
+  }, [onQueryParamDetected, router, isPriceCalculatorExpanded])
+}
+
+export const useIsPriceCalculatorExpanded = () => {
```
robinandeer, line 27: This new hook let's us read the query param from the URL on first render if we are client navigating

apps/store/src/components/ProductPage/PriceIntentContext.tsx
```diff
@@ -79,6 +81,8 @@ const usePriceIntentContextValue = () => {
   useEffect(
     () =>
       onReady(async (shopSession) => {
+        if (priceIntentId) return
```
alebedev, line 84: Can we check for `preloadedPriceIntentId` to be more explicit about when it can happen?

apps/store/src/components/ProductPage/PurchaseForm/usePreloadedPriceIntentId.ts
```diff
@@ -0,0 +1,9 @@
+import { useRouter } from 'next/router'
+
+export const PRELOADED_PRICE_INTENT_QUERY_PARAM = 'pi_id'
```
alebedev, line 3: Style question - do we want query params to be readable (I think this is current style) or short?

alebedev (approved):
Great change!  Some questions inside

apps/store/src/components/ProductPage/PurchaseForm/PurchaseForm.tsx
guilhermespopolin: [Question]
Sounds redundant calling `editorForm` here since now form will be proper initialised with _Edit_ state as `isPriceCalculatorExpanded` can be read within the first render. Additionally, in this case I think it will case a re-render of `PurchaseForm` because `usePurchaseFormState.setFormState('EDIT')` will return a new object. All of those things don't actually matter that much but what is the issue with background videos is this solving again?

apps/store/src/components/ProductPage/PurchaseForm/usePreloadedPriceIntentId.ts
```diff
@@ -0,0 +1,9 @@
...
+
+export const usePreloadedPriceIntentId = () => {
+  const router = useRouter()
+  const { [PRELOADED_PRICE_INTENT_QUERY_PARAM]: priceIntentId } = router.query
+  return typeof priceIntentId === 'string' ? priceIntentId : undefined
+}
```
guilhermespopolin, line 9: Don't we need to check for `router.isReady` here?

guilhermespopolin (approved):
Well done. Just left some [Question] comments.

apps/store/src/components/ProductPage/PurchaseForm/PurchaseForm.tsx
robinandeer: So this is confusing but I will try to explain:

1. If we "client navigate" we have access to URL on first render.

2. If we "server navigate" to this page we will not have access to URL on first render. Therefore, the initial form state will be "IDLE". So we need to update when "router is ready".

3. Right now we just update the form state but we also want to send the related dialog event - which happens in `editForm`.

apps/store/src/components/ProductPage/PriceIntentContext.tsx
robinandeer: Hmm yes I did that at first, however, I think this will cover even more cases

What we actually case about is; do we already have a price intent ID or not.

apps/store/src/components/ProductPage/PurchaseForm/usePreloadedPriceIntentId.ts
robinandeer: right right, forgot to ask that -- we should unify it

I can update to make it more readable 🙂

apps/store/src/components/ProductPage/PurchaseForm/usePreloadedPriceIntentId.ts
robinandeer: I don't think it makes any difference actually

We will just end up returning either price intent ID or undefined anyways, right?

apps/store/src/components/ProductPage/PurchaseForm/PurchaseForm.tsx
guilhermespopolin: Ok. I guess I don't understand '2' happens 😅 
Maybe I'm lacking some fundamental piece of knowledge

robinandeer:
@robinandeer queued this pull request to merge with [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/2192).

PR #2215, opened 2023-04-25
Remove extra padding in `CartEntryItem`

robinandeer:
## Describe your changes

- Remove extra padding from `CartEntryItem` in read-only mode

- Refactor layout a bit to reduce some complexity

- The extra padding was causing the `CartEntryItem` to be too tall in read-only mode


gustaveen (approved):
Great! 👌 Felt a bit overkill to use grid areas for this 🙌

PR #2222, opened 2023-04-25
Update Cart mutations to use new API

robinandeer:
## Describe your changes

Update to use new API for managing the cart

Get rid of "cartId"

### DIFF
--- a/apps/store/src/components/ProductPage/PurchaseForm/useHandleSubmitAddToCart.ts
+++ b/apps/store/src/components/ProductPage/PurchaseForm/useHandleSubmitAddToCart.ts
@@ -13,12 +13,12 @@ import { useCartEntryToReplace } from '../ProductPage'
 import { FormElement } from './PurchaseForm.constants'
 
 type Params = {
-  cartId: string
+  shopSessionId: string
   priceIntentId: string
   onSuccess: (productOfferId: string, nextUrl?: string) => void
 }
 
-export const useHandleSubmitAddToCart = ({ cartId, onSuccess }: Params) => {
+export const useHandleSubmitAddToCart = ({ shopSessionId, onSuccess }: Params) => {
   // ProductRecommendationsQuery needs to be an active query
   // before we can "re"fetch it after adding a new product into
   // the cart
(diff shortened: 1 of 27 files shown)


apps/store/src/components/ProductPage/PurchaseForm/useHandleSubmitAddToCart.ts
```diff
@@ -77,12 +78,17 @@ const useCartEntryAdd = (mutationOptions: CartEntryAddOptions = {}) => {
...
     (options: CartEntryAddOptions = {}) => {
       const handleCompleted = (data: CartEntryAddMutation) => {
-        if (!data.cartEntriesAdd.cart) return
+        const updatedShopSession = data.shopSessionCartEntriesAdd.shopSession
+        console.log('updatedShopSession', updatedShopSession)
+        if (!updatedShopSession) return
 
         const { variables } = options
-        const addedOffer = data.cartEntriesAdd.cart.entries.find(
+        // eslint-disable-next-line @typescript-eslint/no-unsafe-call
```
robinandeer, line 86: couldn't figure out why this is causing a listing error

alebedev: Can you double-check in CLI?  Should not fail

apps/store/src/graphql/CartUnredeemCampaign.graphql
```diff
@@ -1,5 +1,12 @@
-mutation UnredeemCampaign($cartId: UUID!, $campaignId: String!) {
-  cartUnredeemCampaign(input: { cartId: $cartId, campaignId: $campaignId }) {
-    ...CartFragment
+mutation UnredeemCampaign($shopSessionId: UUID!, $campaignId: String!) {
+  shopSessionCartCampaignUnredeem(
+    input: { shopSessionId: $shopSessionId, campaignId: $campaignId }
+  ) {
+    shopSession {
```
alebedev, line 5: Nit: maybe we want to declare ShopSessionMutationOutput and reuse it?

alebedev, line 5: Nit: We only have 1 use of CartFragment remaining, what do you think about inlining it?

alebedev (approved):
Looks good!

What do you think about using context for shopSessionId instead of prop drilling?

PR #2270, opened 2023-05-03
Move switching assistant to confirmation page

robinandeer:
## Describe your changes

- Move switching assistant module to the confirmation page.

- Skip "switching assistant" step during checkout flow.

- Updates from design.

- I didn't remove all of the code related to switching assistant page in this PR to make it easier to revert if needed.

- I haven't been able to test it out end-2-end since I don't have a car 😛


apps/store/src/components/ConfirmationPage/ConfirmationPage.tsx
```diff
@@ -41,7 +42,14 @@ export const ConfirmationPage = (props: Props) => {
                 <CartInventory shopSessionId={props.shopSessionId} cart={cart} readOnly />
               </Space>
 
-              <Space y={1}>
+              {props.switching && (
```
alebedev, line 45: Can you comment in the switching page that it's currently unused?

alebedev (approved):
Nice!

As a follow-up, can you think of a way to track interaction with switching component via Datadog?

PR #2317, opened 2023-05-09
Set correct html lang

gustaveen:

### DIFF
--- a/apps/store/src/utils/l10n/locales.ts
+++ b/apps/store/src/utils/l10n/locales.ts
@@ -8,6 +8,7 @@ export type LocaleData = {
   locale: IsoLocale
   language: Language
   routingLocale: RoutingLocale
+  htmlLang: 'sv' | 'en-SE' | 'no' | 'en-NO' | 'da' | 'en-DK'
   chatWidgetSrc?: string
 }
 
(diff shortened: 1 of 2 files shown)


gustaveen:
* main
  * **PR #2317** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/2317" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/2317?utm_source=stack-comment).

apps/store/src/utils/l10n/locales.ts
```diff
@@ -8,6 +8,7 @@ export type LocaleData = {
   locale: IsoLocale
   language: Language
   routingLocale: RoutingLocale
+  htmlLang: 'sv-SE' | 'en-SE' | 'no-NO' | 'en-NO' | 'da-DK' | 'en-DK'
```
alebedev, line 11: Let's use `locale` above, it's the same value

alebedev (changes requested):

apps/store/src/utils/l10n/locales.ts
gustaveen: I kind of like being explicit on those values and in this case it differs on norwegian htmlLang `no-NO`

alebedev (approved):
Makes sense, let's keep separate field with a comment why we're doing it

PR #2366, opened 2023-05-15
Show discount banner if shopSession has discount

gustaveen:


gustaveen:
* main
  * **PR #2366** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/2366" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/2366?utm_source=stack-comment).

apps/store/src/pages/[[...slug]].tsx
```diff
@@ -51,6 +54,14 @@ const NextProductPage = (props: ProductPageProps) => {
   const { story: initialStory, ...pageProps } = props
   const story = useStoryblokState(initialStory)
 
+  const { shopSession } = useShopSession()
+  const setGlobalBanner = useSetGlobalBanner()
+  const { t } = useTranslation()
+  const hasDiscount = shopSession?.cart.redeemedCampaigns.length !== 0
+  if (hasDiscount) {
```
Comment (alebedev) [inline apps/store/src/pages/[[...slug]].tsx:61]: Would it work? 
- Page can be server-side rendered and we don't want to handle shopSession there
- setState should not be used inline in component body.  Maybe try `onReady` callback from `useShopSession`

alebedev (commented):
Some questions inside.  I'm OK with any solution that would work

apps/store/src/pages/[[...slug]].tsx
Comment (gustaveen) [reply apps/store/src/pages/[[...slug]].tsx:61]: Yeah this solution was not optimal, would ideally like to set the global banner in [code].tsx when we redeem the code but since it's stored in an atom I'm not sure how we should accomplish that. I'll scratch this and see if I can find another solution.

Comment (gustaveen) [reply apps/store/src/pages/[[...slug]].tsx]: Yes, that didn't work properly @guilhermespopolin Updated the solution now 👍

guilhermespopolin (approved):
Nice 👌🏻

PR #2397, opened 2023-05-23
Use exact path for Edge config redirects

gustaveen:
## Describe your changes
Use exact path for Edge config redirects

For these types of redirects it makes more sense to use exact paths

An example:
We want `/se-en/forever` to redirect to `se-en/hedvig/discount` but not catch `/se-en/forever/[code]`

Make it easier to add custom redirects in Edge Config. If we want to use wildcard redirects it makes more sense to put it in next config


robinandeer (approved):
Yeah I think it makes sense to start being very strict until we realize we need something more complex 🙂

PR #2398, opened 2023-05-23
Load C1 script globally

gustaveen:
## Describe your changes
Load C1 script globally

Make sure we initialise the Customer 1 script not only on pages where we have the SupportBlock


robinandeer (approved):
Great!

PR #2410, opened 2023-05-24
Render chat widget on default page template and show in mobile

gustaveen:


guilhermespopolin (approved):
Nice catch

robinandeer (approved):
ahhh good catch

PR #2418, opened 2023-05-25
feat(CartEntry): use same date format in CartEntry - summary and detaisl sheet

guilhermespopolin:
## Describe your changes

Use same date format in CartEntry - summary and details sheet

Requested by design.

### Before

before

### After

after


robinandeer (approved):
Ahh very good, I missed that one

PR #2425, opened 2023-05-26
[GEN-226] Support absolute URLs for the 'next' query param

robinandeer:
## Describe your changes

Support passing absolute URLs in the "next" query parameter.

This will be used by AdTraction for affiliate links.

We will still not allow any next redirects to go to a different domain than the current one.

### DIFF
--- a/apps/store/src/utils/getPathnameFromUrl.ts
+++ b/apps/store/src/utils/getPathnameFromUrl.ts
@@ -1,0 +1,7 @@
+export const getPathnameFromUrl = (rawUrl: string) => {
+  if (rawUrl.startsWith('https://')) {
+    return new URL(rawUrl).pathname
+  }
+
+  return rawUrl
+}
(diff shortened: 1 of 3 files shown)


robinandeer:
* main
  * **PR #2425** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/2425" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/2425?utm_source=stack-comment).

apps/store/src/utils/getPathnameFromUrl.ts
```diff
@@ -0,0 +1,8 @@
+export const getPathnameFromUrl = (rawUrl: string) => {
+  try {
+    return new URL(rawUrl).pathname
```
alebedev, line 3: Nit: you can avoid error when `rawUrl` is path-only:
```
new URL(rawUrl, ORIGIN_URL)
```

alebedev (approved):
Nice, one comment inside

robinandeer (commented):
Hmm you mean we should check first if the URL is relative or absolute?

apps/store/src/utils/getPathnameFromUrl.ts
robinandeer: Hmm you mean we should check first if the URL is relative or absolute?

PR #2446, opened 2023-05-29
fix(QuickPurchaseForm): hidden input for product name where only one product is available

guilhermespopolin:
## Describe your changes

* Renders a hidden input for `productName` when only one product is available;

By design, we shouldn't render product selector when only one option for product is available. However we still to render an input for it to be able to get that value while submitting the form.

### DIFF
--- a/apps/store/src/blocks/QuickPurchaseBlock.tsx
+++ b/apps/store/src/blocks/QuickPurchaseBlock.tsx
@@ -76,7 +76,7 @@ export const QuickPurchaseBlock = ({ blok, nested }: QuickPurchaseBlockProps) =>
       const formState = Object.fromEntries(formData.entries())
       const { [SSN_FIELDNAME]: ssn, [PRODUCT_FIELDNAME]: productName } = formState
 
-      // This should never happen since QuickPurchaseForm conrols are required
+      // This should never happen since QuickPurchaseForm controls are required
       if (typeof ssn !== 'string' || typeof productName !== 'string') {
         throw new Error('[QuickPurchaseBlock]: ssn and product are required')
       }
--- a/apps/store/src/components/QuickPurchaseForm/ProductSelector.tsx
+++ b/apps/store/src/components/QuickPurchaseForm/ProductSelector.tsx
@@ -21,6 +21,10 @@ type Props = {
 export const ProductSelector = ({ productOptions, ...delegated }: Props) => {
   const { t } = useTranslation('purchase-form')
 
+  if (productOptions.length === 1) {
+    return <input type="hidden" name={delegated.name} value={productOptions[0].value} />
+  }
+
   return (
     <SelectPrimitive.Root {...delegated}>
       <SelectTrigger>
--- a/apps/store/src/components/QuickPurchaseForm/QuickPurchaseForm.tsx
+++ b/apps/store/src/components/QuickPurchaseForm/QuickPurchaseForm.tsx
@@ -42,14 +42,12 @@ export const QuickPurchaseForm = ({
   return (
     <form onSubmit={onSubmit}>
       <Space y={0.25}>
-        {productOptions.length > 1 && (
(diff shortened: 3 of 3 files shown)


guilhermespopolin:
* main
  * **PR #2446** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/2446" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/2446?utm_source=stack-comment).

apps/store/src/components/QuickPurchaseForm/ProductSelector.tsx
```diff
@@ -21,6 +21,14 @@ type Props = {
 export const ProductSelector = ({ productOptions, ...delegated }: Props) => {
   const { t } = useTranslation('purchase-form')
 
+  if (productOptions.length === 0) {
+    return null
+  }
+
+  if (productOptions.length === 1) {
```
alebedev, line 28: Nit: `esle if` would make reading the code easier

alebedev (approved):
Looks good

apps/store/src/blocks/QuickPurchaseBlock.tsx
```diff
@@ -76,7 +76,7 @@ export const QuickPurchaseBlock = ({ blok, nested }: QuickPurchaseBlockProps) =>
       const formState = Object.fromEntries(formData.entries())
       const { [SSN_FIELDNAME]: ssn, [PRODUCT_FIELDNAME]: productName } = formState
 
-      // This should never happen since QuickPurchaseForm conrols are required
+      // This should never happen since QuickPurchaseForm controls are required
```
guilhermespopolin, line 79: This issue kind of broke this comment 😅

apps/store/src/components/QuickPurchaseForm/ProductSelector.tsx
```diff
@@ -21,6 +21,12 @@ type Props = {
 export const ProductSelector = ({ productOptions, ...delegated }: Props) => {
   const { t } = useTranslation('purchase-form')
 
+  if (productOptions.length === 0) {
+    return null
```
robinandeer, line 25: What are we supporting here? Should we through an error instead?

apps/store/src/blocks/QuickPurchaseBlock.tsx
robinandeer: We are missing an error boundary if we are throwing errors in the render function ;)

robinandeer (approved):
Looks good, just had a question about case where we don't have any product options...

apps/store/src/components/QuickPurchaseForm/ProductSelector.tsx
guilhermespopolin: 🤔 I see your point. But maybe throwing an error in `ProductSelector` would not be the best because the `productOptions` list is built on the client side so at initial render that list will be empty. I'm not sure what would the best option here. Still throw the error while submitting the form, or maybe inside `QuickPurchaseBlok` do something like the following:
```
if (availableProducts.length > 0 && blok.products.length > 0 && productOptions.length === 0) {
   // Throw an error informing no matches for the provided products were found
}
```

robinandeer: Hmm yes it's a good point

1. We could treat dev differently from production - like prevent build maybe

2. We could show some error banner to the editor

3. Add some default option so we never end up in this case

4. Or maybe just adding a comment is enough

guilhermespopolin: Thanks for the response. Option 2 and 1 sounds promising. If that's ok I'd like to merge this to unblock Peter and improve it in a different PR.

PR #2448, opened 2023-05-29
fix(QuickPurchaseForm): avoid flashing UI

guilhermespopolin:
## Describe your changes

Don't render `ProductSelector` if no `productOptions` list is empty.

To avoid "flashing UI"

https://github.com/HedvigInsurance/racoon/assets/19200662/5c6183fc-4dda-4d92-8b5e-8d20d2d9c69d

**Disclaimer**

It will happen when more than one `productOption` is available. But adding UI pieces after something gets loaded seems less disruptive then removing it. 

https://github.com/HedvigInsurance/racoon/assets/19200662/1e23a595-ad49-4573-80b0-87eb024184a7

## Relates with

https://github.com/HedvigInsurance/racoon/pull/2446

### DIFF
--- a/apps/store/src/components/QuickPurchaseForm/ProductSelector.tsx
+++ b/apps/store/src/components/QuickPurchaseForm/ProductSelector.tsx
@@ -21,10 +21,6 @@ type Props = {
 export const ProductSelector = ({ productOptions, ...delegated }: Props) => {
   const { t } = useTranslation('purchase-form')
 
-  if (productOptions.length === 1) {
-    return <input type="hidden" name={delegated.name} value={productOptions[0].value} />
-  }
-
   return (
     <SelectPrimitive.Root {...delegated}>
       <SelectTrigger>
--- a/apps/store/src/blocks/QuickPurchaseBlock.tsx
+++ b/apps/store/src/blocks/QuickPurchaseBlock.tsx
@@ -4,6 +4,7 @@ import { useRouter } from 'next/router'
 import { useState, useMemo, type FormEventHandler } from 'react'
 import { useTranslation } from 'react-i18next'
 import { theme } from 'ui'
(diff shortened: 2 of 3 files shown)


robinandeer (commented):
Hmm is this an issue in production? Just trying to understand.

Seems like we would have all the information to decide whether to render or not from the start. Or are we fetching some async data after the first render?

guilhermespopolin:
> Hmm is this an issue in production? Just trying to understand.
> 
> Seems like we would have all the information to decide whether to render or not from the start. Or are we fetching some async data after the first render?

Yes. It was added by [this](https://github.com/HedvigInsurance/racoon/pull/2446). It wasn't happening before because we were not rendering the product selector if the number of product options were <= 1. But that was causing another issue where the selected product was not available while submitting the form. The mentioned PR solved the form bug by rendering a hidden input with the selected product when only one options is available. However, that list is built on the client side so on initial render it will be empty. So I thought about preventing it of being rendered while the list is being built and on a following PR, I intent to add a banner only for editors in case something goes south while building that list.

apps/store/src/components/QuickPurchaseForm/ProductSelector.tsx
```diff
@@ -21,7 +21,9 @@ type Props = {
 export const ProductSelector = ({ productOptions, ...delegated }: Props) => {
   const { t } = useTranslation('purchase-form')
 
-  if (productOptions.length === 1) {
+  if (productOptions.length === 0) {
+    return null
+  } else if (productOptions.length === 1) {
     return <input type="hidden" name={delegated.name} value={productOptions[0].value} />
   }
```
robinandeer, line 28: why don't we move all this outside this component and handle it in the parent component?

robinandeer (approved):
Looks great!

[approveyoda](https://app.graphite.dev/macros/HedvigInsurance)

PR #2451, opened 2023-05-29
fix(DetailsSheet): sanitize rows

guilhermespopolin:
## Describe your changes

'Sanitize' details sheet offer related rows by not showing rows that doesn't contain values associated with it. 

For hedvig-com created pet offers we want to display 'Boyta' (living space) row but for many pets offers we don't have that info and therefore want to hide from the table.

**Before**

Screenshot 2023-05-29 at 14 03 22

**After**

Screenshot 2023-05-29 at 14 03 44

### DIFF
--- a/apps/store/src/components/CartInventory/CartEntryItem/DetailsSheet.tsx
+++ b/apps/store/src/components/CartInventory/CartEntryItem/DetailsSheet.tsx
@@ -13,15 +13,20 @@ export const DetailsSheet = (props: CartEntry) => {
   const dataTableRows = getDataTable(productName)
   const getDataTableValue = useGetDataTableValue()
 
+  const allRows =
+    dataTableRows?.map((item) => ({
+      title: t(item.label, { defaultValue: `${item.label} MISSING` }),
+      value: getDataTableValue(item, data),
+    })) ?? []
+  const rowsWithValues = allRows.filter((item) => item.value !== null)
+
   return (
     <Root y={1}>
       <Table>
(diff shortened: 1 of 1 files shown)


guilhermespopolin:
* main
  * **PR #2451** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/2451" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/2451?utm_source=stack-comment).

apps/store/src/components/CartInventory/CartEntryItem/DetailsSheet.tsx
```diff
@@ -1,6 +1,7 @@
 import styled from '@emotion/styled'
 import type { TFunction } from 'i18next'
 import { useTranslation } from 'next-i18next'
+import { ReactNode } from 'react'
```
robinandeer, line 4: nit: `import { type ReactNode } ...`

robinandeer (approved):
Another more "functional" approach would be map/reduce

Pseudo code

```
const allRows = dataTableRows.map((item) => ({
  title: t(item.label....),
  value: getDataTableValue(item, data),
}))

const rowsWithValues = allRows.filter((item) => item.value !== null)

return (
  <Table>{rowsWithValues.map((item) => (...)}</Table>
)
```

Might read a bit easier since we avoid mixing JSX and data transformation 🙂

PR #2455, opened 2023-05-30
[GEN-390] Add script to upload sourcemaps from Vercel

robinandeer:
## Describe your changes

Add script to upload sourcemaps to Datadog from Vercel.

Vercel has a limit of amount of characters in the custom build script so we can't inline the full command.


alebedev (approved):
Looks good

PR #2458, opened 2023-05-30
GEN-392 - feat(GlobalBanner): make usage of SessionStorage to keep banner opened/closed state

guilhermespopolin:
## Describe your changes

* Make usage of SessionStorage to keep banner opened/closed state

When it comes about `atomWithStorage`:

> If not specified, the default storage implementation uses localStorage for storage/retrieval, JSON.stringify()/JSON.parse() for serialization/deserialization, and subscribes to storage events for cross-tab synchronization.

Then Peter mentioned this:

> How is the cache for closed/removed banner configured? I closed a discount banner in my browser two weeks ago. Now it never shows up again, only in incognito. Could we cache it only during the user session.

So I guess it would be better to keep that temporary state into `SessionStorage` despite the tradeoff of not having it shared between multiple tabs.


robinandeer (approved):
aha nice change -- we don't need the tab syncing 😅

PR #2459, opened 2023-05-30
[GEN-339] Add support for basic experiments (A/B testing)

robinandeer:
## Describe your changes

Add setup to support basic experiments / A/B testing.

See README for more information.

We should remove the "new site" experiment in a follow-up PR.

This will allow us to run experiments on the same page/slug.

### DIFF
--- a/apps/store/src/services/Tracking/Tracking.ts
+++ b/apps/store/src/services/Tracking/Tracking.ts
@@ -10,7 +10,6 @@ import {
 } from '@/services/gtm'
 import { PriceIntent } from '@/services/priceIntent/priceIntent.types'
 import { ShopSession } from '@/services/shopSession/ShopSession.types'
-import { newSiteAbTest } from '../../newSiteAbTest'
 import { getAdtractionProductCategories } from './adtraction'
 
 type TrackingContext = Partial<Record<TrackingContextKey, unknown>>
@@ -104,11 +103,11 @@ export class Tracking {
     pushToGTMDataLayer(event)
   }
 
-  public reportExperimentImpression(variantId: string) {
+  public reportExperimentImpression(experimentId: string, variantId: string) {
     const event = {
       event: TrackingEvent.ExperimentImpression,
       eventData: {
-        experiment_id: newSiteAbTest.optimizeExperimentId,
+        experiment_id: experimentId,
         variant_id: variantId,
       },
     }
--- a/apps/store/src/services/Tracking/experiment.constants.ts
+++ b/apps/store/src/services/Tracking/experiment.constants.ts
@@ -1,0 +1,38 @@
+export const EXPERIMENT_COOKIE_NAME = 'hedvig-experiment'
+
+export type Experiment = {
+  id: string
+  name: string
+  variants: ExperimentVariant[]
+  slug: string
+}
+
+export type ExperimentVariant = {
+  id: number
+  name: string
+  weight: number
+  slug?: string
+}
+
+const EXPERIMENT_ID = process.env.NEXT_PUBLIC_EXPERIMENT_ID
+
+export const CURRENT_EXPERIMENT: Experiment | undefined = EXPERIMENT_ID
+  ? {
+      name: 'Start page',
+      id: EXPERIMENT_ID,
+      slug: '/se',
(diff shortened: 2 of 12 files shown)


robinandeer:
* main
  * **PR #2459** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/2459" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/2459?utm_source=stack-comment).

apps/store/src/services/Tracking/Tracking.ts
```diff
@@ -104,11 +103,11 @@ export class Tracking {
     pushToGTMDataLayer(event)
   }
 
-  public reportExperimentImpression(variantId: string) {
+  public reportExperimentImpression(experimentId: string, variantId: string) {
```
alebedev, line 106: Will break reporting new site experiment, let's remove it soon

apps/store/src/services/Tracking/experiment.constants.ts
```diff
@@ -0,0 +1,38 @@
+export const COOKIE_NAME = 'hedvig-experiment'
```
alebedev, line 1: Nit: `EXPERIMENT_COOKIE_NAME` so it's easy to understand without looking at import

apps/store/src/services/Tracking/experiment.constants.ts
```diff
@@ -0,0 +1,38 @@
...
+export type Experiment = {
+  id: string
+  name: string
+  variants: ExperimentVariant[]
+  slug: string
+}
+
+export type ExperimentVariant = {
+  id: number
+  name: string
```
alebedev, line 12: Do we need name here?  I thought it's usually managed in Optimize or other A/B tracking software

apps/store/src/services/Tracking/experimentMiddleware.ts
```diff
@@ -0,0 +1,82 @@
...
+
+    url.pathname = assignedVariant.variant.slug
+  }
+
+  const response = NextResponse.rewrite(url)
+
+  if (!req.cookies.has(COOKIE_NAME)) {
+    console.debug(`AB test | setting cookie: ${assignedVariant.cookieValue}`)
+    response.cookies.set(COOKIE_NAME, assignedVariant.cookieValue, {
+      maxAge: 7 * 24 * 3600,
```
alebedev, line 77: Should we have a name constant here?

alebedev (approved):
Nice!  Some minor comments

apps/store/src/services/Tracking/Tracking.ts
robinandeer: that experiment is already closed down, yes?

but what will break it?

apps/store/src/services/Tracking/experiment.constants.ts
robinandeer: yeah I don't think we actually use it, was just included from the example

robinandeer: it's not really used but I include it in some log messages so they are easier to understand :)

apps/store/src/services/Tracking/experimentMiddleware.ts
```diff
@@ -0,0 +1,82 @@
...
+  const currentExperiment = getCurrentExperiment()
+  if (!currentExperiment) return
+
+  const slug = currentExperiment.slug.replace(req.nextUrl.locale, '')
+  if (req.nextUrl.pathname !== slug) return
+
+  let assignedVariant = getAssignedVariant(req)
+
+  if (!assignedVariant) {
+    assignedVariant = assignVariant()
```
guilhermespopolin, line 57: Can this logic of creating and assign a variant in case we don't have one already be placed inside `getAssignedVariant`? Then we could simplify here with:

```suggestion
  if (!assignedVariant) return
```

guilhermespopolin (approved):
Nice. Just left one comment :)

guilhermespopolin (approved):
Nice. Just left one comment :)

PR #2469, opened 2023-06-01
feat(Banner): added banner component

guilhermespopolin:
## Describe your changes

* Extract `GlobalBanner` UI into it's own component. Down in the stack I'm gonna reuse that component - `Banner` - within `GlobalBanner` and `AnnouncementBlock`

Changes:

* It supports 4 variants: _info_, _campaign_, _warning_ and _error_ 
* It's content is now perfectly centered in the screen as it's not taking the _close button_ into account while aligning as it was before.

image

## Justify why they are needed

We need to provide a way for editors to render an _announcement_ (I've named it this while lacking a better name - we already have a banner block) by page. Despite looking exactly the same as the `GlobalBanner` it should function differently. That's why I thought about extracting the UI into a separate component and reuse it between `GlobalBanner` and `AnnouncementBlock`.

PR #2470, opened 2023-06-01
refact(GlobalBanner): integrate new Banner UI componnet

guilhermespopolin:
## Describe your changes

* Updates `GlobalBanner` so it makes usage `Banner` component.
 
`GlobalBanner` is only responsible for displaying a `Banner` globally - UI is managed by `Banner` which down the stack will also be used by `AnnouncementBlock`

### DIFF
--- a/apps/store/src/components/GlobalBanner/GlobalBanner.tsx
+++ b/apps/store/src/components/GlobalBanner/GlobalBanner.tsx
@@ -1,20 +1,17 @@
-import { RESET } from 'jotai/utils'
 import { Banner } from '@/components/Banner/Banner'
 import { useDebugShopSession } from '@/utils/useDebugShopSession'
-import { useGlobalBanner, useGlobalBannerClosed } from './useGlobalBanner'
+import { useGlobalBanner } from './useGlobalBanner'
 
 export const GlobalBanner = () => {
-  const [banner, setBanner] = useGlobalBanner()
-  const [closed, setIsClosed] = useGlobalBannerClosed()
+  const { banner, dismissBanner } = useGlobalBanner()
 
   useDebugShopSession()
 
   const handleClose = () => {
-    setBanner(RESET)
-    setIsClosed(true)
+    dismissBanner()
   }
 
-  if (banner === null || closed) return null
+  if (banner === null) return null
 
   return (
     <Banner variant={banner.variant} handleClose={handleClose}>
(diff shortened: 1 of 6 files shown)


guilhermespopolin:
* main
  * **PR #2469** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/2469" target="_blank">Graphite</a> 
    * **PR #2470** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/2470" target="_blank">Graphite</a>  👈
      * **PR #2471** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/2471" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/2470?utm_source=stack-comment).

apps/store/src/components/GlobalBanner/GlobalBanner.tsx
```diff
@@ -1,65 +1,24 @@
...
 import { useGlobalBanner, useGlobalBannerClosed } from './useGlobalBanner'
 
 export const GlobalBanner = () => {
-  const [value, setValue] = useGlobalBanner()
+  const [banner, setBanner] = useGlobalBanner()
   const [closed, setIsClosed] = useGlobalBannerClosed()
 
   useDebugShopSession()
 
   const handleClose = () => {
```
alebedev, line 12: Is there a way to store banner and closed in the same state element?  It looks odd that they are unrelated and have to changed independently

robinandeer (approved):
Other than @alebedev comment -- I think it looks good

apps/store/src/components/GlobalBanner/GlobalBanner.tsx
guilhermespopolin: They need to be separated things because of the way `useDiscountBanner` works. That hook will check for check for redeemed campaigns within the shop session **and** check if the user, during his session, haven't dismissed the banner and on that scenario add a banner informing the about campaign code. If I keep them together it will be very hard to check if I should add a banner about campaign code or not.

alebedev: As an idea, look at https://jotai.org/docs/guides/composing-atoms and derived atoms.  I think we're dealing with something that can be described as a value derived from 2 different inputs.  Let's try to express it via library API

guilhermespopolin: I've updated the PR with derivated atoms. Please, give a look :)

alebedev (approved):

PR #2478, opened 2023-06-02
feat(MigrationPageBlock): support announcements

guilhermespopolin:
## Describe your changes

* Updates _ManyPetsMigrationPage_ content type so it also supports _Announcements_ (`AnnouncementBlock`)

### DIFF
--- a/apps/store/src/pages/manypets/migration/[shopSessionId].tsx
+++ b/apps/store/src/pages/manypets/migration/[shopSessionId].tsx
@@ -47,13 +47,16 @@ const NextManyPetsMigrationPage: NextPage<Props> = ({
   latestAdoptionDate,
   comparisonTableData,
 }) => {
-  const { preOfferContent, postOfferContent } = story.content
+  const { announcement, preOfferContent, postOfferContent } = story.content
 
   return (
     <>
       <HeadSeoInfo story={story} />
       <ManyPetsMigrationPage
         migrationSessionId={migrationSessionId}
+        announcement={announcement.map((blok) => (
+          <StoryblokComponent key={blok._uid} blok={blok} />
+        ))}
         preOfferContent={preOfferContent?.map((blok) => (
           <StoryblokComponent key={blok._uid} blok={blok} />
         ))}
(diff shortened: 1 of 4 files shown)


apps/store/src/pages/manypets/migration/[shopSessionId].tsx
```diff
@@ -47,13 +47,16 @@ const NextManyPetsMigrationPage: NextPage<Props> = ({
   latestAdoptionDate,
   comparisonTableData,
 }) => {
-  const { preOfferContent, postOfferContent } = story.content
+  const { announcement, preOfferContent, postOfferContent } = story.content
```
Comment (robinandeer) [inline apps/store/src/pages/manypets/migration/[shopSessionId].tsx:50]: Something feels odd here:

1. either call it announcements (plural) and keep the rest

2. slice out the first item and discard the rest of the array if we only want to support a single announcement

guilhermespopolin:
> Don't really understand the difference between "ManyPetsMigrationPageBlock" and "ManyPetsMigrationPage" 🤔 but seems like we are repeating a lot of stuff there
> 
> Otherwise looks good :)

The _Many Pets Migration Page_ has a dynamic segment which references the _migration shop session id_, so it's path is `/manypets/migration/[shopSessionId]`. Something like 90% of that page is static and could be built by editors using existing blocks. However there's a section of the page, the one that renders the comparison table and cart entries, that needs to fetch info based on the _migration shop session id_ present in the URL. To enable editors to build the static parts of that page I've created a new _Content Type_ - `ManyPetsMigrationPageBlock`. So when loading `manypets/migration` in storyblok, `src/pages/manypets/[[...slug]].tsx` file will be used, which will then render the referred block - `ManyPetsMigrationPageBlock`. Under the hood that block just renders `<ManyPetsMigrationPage>` component and since in storyblok there's no way to preview the dynamic segment of the page, we just pass to it empty values for stuff like `offers` and `totalCost`.

Now when loading something like `/manypets/migration/asb1b23bs`, `src/pages/manypets/migration/[shopSessionId].tsx` file will be used. That file will them load the dynamic info and render  the same component as the content type - `<ManyPetsMigrationPage>` but this time we can provide the data necessary for the dynamic section, like `offers` and `totalCost`.

Both the _Content Type_ and the _Migration Page_ use the same UI to keep consistency between the page editors change and the final result. The only difference is that in the final result will also include the dynamic part of it.

apps/store/src/pages/manypets/migration/[shopSessionId].tsx
Comment (guilhermespopolin) [reply apps/store/src/pages/manypets/migration/[shopSessionId].tsx:50]: The issue is that `ExpectedBlockType` is always an array.

```
export type ManyPetsMigrationStory = ISbStoryData<
  {
    announcement: ExpectedBlockType<ReusableBlockReferenceProps>
    preOfferContent?: Array<SbBlokData>
    postOfferContent: Array<SbBlokData>
  } & SEOData
>
```

PR #2492, opened 2023-06-05
Add breadcrumb list above footer

robinandeer:
## Describe your changes

- Add Breadcrumb List above the page footer (and option to hide from Storyblok)

- Include structured data based on the breadcrumb list

- Add helper function to generate breadcrumb list

- Add new text color to Text-component: "textSecondaryOnGray"

- Make link styles more re-usable

SEO and site improvements

### DIFF
--- a/apps/store/src/components/LayoutWithMenu/LayoutWithMenu.tsx
+++ b/apps/store/src/components/LayoutWithMenu/LayoutWithMenu.tsx
@@ -1,10 +1,11 @@
 import styled from '@emotion/styled'
-import { ReactElement } from 'react'
+import { type ReactElement } from 'react'
 import { FooterBlock } from '@/blocks/FooterBlock'
 import { HeaderBlock } from '@/blocks/HeaderBlock'
 import { PageStory, StoryblokPageProps } from '@/services/storyblok/storyblok'
 import { filterByBlockType, isProductStory } from '@/services/storyblok/Storyblok.helpers'
 import { useChangeLocale } from '@/utils/l10n/useChangeLocale'
+import { BreadcrumbList, BreadcrumbListItem } from './BreadcrumbList'
 import { GlobalProductMetadata, GLOBAL_PRODUCT_METADATA_PROP_NAME } from './fetchProductMetadata'
 import { useHydrateProductMetadata } from './ProductMetadataContext'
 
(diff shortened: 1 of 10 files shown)


robinandeer:
* main
  * **PR #2492** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/2492" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/2492?utm_source=stack-comment).

apps/store/src/components/LayoutWithMenu/fetchBreadcrumbs.ts
```diff
@@ -0,0 +1,28 @@
...
+): Promise<Array<BreadcrumbListItem>> => {
+  const absoluteSlug = slug.startsWith('/') ? slug : `/${slug}`
+
+  const parentSlugs = absoluteSlug
+    .split('/')
+    .slice(0, -1)
+    .map((_, index, array) => array.slice(0, index + 1).join('/'))
+    .filter((item) => item !== slug)
+
+  const parentStories = await Promise.allSettled(
```
alebedev, line 17: Nit: why not `Promise.all`?

Also, maybe you can use "slug in (a,b,c)" type of filter to load everything with single query?

https://www.storyblok.com/docs/api/content-delivery/v2#filter-queries/operation-in

apps/store/src/components/LayoutWithMenu/LayoutWithMenu.tsx
```diff
@@ -19,46 +20,52 @@ type LayoutWithMenuProps = {
...
 
-  useHydrateProductMetadata(children.props[GLOBAL_PRODUCT_METADATA_PROP_NAME])
+  useHydrateProductMetadata(props.children.props[GLOBAL_PRODUCT_METADATA_PROP_NAME])
   const handleLocaleChange = useChangeLocale(story)
 
   const headerBlock = filterByBlockType(globalStory.content.header, HeaderBlock.blockName)
   const footerBlock = filterByBlockType(globalStory.content.footer, FooterBlock.blockName)
 
-  const showFooter = !hideFooter && !story?.content.hideFooter
+  const showHeader = !story?.content.hideMenu
```
alebedev, line 39: Should we rename it in Storyblok too?

alebedev (approved):

apps/store/src/components/LayoutWithMenu/LayoutWithMenu.tsx
robinandeer: hmmm not sure actually, in Storyblok the default is "show" and we have an option to "hide" so I think that's correct

but I like to keep JSX as simple as possible which is why I transform it here - makes sense?

apps/store/src/components/LayoutWithMenu/fetchBreadcrumbs.ts
robinandeer: aha interesting -- will check that out

I can't use `Promise.all` since the code throws exception for slugs that don't exist (404). But we don't want to fail - just exclude those paths in the breadcrumb list.

robinandeer:
@robinandeer started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/2492).

PR #2511, opened 2023-06-08
fix(ManyPetsMigrationPage): make announcements optional

guilhermespopolin:
## Describe your changes

Fix an issue while trying to render Many Pets Migration page when no announcements are configured for that page. 

I've forgot to properly type/check for the presence of announcements for Many Pets Migration page so trying to render it without giving it an announcement would break the code - cannot map over underfined.

https://hedviginsurance.slack.com/archives/C054YL76SSU/p1686218003173419


robinandeer (approved):
Added via Giphy

PR #2514, opened 2023-06-08
Make background optional in HeroBlock

gustaveen:
Request from William

### DIFF
--- a/apps/store/src/blocks/HeroBlock.tsx
+++ b/apps/store/src/blocks/HeroBlock.tsx
@@ -8,7 +8,7 @@ import { filterByBlockType } from '@/services/storyblok/Storyblok.helpers'
 type VerticalAlignment = 'top' | 'center' | 'bottom'
 
 type HeroBlockProps = SbBaseBlockProps<{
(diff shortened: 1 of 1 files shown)


gustaveen:
* main
  * **PR #2514** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/2514" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/2514?utm_source=stack-comment).

robinandeer (approved):
Nice refactor!

I wonder if it makes sense to setup storybook stories for each of the different states we support

I feel like as we keep adding CMS props we will end up with so many possible states it is going to be hard to keep track of...

gustaveen:
> Nice refactor!
> 
> I wonder if it makes sense to setup storybook stories for each of the different states we support
> 
> I feel like as we keep adding CMS props we will end up with so many possible states it is going to be hard to keep track of...

I totally agree! I'll look into that

PR #2521, opened 2023-06-12
GEN-527 - fix(Header): update menu overlay implementation so it takes banners/announcements into account

guilhermespopolin:
## Describe your changes

Updates _menu overlay_ implementation so it uses negative margins instead of absolute positioning.

Using absolute positioning to implement _menu overlay_ means to remove the menu from doc's flow which makes everything else on the page be pulled up. However that causes an issue when we have _banners/announcements_ on the screen (see image below). So what we actually want is to pull page's content by the exact menu's height. Negative margins are perfect for that use case since we don't need to do complex calculations for positioning the menu accordingly.

Screenshot 2023-06-12 at 09 48 24

Screenshot 2023-06-12 at 11 52 42

### DIFF
--- a/apps/store/src/components/Header/Header.tsx
+++ b/apps/store/src/components/Header/Header.tsx
@@ -49,10 +49,9 @@ export const Header = (props: HeaderProps) => {
   const scrollState = useScrollState({ threshold: MENU_BAR_HEIGHT_PX * 2 })
   const { t } = useTranslation('common')
 
-  const defaultPosition = overlay ? 'absolute' : 'relative'
   const backgroundColor = opaque ? theme.colors.backgroundStandard : TRANSPARENT_HSL_COLOR
 
-  const initialStyles = { position: defaultPosition, backgroundColor } as const
+  const initialStyles = { backgroundColor } as const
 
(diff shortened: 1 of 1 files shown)


apps/store/src/components/Header/Header.tsx
```diff
@@ -81,15 +80,20 @@ export const Header = (props: HeaderProps) => {
   )
 }
 
-const GhostWrapper = styled.div({
-  top: 0,
-  left: 0,
-  right: 0,
+const GhostWrapper = styled.div<{ overlay: boolean }>(({ overlay }) => ({
+  '--height': HEADER_HEIGHT_MOBILE,
```
alebedev, line 84: Is this compiled to local var with different name?  Otherwise let's make it more specific

alebedev (approved):
Nice CSS solution!

apps/store/src/components/Header/Header.tsx
guilhermespopolin: It's not compiled to a local var but it will not interfere with "outside world". I mean, `GhostWrapper`'s children can access it if they want to, but they also can redefine it without causing any problems to `GhostWrapper` .

PR #2534, opened 2023-06-13
Feat / Add new MediaListBlock that supports both Image and Video

gustaveen:
## Describe your changes
Extend VideoListBlock to support ImageBlock as well. 
Instead of trying to rename existing block I created a new block called MediaListBlock.
Once migrated, we can remove VideoListBlock

https://github.com/HedvigInsurance/racoon/assets/6661511/a377df19-86a4-41bf-a592-f100bd90a15e

We wanted to support images in VideoListBlock


robinandeer (approved):
Very nice 🎯

PR #2537, opened 2023-06-13
GEN-468 - chore: redirect [no|no-en|dk|dk-en]/new-member --> [no|no-en|dk|dk-en]/info

guilhermespopolin:
**This should wait for Peters 'go ahead'**

## Describe your changes

Redirect `new-member` routes to info pages. That following should apply:

`[no|no-en|dk|dk-en]/new-member --> [no|no-en|dk|dk-en]/info`

Part of _Home Coming_ project.


guilhermespopolin:
* main
  * **PR #2537** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/2537" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/2537?utm_source=stack-comment).

robinandeer (commented):
It seems like we also redirect `/se/new-member` - is that intended? Will it cause any issues?

guilhermespopolin:
> It seems like we also redirect `/se/new-member` - is that intended? Will it cause any issues?

You're right. I thought it wouldn't be a problem but it is. Have changed.

alebedev (approved):
Looks good

apps/store/next.config.js
```diff
@@ -92,6 +92,16 @@ module.exports = withBundleAnalyzer({
       ],
     }
   },
+  async redirects() {
+    return [
+      {
+        source: '/(no|no-en|dk|dn-en)/new-member/:slug*',
+        destination: '/info',
```
alebedev, line 99: Will it keep original locale?  Let's test

guilhermespopolin: While testing I could verify that the original locale is kept. About the destination I need to double check tho.

robinandeer: Yeah I think the flow is Redirect -> Not found on racoon -> Rewrite (to old hedvig-com)

robinandeer (commented):
Hmmmm 🤔 now it redirects `/no/new-member` to `/info` (without locale)

Are we sure it's working correctly?

apps/store/next.config.js
```diff
@@ -92,6 +92,34 @@ module.exports = withBundleAnalyzer({
       ],
     }
   },
+  async redirects() {
```
guilhermespopolin, line 95: So I've change it as described in a comment:

1. Redirect accordingly
2. Leave for the rewrite definition, the last item of `fallback` to be more precise, to change the origin to market web.

However it's not working and I'm not sure how to debug redirects. Do you guys have an idea?
I've checked that we have `FALLBACK_ORIGIN_URL` configured for preview apps so I believe `fallback` redirects are being returned for it...

image

guilhermespopolin: This is solved!

apps/store/src/utils/Features.ts
```diff
@@ -14,6 +14,7 @@ const config = {
   MANYPETS_MIGRATION: process.env.NEXT_PUBLIC_FEATURE_MANYPETS_MIGRATION === 'true',
   SAS_PARTNERSHIP: process.env.NEXT_PUBLIC_FEATURE_SAS_PARTNERSHIP === 'true',
   INSURELY_CAR: process.env.NEXT_PUBLIC_FEATURE_INSURELY_CAR === 'true',
+  HOMECOMING: process.env.NEXT_PUBLIC_FEATURE_HOMECOMING === 'true',
```
robinandeer, line 17: guess this is technically not needed but why not ;)

robinandeer (approved):
Nice!

apps/store/src/utils/Features.ts
guilhermespopolin: Yes, that's right. I can't use it in the config file but thought it could exist here as well.

PR #2558, opened 2023-06-15
Display custom message for incomparable price match

robinandeer:
## Describe your changes

- Display a custom message when the user gets price matched but with a larger scope than the competitor (e.g. 2 vs 3 co-insured)

- Update `DiscountTooltip` with option to display a gray variant

- Right now we display a confusing message like "You save -20 kr" in these cases

### DIFF
--- a/apps/store/public/locales/en/purchase-form.json
+++ b/apps/store/public/locales/en/purchase-form.json
@@ -107,6 +107,8 @@
   "OPEN_PRICE_CALCULATOR_BUTTON": "See your price",
   "PRESENT_OFFER_EDIT_BUTTON": "Edit your info",
   "PRICE_CALCULATOR_SECTION_EDIT_BUTTON": "Edit",
+  "PRICE_MATCH_BUBBLE_INCOMPARABLE_SUBTITLE": "We couldn't match your price since you have entered a larger scope",
+  "PRICE_MATCH_BUBBLE_INCOMPARABLE_TITLE": "You pay {{amount}} at {{company}}",
   "PRICE_MATCH_BUBBLE_SUCCESS_SUBTITLE": "compared to {{company}}",
   "PRICE_MATCH_BUBBLE_SUCCESS_TITLE": "You save {{amount}}",
   "QUICK_CHECKOUT_BUTTON_LABEL": "Go to checkout",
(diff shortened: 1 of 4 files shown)


robinandeer:
* main
  * **PR #2558** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/2558" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/2558?utm_source=stack-comment).

apps/store/src/components/ProductPage/PurchaseForm/DiscountTooltip/DiscountTooltip.tsx
```diff
@@ -4,22 +4,28 @@ import { Text, theme } from 'ui'
 type Props = {
```
alebedev, line 4: Since it's quite tricky to setup real test, can you add Storybook for new mode?  And perhaps old one if it's not available there yet

apps/store/public/locales/en/purchase-form.json
```diff
@@ -107,6 +107,8 @@
   "OPEN_PRICE_CALCULATOR_BUTTON": "See your price",
   "PRESENT_OFFER_EDIT_BUTTON": "Edit your info",
   "PRICE_CALCULATOR_SECTION_EDIT_BUTTON": "Edit",
+  "PRICE_MATCH_BUBBLE_INCOMPARABLE_SUBTITLE": "We couldn't match your price since you have entered a larger scope",
```
alebedev, line 110: Not sure this makes sense to me as potential buyer.  "Your coverage at Hedvig would be larger and therefor our price is higher", maybe?  I assume it also has to be legally correct in addition to being clear

alebedev (approved):
Very nice!

alebedev (approved):
Very nice!

apps/store/public/locales/en/purchase-form.json
robinandeer: Yes, we should iterate on the phrasing for sure

But this should ideally be owned by Peter so I will defer to him 🙂

gustaveen (approved):
Nice fix! 🙌

PR #2585, opened 2023-06-19
Insurely: use native close button

robinandeer:
## Describe your changes

- Use native Insurely close button instead of overlaying our own

- Center Insurely iframe on small devices based on screen height

- The native close button looks more native we don't need to worry about
  positioning it correctly

- We can't always center the iframe since the content might be longer than the
  screen height. But if the content is shorter than the screen height, we should center it.


guilhermespopolin (approved):
🚀

PR #2602, opened 2023-06-21
Only download SE/EN locales from Lokalise

robinandeer:
## Describe your changes

- Only download en and sv translations from Lokalise

- Homecoming

- We should do a bigger PR and remove the locales from the rest of the code base as well


alebedev (approved):
Nice cleanup
Should we delete dictionaries as well?

robinandeer:
> Nice cleanup
> Should we delete dictionaries as well?

I was thinking about that but I think it might break next i18n setup. So we can do a PR that just cleans up those locales entierly.

PR #2603, opened 2023-06-21
fix(middleware): make sure query params are preserved/merged when redirecting

guilhermespopolin:


guilhermespopolin:
* main
  * **PR #2603** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/2603" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/2603?utm_source=stack-comment).

apps/store/tsconfig.json
```diff
@@ -23,7 +23,8 @@
         "name": "next"
       }
     ],
-    "strictNullChecks": true
+    "strictNullChecks": true,
+    "downlevelIteration": true
```
guilhermespopolin, line 27: Without this I couldn't use iterators in the file above ☝🏻

apps/store/src/middleware.ts
```diff
@@ -88,9 +88,17 @@ const redirectMiddleware = async (req: NextRequest): Promise<NextResponse | unde
       console.log(
         `Applying ${redirect.temporary ? 'temporary' : 'permanent'} redirect ${redirect.from} -> ${
           redirect.to
-        }`,
+        } - Query params are preserved/merged`,
       )
+
       const redirectUrl = new URL(redirect.to, req.nextUrl)
+      // Make sure query params are preserved/merged
```
robinandeer, line 95: nit: avoid "what" comments -- ideally the code should be self-documenting. Otherwise it's another line which need to be maintained and updated when we change what we do in the code block.

"why" comments are okey tho 🙂

robinandeer (commented):
Nice work! I wonder tho; can we do this differently? 

1. Clone the `reqUrl`
2. Update the pathname to `redirect.to`
(3. Perhaps delete already consumed query params)

Done! No iteration needed

apps/store/src/middleware.ts
guilhermespopolin: Makes sense. Thanks for the input.

guilhermespopolin:
> Nice work! I wonder tho; can we do this differently?
> 
> 1. Clone the `reqUrl`
> 2. Update the pathname to `redirect.to`
>    (3. Perhaps delete already consumed query params)
> 
> Done! No iteration needed

By doing so we'd keep search params present in request URL, but query params present in redirect URL would be ignored, no?

robinandeer:
> > Nice work! I wonder tho; can we do this differently?
> > 
> > 1. Clone the `reqUrl`
> > 2. Update the pathname to `redirect.to`
> >    (3. Perhaps delete already consumed query params)
> > 
> > Done! No iteration needed
> 
> By doing so we'd keep search params present in request URL, but query params present in redirect URL would be ignored, no?

aha yeah I understand... do we support redirects with query params in the edge config? Perhaps ask @alebedev what the idea was there :)

guilhermespopolin:
> > > Nice work! I wonder tho; can we do this differently?
> > > 
> > > 1. Clone the `reqUrl`
> > > 2. Update the pathname to `redirect.to`
> > >    (3. Perhaps delete already consumed query params)
> > > 
> > > Done! No iteration needed
> > 
> > 
> > By doing so we'd keep search params present in request URL, but query params present in redirect URL would be ignored, no?
> 
> aha yeah I understand... do we support redirects with query params in the edge config? Perhaps ask @alebedev what the idea was there :)

Yeah, we do. There's a SAS redirect URL that includes query parameters.
https://vercel.com/dashboard/hedvig/stores/edge-config/ecfg_viia3ntdquafelygrc3g6lfs3d2a/items

apps/store/src/middleware.ts
```diff
@@ -88,9 +88,17 @@ const redirectMiddleware = async (req: NextRequest): Promise<NextResponse | unde
...
+      for (const [key, value] of reqUrl.searchParams.entries()) {
+        if (!redirectUrl.searchParams.has(key)) {
```
alebedev, line 97: Is it needed?  I thought by default merging `?a=1` and `?a=2` will product `?a=1&a=2` which semantically correct

alebedev (approved):
Looks good

PR #2672, opened 2023-07-03
Add field for specifying AB test origin story in Storyblok

gustaveen:
## Describe your changes
- Add field in Page config to specify A/B test origin Story
- Use `abTestOriginStory` in `HeadSeoInfo` and pass canonical url
- Remove `canonicalUrl` field in Page config and use slug from `abTestOriginStory`

Screenshot 2023-07-03 at 17 20 58

Screenshot 2023-07-03 at 17 02 06

To avoid that Google index the A/B test variants as a new page, we should use the meta data from the origin page and add a canonical tag

### DIFF
--- a/apps/store/src/components/HeadSeoInfo/HeadSeoInfo.tsx
+++ b/apps/store/src/components/HeadSeoInfo/HeadSeoInfo.tsx
@@ -6,14 +6,19 @@ import { isRoutingLocale, toIsoLocale } from '@/utils/l10n/localeUtils'
 import { ORIGIN_URL } from '@/utils/PageLink'
 import { StructuredDataOrganization } from './StructuredDataOrganization'
 
-export const HeadSeoInfo = ({ story }: { story: ISbStoryData<SEOData> }) => {
+type Props = {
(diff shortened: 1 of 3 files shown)


gustaveen:
* main
  * **PR #2672** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/2672" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/2672?utm_source=stack-comment).

robinandeer (approved):
Nice work!

The only thing I would consider is if this needs to be a "a/b test" feature 🤔 But it's probably a good idea to make it very explicit right now to not generalize too early.

gustaveen:
> Nice work!
> 
> The only thing I would consider is if this needs to be a "a/b test" feature 🤔 But it's probably a good idea to make it very explicit right now to not generalize too early.

Gotcha! Couldn't think of another use case at the moment but let's rename it if we will use it more generally 👍 @robinandeer

PR #2675, opened 2023-07-04
Use videos from S3 bucket to save traffic usage on Storyblok

gustaveen:
## Describe your changes
- Add field for `videoUrl` to replace asset field
- Setup new S3 bucket for `hedvig-dot-com` content
- Add deprecation flag in Storyblok for video field

Screenshot 2023-07-04 at 15 31 43

To consider:
We currently only use one video format in sources, mp4. We should probably use webm as well. Should we add another field for that in Storyblok? In market-web we relied on the filename being the same and just added the file format to the path but that leaves for some errors if we wouldn't upload both formats in S3. 

We used around 3TB of traffic last month, most traffic is are due to videos stored in Storyblok. We currently pay $60 per 250GB for Storyblok traffic compared to $5.75 if we store it on S3.

### DIFF
--- a/apps/store/next-csp.config.js
+++ b/apps/store/next-csp.config.js
@@ -93,6 +93,7 @@ const mediaSrc = [
   'https://dc.insurely.com',
   'https://vercel.live',
   'https://a.storyblok.com',
+  'https://s3.eu-central-1.amazonaws.com',
   "'self'",
 ]
 const connectSrc = [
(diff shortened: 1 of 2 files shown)


gustaveen:
* main
  * **PR #2675** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/2675" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/2675?utm_source=stack-comment).

alebedev (approved):
Looks good and straighforward.  But how are we going to upload videos to S3?

robinandeer (approved):
Good stuff! Agree that we should investigate different video formats in the future

gustaveen:
> Looks good and straighforward. But how are we going to upload videos to S3?

I think we have to do it manually to start with, but let's see if we can find a good workflow for it. @alebedev

PR #2691, opened 2023-07-05
DON'T MEGE: GEN-644 - Contigency plan: revert discount related changes (ProductOffer.cost)

guilhermespopolin:
## Describe your changes

This is a contingency plan that reverts changes related with the addition of `ProductOffer.cost` field. It's here se it get's easy to revert in case the API shows any issues.

PR #2698, opened 2023-07-06
fix(CostSummary): define applied discount based on cost field

guilhermespopolin:
## Describe your changes

Update helper functions used to provide data to `CostSummary` so they rely on `cost` field while identifying if there's some discount being applied to your purchase.

**Before**

after

**After**

before

So the Cost summary doesn't look so confusing since it might be some cases where the product you have added into the cart is not covered by the campaign code you've added.

Also going to update the [contingency plan](https://github.com/HedvigInsurance/racoon/pull/2691) by removing these changes after this get's merged.

### DIFF
--- a/apps/store/src/components/CartInventory/CartInventory.helpers.tsx
+++ b/apps/store/src/components/CartInventory/CartInventory.helpers.tsx
@@ -46,7 +46,7 @@ export const getTotal = (shopSession: ShopSession) => {
 }
 
 export const getCrossOut = (shopSession: ShopSession) => {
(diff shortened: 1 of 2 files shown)


guilhermespopolin:
* main
  * **PR #2698** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/2698" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/2698?utm_source=stack-comment).

gustaveen (approved):
LGTM!

PR #2745, opened 2023-07-11
Store Storyblok cache version in environment variable

robinandeer:
## Describe your changes

- During the build, read latest Storyblok cache version and store in env variable

- Remove logic related to reading and writing cache version to Edge Config

- Edge config is eventually consistent so it's not ideal to store the cache version there if we need to read it again right after


robinandeer:
* main
  * **PR #2745** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/2745" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/2745?utm_source=stack-comment).

bin/vercel-ignore-step.sh
```diff
@@ -13,4 +13,4 @@ if [ "$VERCEL_GIT_COMMIT_SHA" = "$VERCEL_GIT_PREVIOUS_SHA" ]; then
   exit 1
 fi
 
-yarn dlx turbo-ignore
+npx turbo-ignore
```
alebedev, line 16: Why npx?

alebedev (approved):
Looks good

bin/vercel-ignore-step.sh
robinandeer: it's what vercel recommended -- however, I don't think it fixed the problems I was seeing so let's revert

PR #2748, opened 2023-07-12
GEN-664 - feat: added MobileComparisonTable

guilhermespopolin:
## Describe your changes

* Adds a new Comparison Table component to be used on smaller viewports. It leverages tabs to compare coverage levels between tiers.

### How the whole thing is gonna work

In order to achieve the end result, we'll have two versions of it: _Mobile_ and _Desktop_. We can then show them accordingly based on the available viewport. For that to work smoothly both versions share the same data structure for table representation:

```
type Table = {
  head: Head
  body: Body
}

type Row = string[]
type Head = string[]
type Body = Row[]

```

Inside each version we can then _massage_ the data to get the correct visual representation of it. This PR is just about adding the _Mobile_ version but it also includes some foundation to be used for the _Desktop_ version as well (down the stack). Below one can find a video with the end result:

https://github.com/HedvigInsurance/racoon/assets/19200662/5374be19-03a0-4ae5-a353-1b77eaa90406

We got a new [design](https://www.figma.com/file/qUhLjrKl98PAzHov9ilaDH/Hedvig-UI-Kit?type=design&node-id=3380-13894&mode=design&t=snZZXMBD7zxyZULO-4) for a _mobile_ version of the comparison table:
image


guilhermespopolin:
* main
  * **PR #2748** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/2748" target="_blank">Graphite</a>  👈
    * **PR #2749** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/2749" target="_blank">Graphite</a> 
      * **PR #2750** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/2750" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/2748?utm_source=stack-comment).

apps/store/src/components/ComparisonTable/ComparisonTable.helpers.tsx
```diff
@@ -0,0 +1,20 @@
+import { CheckIcon, MissingIcon } from './ComparisonTable'
+import { TableMarkers } from './ComparisonTable.types'
+
+export const getCellValue = (value: string) => {
+  if (value === TableMarkers.Covered)
+    return (
+      // TODO: Lokalise this
+      <span aria-label="Covered">
+        <CheckIcon />
```
robinandeer, line 9: Can we add aria-label directly to SVG?

apps/store/src/components/ComparisonTable/ComparisonTable.types.ts
```diff
@@ -0,0 +1,16 @@
+export const TableMarkers = {
```
robinandeer, line 1: Nice!

apps/store/src/components/ComparisonTable/MobileComparisonTable.tsx
```diff
@@ -0,0 +1,117 @@
...
+  head: Head
+  body: Body
+  defaultSelectedTier?: string
+  className?: string
+}
+
+export const MobileComparisonTable = ({
+  head,
+  body,
+  defaultSelectedTier,
```
robinandeer, line 21: Can we name this in a more generic way? Like selected `defaultSelectedColumn`

apps/store/src/components/ComparisonTable/MobileComparisonTable.tsx
```diff
@@ -0,0 +1,117 @@
...
+  className,
+}: MobileComparisonTableProps) => {
+  const [tierIndex, setTierIndex] = useState(() => {
+    if (defaultSelectedTier) {
+      return head.findIndex((headerValue) => headerValue === defaultSelectedTier)
+    }
+
+    // First header value is usually an Markers.EmptyHeader.
+    return 1
+  })
```
robinandeer, line 31: A different data structure might avoid this conditional logic

apps/store/src/components/ComparisonTable/MobileComparisonTable.tsx
```diff
@@ -0,0 +1,117 @@
...
+      {head.map((headerValue) => {
+        if (headerValue === TableMarkers.EmptyHeader) return null
+
+        return (
+          <TabContent key={headerValue} value={headerValue} data-value={headerValue}>
+            <ComparisonTable.Root>
+              <ComparisonTable.Body>
+                {body.map((row) => {
+                  const attribute = row[0]
+                  const value = row[tierIndex]
```
robinandeer, line 62: What if these are empty?

robinandeer (approved):
Neat stuff!

apps/store/src/components/ComparisonTable/ComparisonTable.helpers.tsx
guilhermespopolin: Ofc; What was I thinking?! 😅

gustaveen (commented):
Great work with this @guilhermespopolin It works nicely! 

Maybe I missed if there was a case where using the same table solution didn't work, was just quickly testing to `display: none` for the non `active` cells of the current solution. And then in mobile we could just not render the table header and display buttons that toggles the active state. 

But I'm all good with having two components as well 🙌

https://github.com/HedvigInsurance/racoon/compare/test/comparison-table


https://github.com/HedvigInsurance/racoon/assets/6661511/10bc5dcc-9bbd-4da1-b789-95fd438c024d

guilhermespopolin:
> Great work with this @guilhermespopolin It works nicely!
> 
> Maybe I missed if there was a case where using the same table solution didn't work, was just quickly testing to `display: none` for the non `active` cells of the current solution. And then in mobile we could just not render the table header and display buttons that toggles the active state.
> 
> But I'm all good with having two components as well 🙌
> 
> https://github.com/HedvigInsurance/racoon/compare/test/comparison-table

I see. So would you change it directly in `ComparisonTable` or create a new _responsive table_ component that changes `ComparisonTable` styles to achieve the end result? Because today `ComparisonTable` offers only primitives to build tables and for this particular scenario we also want to display buttons as table header's when on mobile. I don't see how we can change `ComparisonTable.Header` to show buttons instead of regular table headers on small screens without making it too complex 🤔

gustaveen:
> > Great work with this @guilhermespopolin It works nicely!
> > Maybe I missed if there was a case where using the same table solution didn't work, was just quickly testing to `display: none` for the non `active` cells of the current solution. And then in mobile we could just not render the table header and display buttons that toggles the active state.
> > But I'm all good with having two components as well 🙌
> > https://github.com/HedvigInsurance/racoon/compare/test/comparison-table
> 
> I see. So would you change it directly in `ComparisonTable` or create a new _responsive table_ component that changes `ComparisonTable` styles to achieve the end result? Because today `ComparisonTable` offers only primitives to build tables and for this particular scenario we also want to display buttons as table header's when on mobile. I don't see how we can change `ComparisonTable.Header` to show buttons instead of regular table headers on small screens without making it too complex 🤔

Right, might be better to keep the primitives intact and make a new component for that 👍

apps/store/src/components/ComparisonTable/MobileComparisonTable.tsx
guilhermespopolin: Do you have other in mind by chance? All the ones I could think of requires this reasoning behind it..

apps/store/src/components/ComparisonTable/MobileComparisonTable.tsx
guilhermespopolin: In that case we'd rendering nothing as table cell content. It kind of relies that the provided table table be complete. I can't see how can we make fail proof. I mean, I could at least log that situation so it become easier to debug if necessary. Would you think that this would be enought?

apps/store/src/components/ComparisonTable/MobileComparisonTable.tsx
robinandeer: Yeah I should have mentioned it's just a thought --- let's go with your solution!

guilhermespopolin:
@guilhermespopolin started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/2748).

guilhermespopolin:
@guilhermespopolin merged this pull request with [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/2748).

PR #2763, opened 2023-07-14
GEN-751 | Add Retargeting CMS block

robinandeer:
## Describe your changes

- Implement CMS block for listing retargeting offers

- Refactor `/session/resume/:id` page

  - Fetch predefined Story from CMS
  - Use ShopSession from URL params

- Collect all retargeting code inside `retargeting` feature folder

- With a CMS block - editors can customize the rest of the page just like they want it

- I tried to preload all the offers (price intents) and pass to Apollo cache but that made the initial payload so big that Next.js warned me about it hurting performance. So I decided to fetch the offers client side instead and implement a simple loading state.


robinandeer:
* main
  * **PR #2763** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/2763" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/2763?utm_source=stack-comment).

apps/store/src/features/retargeting/useRetargetingOffers.ts
```diff
@@ -0,0 +1,53 @@
+import { useMemo } from 'react'
+import { type ComponentProps } from 'react'
```
guilhermespopolin, line 2: nit:
```suggestion
import { useMemo, type ComponentProps } from 'react'
```

apps/store/src/features/retargeting/useRetargetingOffers.ts
```diff
@@ -0,0 +1,53 @@
...
+
+  const result = usePriceIntentsQuery({
+    skip: !shopSession,
+    variables: shopSession ? { shopSessionId: shopSession.id } : undefined,
+  })
+
+  const offers = useMemo(() => {
+    if (!shopSession || !result.data) return null
+
+    const cartOffers = new Set(result.data.shopSession.cart.entries.map((item) => item.id))
```
guilhermespopolin, line 23: nit:
```suggestion
    const cartOffersIds = new Set(result.data.shopSession.cart.entries.map((item) => item.id))
```

apps/store/src/features/retargeting/useRetargetingOffers.ts
```diff
@@ -0,0 +1,53 @@
...
+
+    return result.data.shopSession.priceIntents.reduce<Array<Offer>>((total, item) => {
+      if (item.offers.some((offer) => cartOffers.has(offer.id))) {
```
guilhermespopolin, line 26: Just so I better understand the feature: Why do we discard all offers of a price intent if we found at least one that's already added to the cart?

guilhermespopolin (commented):
Based on the code, editors should be able to preview `session/resume` page in storyblok with _skeletons_ replacing real cart entries. However I couldn't preview it while testing locally. Is there something I missed?

image

apps/store/src/features/retargeting/useRetargetingOffers.ts
robinandeer: That's how price intents work, no? You confirm it -- get a list of offers to pick from -- then add one to the cart :)

You can't add both HALF and FULL car insurance for example

guilhermespopolin: Aha, yes ofc. Makes totally sense. I've failed to put into perspective. Thanks for the clarification.

guilhermespopolin (approved):
Well done

robinandeer:
> Based on the code, editors should be able to preview `session/resume` page in storyblok with _skeletons_ replacing real cart entries. However I couldn't preview it while testing locally. Is there something I missed?
> 
> image
> 

Hmmm could it be some problems requesting draft data?

gustaveen (approved):
Great work on this feature!  
Same problem for me, couldn't preview but I guess that's because it's a new page?

PR #2770, opened 2023-07-14
fix language switcher

guilhermespopolin:
## Describe your changes

Uses _full page reload_ on language switcher

We're having a bug where product cards just disappear while changing languages. Causing a _full page reload_ is probably the best option when it comes about changing app's locale.

### DIFF
--- a/apps/store/src/utils/l10n/useChangeLocale.ts
+++ b/apps/store/src/utils/l10n/useChangeLocale.ts
@@ -32,15 +32,12 @@ export const useChangeLocale = (currentPageStory: ISbStoryData | undefined) => {
           window.location.href = `/${routingLocale}`
           return
         }
+
         const targetAlternate = currentPageStory?.alternates.find((alternate) =>
           alternate.full_slug.startsWith(routingLocale),
         )
-        if (targetAlternate) {
-          router.push(targetAlternate.full_slug, undefined, { locale: routingLocale })
-        } else {
-          // Base case, failed to find alternate in new language
-          router.push('/', undefined, { locale: routingLocale })
-        }
+        const url = targetAlternate ? targetAlternate.full_slug : `/${routingLocale}`
+        window.location.pathname = url
       } catch (error) {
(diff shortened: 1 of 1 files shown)


guilhermespopolin:
* main
  * **PR #2770** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/2770" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/2770?utm_source=stack-comment).

apps/store/src/utils/l10n/useChangeLocale.ts
```diff
@@ -32,15 +32,12 @@ export const useChangeLocale = (currentPageStory: ISbStoryData | undefined) => {
...
         const targetAlternate = currentPageStory?.alternates.find((alternate) =>
           alternate.full_slug.startsWith(routingLocale),
         )
-        if (targetAlternate) {
-          router.push(targetAlternate.full_slug, undefined, { locale: routingLocale })
-        } else {
-          // Base case, failed to find alternate in new language
-          router.push('/', undefined, { locale: routingLocale })
-        }
+        const url = targetAlternate ? targetAlternate.full_slug : `/${routingLocale}/`
```
robinandeer, line 39: Is there any difference reading pathname from router?

robinandeer (approved):
Nice looks good to me -- and worked when I tried to switch language

apps/store/src/utils/l10n/useChangeLocale.ts
guilhermespopolin: 🤔 you mean keep the `router.push` and them also change `windonw.location`?

robinandeer: Ahhh no just _reading_ pathname from router instead of window

This is an event handler so shouldn't matter though. Otherwise "router" is the only thing which will trigger rerenders correctly.

PR #2851, opened 2023-07-25
feat: enable next app router

guilhermespopolin:
## Describe your changes

* Adds a workaround so we can start using Next App Router (app folder) while having _emotion compiler_ enabled. `app/layout.tsx` can be a server component now.

After a some  [in depth investigation](https://www.notion.so/hedviginsurance/Adopting-App-router-61b33e6d19df45f79237cc596a851d48?pvs=4#d9820ba991e34687a75cb8fa86f4ec3d) we concluded a reasonable first step towards _app router_ when it comes about styling would be just to be able to use it with our current setup, even with some limitations (more on that later). This is a reasonable workaround towards that which requires the addition of `/** @jsxImportSource react */` at the beginning of _Server Components_ files. That instructs _nextjs_ compiler to transform those files as regular React components instead of using _emotion jsx_ transformation. Using the later for Server components breaks the build because the end result uses features not supported by Server Components like _React Context_.

It's expected that `emotion` provides support for _React Server Components_; We also can decide upon moving away from `emotion` as other libraries like [_chakra UI_](https://www.adebayosegun.com/blog/the-future-of-chakra-ui) seems to be doing. Until then we just need to follow 2 rules:

1. Make sure to add  `/** @jsxImportSource react */` into _Server Components_ files.
2. To turn every other file that loads emotion in some way into a client component with `"use client"` directive.


guilhermespopolin:
* main
  * **PR #2851** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/2851" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/2851?utm_source=stack-comment).

robinandeer (approved):
Amazing!!!

Annoying but a great first step :))

PR #2853, opened 2023-07-26
chore: replace nth-child --> nth-of-type since it is safier when SSR

guilhermespopolin:
## Describe your changes

* Replace nth-child --> nth-of-type since it is safier when SSR

Screenshot 2023-07-26 at 16 57 53


alebedev (approved):
Good catch!

PR #2892, opened 2023-08-04
fix(GlobalBanner): make sure it only renders client side avoiding hydration errors

guilhermespopolin:
## Describe your changes

* Make sure `GlobalBanner` only gets rendered client side as it depends on storage based atom. By doing so we can avoid hydration errors.

We've been getting [hydration error logs](https://app.datadoghq.eu/rum/error-tracking?query=%40application.id%3Ad86e86d8-8d9a-4b00-a2cb-10b8d7671fa2&from_ts=1691154363205&to_ts=1691157963205&live=true) for a while now. I could track them down to a particular [deployment in prod](https://github.com/HedvigInsurance/racoon/commit/481cf34aaf7ef5b049e29a6201f574e08b6c0448) where I've changed `GlobalBanner` atom state to leverage `sessionStorage`. That was causing the hydration errors. Here I'm using the [Jotai's suggested solution](https://jotai.org/docs/utilities/storage#server-side-rendering) to fix the issue.

### DIFF
--- a/apps/store/src/components/GlobalBanner/GlobalBanner.tsx
+++ b/apps/store/src/components/GlobalBanner/GlobalBanner.tsx
@@ -3,7 +3,7 @@ import { Banner } from '@/components/Banner/Banner'
 import { useDebugShopSession } from '@/utils/useDebugShopSession'
 import { useGlobalBanner } from './useGlobalBanner'
 
-export const GlobalBanner = () => {
+const GlobalBanner = () => {
   const { banner, dismissBanner } = useGlobalBanner()
 
   useDebugShopSession()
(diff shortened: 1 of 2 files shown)


guilhermespopolin:
* main
  * **PR #2892** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/2892" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/2892?utm_source=stack-comment).

robinandeer (commented):
Great find!

However, I think we might be able to do this in a more Next.js-y way like this:

```
import dynamic from 'next/dynamic'

const ClientComponent = dynamic(() => import('@/path/to/Component'), { ssr: false })
```

what do you think of that?

alebedev (approved):
Neat!

PR #2961, opened 2023-08-14
hotfix(PurchaseForm): make sure we set a selectedOffer after confirming price intent

guilhermespopolin:
Disclaimer: That was magically working before, even without these changes. Looking at the commit history, it seems there's nothing that could be causing the issue despite one update on jotai. However I've also tried to revert that change but the issue persisted. I think we can start with this just to make the app work again and I can continue looking at it to spot the real issue.

## Describe your changes

* Make sure we set a selectedOffer after confirming price intent

Raised [here](https://hedviginsurance.slack.com/archives/CPCUHMMJQ/p1692001712502579)

### DIFF
--- a/apps/store/src/components/ProductPage/PurchaseForm/PurchaseForm.tsx
+++ b/apps/store/src/components/ProductPage/PurchaseForm/PurchaseForm.tsx
@@ -305,6 +305,7 @@ type EditingStateProps = {
 const EditingState = (props: EditingStateProps) => {
   const { shopSession, priceIntent, onComplete } = props
   const { t } = useTranslation('purchase-form')
+  const [, setSelectedOffer] = useSelectedOffer()
   const tracking = useTracking()
 
   const [confirmPriceIntent, result] = usePriceIntentConfirmMutation({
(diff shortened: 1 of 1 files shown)


guilhermespopolin:
* main
  * **PR #2961** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/2961" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/2961?utm_source=stack-comment).

apps/store/src/components/ProductPage/PurchaseForm/PurchaseForm.tsx
```diff
@@ -329,6 +330,8 @@ const EditingState = (props: EditingStateProps) => {
         tracking.setContext(TrackingContextKey.Customer, shopSession.customer)
         tracking.setPriceIntentContext(updatedPriceIntent)
         updatedPriceIntent.offers.forEach((offer) => tracking.reportOfferCreated(offer))
+        // TODO: set initially selectedOffer based on some logic; Ideally get it from the the API
+        setSelectedOffer(updatedPriceIntent.offers[0])
```
alebedev, line 334: What would happen with car/pet if you change something?  Wouldn't want updating startDate to reset selected tier

alebedev (approved):
Feel free to merge if car/pet case I commented about works as expected

apps/store/src/components/ProductPage/PurchaseForm/PurchaseForm.tsx
guilhermespopolin: That's true. I saw some logic in `PriceIntentContext` that handles that. But this would break that functionality. Gonna close this one.

PR #2982, opened 2023-08-16
GEN-860 - chore: bump @apollo/client 3.7.17 - 3.8.1

guilhermespopolin:
## Describe your changes

* Bump `@apollo/client`: `3.17.1` --> `3.8.1`
* Update  apollo's `useQuery` instances in our code that relies on `onCompleted` callback.

I've went through the code looking for usages of `useQuery`'s `onCompleted` and proper updated them.

There's some occurrences where I think we don't need to do anything:

```
apps/store/src/components/InitiateCarCancellationPage/InitiateCarCancellationPage.tsx:
  34:     onCompleted(data) {
apps/store/src/features/manyPets/ManyPetsMigrationPage/ManyPetsMigrationPage.tsx:
  73:     onCompleted() {

```

### Some context

More details can be found [here](https://github.com/apollographql/apollo-client/pull/10229) but in summary `@apollo/client@3.8.1` changes the circumstances when `useQuery`'s `onCompleted` callback gets executed. Before:

> writes to the cache (it doesn't matter how the cache update occurs: directly via cache.writeQuery, via a separate useMutation call, etc.) trigger the onCompleted callback passed to useQuery.

Now `onCompleted` execution is directly related with query's `result` loading states: if the loading state doesn't change, `onCompleted` won't re-fire.

### Why does our code doesn't just works with `3.8.1`?

We've been "exploiting" that previous behaviour to write "side effect" code using apollo client's API instead of `useEffects` because, you know, nobody likes `useEffects`. This new version, which seems to be implementing the expect behaviour when it comes  about `onCompleted`,  then means to us that we'd need to stop looking at `onCompleted` as a escape hatch for `useEffects`. It also means it's not connected to apollo's cache in a global sense anymore meaning it only cares about the query that defines it. With that said `3.8.1` defines basically two use cases for `onCompleted`:

1) When you want to execute a callback with the data retrieved by a **particular query** after **its first successful data retrieval from the network**;
2) When you want to execute a callback with the data retrieved by a **particular query** after **every successful data retrieval from the network**

For option 1) one only needs to provide a `onCompleted` callback function. For option 2), `notifyOnNetworkStatusChange ` should also be set to `true`.

### DIFF
--- a/apps/store/src/services/bankId/useBankIdCheckoutSign.ts
+++ b/apps/store/src/services/bankId/useBankIdCheckoutSign.ts
@@ -86,6 +86,9 @@ export const useBankIdCheckoutSignApi = ({ dispatch }: Options) => {
           fetchSigning({
             variables: { shopSessionSigningId },
             pollInterval: 1000,
+            // Make sure 'onCompleted' gets called during the whole pooling process
+            // https://github.com/apollographql/apollo-client/pull/10229
+            notifyOnNetworkStatusChange: true,
             async onCompleted(data) {
               if (subscriber.closed) return
               const { status, completion } = data.shopSessionSigning
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -6,7 +6,7 @@
   "labels": ["needs approval/merge"],
   // Don't update nextjs until client-side navigation issue is solved
   // https://github.com/vercel/next.js/issues/53967
-  "ignoreDeps": ["next", "eslint-config-next", "@next/bundle-analyzer", "@apollo/client"],
+  "ignoreDeps": ["next", "eslint-config-next", "@next/bundle-analyzer"],
   "packageRules": [
     {
       "matchDepTypes": ["dependencies"],
--- a/apps/store/package.json
+++ b/apps/store/package.json
@@ -21,7 +21,7 @@
   },
   "dependencies": {
     "@adyen/adyen-web": "3.23.0",
-    "@apollo/client": "3.7.17",
+    "@apollo/client": "3.8.1",
     "@apollo/experimental-nextjs-app-support": "0.4.1",
     "@datadog/browser-logs": "4.46.0",
     "@datadog/browser-rum": "4.46.0",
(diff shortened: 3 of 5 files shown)


guilhermespopolin:
* main
  * **PR #2982** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/2982" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/2982?utm_source=stack-comment).

apps/store/src/services/bankId/useBankIdCheckoutSign.ts
```diff
@@ -86,6 +86,9 @@ export const useBankIdCheckoutSignApi = ({ dispatch }: Options) => {
           fetchSigning({
             variables: { shopSessionSigningId },
             pollInterval: 1000,
+            // Make sure 'onCompleted' gets called during the whole pooling process
+            // https://github.com/apollographql/apollo-client/pull/10229
+            notifyOnNetworkStatusChange: true,
             async onCompleted(data) {
```
guilhermespopolin, line 92: **Why this doesn't work with 3.8.1 version**
Without `notifyOnNetworkStatusChange: true`, `onCompleted` function is being called only once: after the first successful sign information retrieval from the network. But we actually need to keep calling it because occasionally we'd get an "approved" status and all the remaining code about stop polling and move forward to the next page, which is placed  inside `onCompleted`, would then be executed.

alebedev:
Good work!  Let's wait until we solve BankSignering issue, just in case it's related somehow

gustaveen (approved):
GJ solving this! 🙌

guilhermespopolin:
@guilhermespopolin started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/2982).

guilhermespopolin:
[Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/2982) rebased this pull request as part of a merge.

guilhermespopolin:
@guilhermespopolin merged this pull request with [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/2982).

PR #3009, opened 2023-08-22
GEN-897 - feat: allow 'sizes' property to be added to ImageBlock and ProductGrid

guilhermespopolin:
## Describe your changes

- Update `ImageBlock` so it can accept a [`sizes`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/sizes) property. We can use that to better leverage `next/image` features for [_responsive images_](https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Responsive_images).
- Define a `sizes` property for images usaged in the `ProductGridBlock`.

More info can be found in the [ticket](https://www.notion.so/hedviginsurance/ImageBlock-enable-images-load-performance-with-sizes-config-a9b5a31946664cd28194b55bfe4890c2?pvs=4) but with this we can improve page load by only requesting big images when we actually need them. Take `ProductGridBlock`'s images for example:

**Before**

image

**After**

Screenshot 2023-08-22 at 11 51 01

**Disclaimer**

Observe tho that this is not a definitive solution as we'd still need to proper provide those `sizes` values in Storyblok and update some of our resources - like student gif - since there's nothing `next/image` can do about them.


guilhermespopolin:
* main
  * **PR #3009** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3009" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3009?utm_source=stack-comment).

apps/store/src/components/ProductCard/ProductCard.tsx
```diff
@@ -45,7 +40,7 @@ export const ProductCard = ({
   return (
     <Container y={1}>
       <ImageWrapper aspectRatio={aspectRatio}>
-        <Image {...imageProps} alt={alt} fill sizes="100vw" />
+        <Image {...imageProps} alt={alt} fill />
```
robinandeer, line 43: nit `fill={true}`

apps/store/src/blocks/ImageBlock.tsx
```diff
@@ -52,6 +53,7 @@ export const ImageBlock = ({ blok, nested }: ImageBlockProps) => {
           roundedCorners={!blok.fullBleed}
           alt={blok.image.alt}
           fill
+          sizes={blok.sizes}
```
robinandeer, line 56: What do we expect the editors to fill in here?

A string like this seems way too technical, no?

```
`
  (max-width: ${getMediaQueryBreakpoint('md')}px) 100vw,
  (max-width: ${getMediaQueryBreakpoint('lg')}px) 50vw,
  25vw
`
```

robinandeer (commented):
Good work, just thinking of the best API for the CMS editors

gustaveen (commented):
Nice job with this, could be some nice savings! 👌 
I wonder the same though about if this is to complex for editors to handle and it might be hard to keep layout changes in sync. Meaning if you would set `sizes` to work nicely and someone would change the layout to be wider/narrower it would need to be updated as well. 

I wonder if we could calculate it based on the layout but might be a bit tricky for some components. 🙃

apps/store/src/blocks/ImageBlock.tsx
guilhermespopolin: I understand. I've tried to be as clear as possible about how to use it but it may still be too technical. One would need to have some knowledge on [responsive images](https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Responsive_images#how_do_you_create_responsive_images). I also though on the possibility of us filling it up. The issue is that I don't see other way: in order to proper set those one would need to know beforehand how the images are suppose to scale on different viewports. For components that we control the layout entirely like `ProductGridBlock`, we can do it. However for generic components like `ImageBlock`, which dimensions rely entirely on the context they are inserted, only the one who built a page with it knows beforehand how it should scale across different viewports.

image

guilhermespopolin:
> Nice job with this, could be some nice savings! 👌 I wonder the same though about if this is to complex for editors to handle and it might be hard to keep layout changes in sync. Meaning if you would set `sizes` to work nicely and someone would change the layout to be wider/narrower it would need to be updated as well.
> 
> I wonder if we could calculate it based on the layout but might be a bit tricky for some components. 🙃

That's true. I was considering components that we control the layout like `ProductGridBlock` immune to those changes but the truth is that they aren't. Changes in Storyblok could still influence our decision on the value we set for `sizes`. I'm stating to think we can't implement these on a very flexible setup like CSM driven apps 🤔

guilhermespopolin:
Closing this as this changes might not be sufficient.

PR #3083, opened 2023-08-28
fix(PurchaseForm): make sure shows general error message only when necessary

guilhermespopolin:
## Describe your changes

- PurchaseForm: only displays general error message dialog when there was no API errors while confirming the price intent and we also don't get `updatedPriceIntent` object from the API.

## Justify why they are needed

While reverting [min/max constraints for pet insurance birthdate field](https://github.com/HedvigInsurance/racoon/pull/3081) I've noticed we were overriding price intent confirm API error dialogs with general error message dialog. We probalby want to display the later only when there was no API errors while confirming the price intent but for some reason we didn't get `updatedPriceIntent` object.

**Before**

https://github.com/HedvigInsurance/racoon/assets/19200662/9cc6c444-c167-4d10-9847-05b747d6c63d

**After**

https://github.com/HedvigInsurance/racoon/assets/19200662/758f39d3-1724-40ec-8567-ffb5b15ef6cd

PR #3084, opened 2023-08-28
GEN-973 | Fix displaying error messages in purchase form

robinandeer:
## Describe your changes

- Avoid displaying multiple error messages when confirming price intent

  - Separate error messages for user errors (UW filters) and system errors (e.g. network error)

- Display generic error dialog when an error occurs while updating price intent data (e.g. network error)

Co-author: @guilhermespopolin 

- We currently first display user error (UW filters) and then quickly switch to a generic error message

### DIFF
--- a/apps/store/src/components/ProductPage/PurchaseForm/PurchaseForm.tsx
+++ b/apps/store/src/components/ProductPage/PurchaseForm/PurchaseForm.tsx
@@ -321,24 +321,24 @@ const EditingState = (props: EditingStateProps) => {
   const { t } = useTranslation('purchase-form')
   const tracking = useTracking()
 
+  const priceLoaderPromise = useRef<Promise<void> | null>(null)
   const [confirmPriceIntent, result] = usePriceIntentConfirmMutation({
     variables: { priceIntentId: priceIntent.id },
     onError(error) {
-      datadogLogs.logger.error('Failed to confirm price intent', {
+      datadogLogs.logger.warn('Failed to confirm price intent', {
         error,
         priceIntentId: priceIntent.id,
       })
-      onComplete(error.message)
+      if (error.networkError || error.graphQLErrors.length > 0) {
+        // Unknown error
+        onComplete(t('GENERAL_ERROR_DIALOG_PROMPT'))
+      } else {
+        // User error
+        onComplete(error.message)
+      }
     },
-  })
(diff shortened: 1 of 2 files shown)


apps/store/src/components/ProductPage/PurchaseForm/PurchaseForm.tsx
```diff
@@ -322,13 +322,19 @@ const EditingState = (props: EditingStateProps) => {
   const tracking = useTracking()
 
   const [confirmPriceIntent, result] = usePriceIntentConfirmMutation({
-    variables: { priceIntentId: priceIntent.id },
```
alebedev, line 325: Debug code?

apps/store/src/components/ProductPage/PurchaseForm/PurchaseForm.tsx
```diff
@@ -322,13 +322,19 @@ const EditingState = (props: EditingStateProps) => {
...
+    variables: { priceIntentId: 'adwiojoaijwdoij' },
     onError(error) {
       datadogLogs.logger.error('Failed to confirm price intent', {
         error,
         priceIntentId: priceIntent.id,
       })
-      onComplete(error.message)
+      if (error.networkError || error.graphQLErrors.length > 0) {
```
alebedev, line 331: Would be easier to read as util call.  Or maybe we should move it to GraphQL filters layer?

alebedev (approved):
Nice!  Some ideas for future refactoring inside, but we probably can merge this as is

apps/store/src/components/ProductPage/PurchaseForm/PurchaseForm.tsx
robinandeer: Hmm yeah, it's pretty explicit now which is good

But would be tedious to do this in many places

What do you mean with "graphql filters layer"?

guilhermespopolin (approved):
I guess we both did the same thing 😬 
What do you think about using `onCompleted` to achieve that?

https://github.com/HedvigInsurance/racoon/pull/3083

robinandeer:
> I guess we both did the same thing 😬 
> What do you think about using `onCompleted` to achieve that?
> 
> https://github.com/HedvigInsurance/racoon/pull/3083

Hmmm isn't that what I have done? 🤔 Let's discuss tomorrow!

PR #3086, opened 2023-08-29
GEN-988 | CRM: redirect users with cart items to the cart

robinandeer:
## Describe your changes

- Redirect users with items added to the cart to the cart page

- Check if URL is valid and otherwise redirect to the store page

- Start of logic for where to redirect users in CRM retargeting

- There's a visual bug where we show a white page on the `/se/session/:id` route. I think we want to make this into an API route to make it behave like we want => to use browser loading UI.

### DIFF
--- a/apps/store/src/features/retargeting/useRedirectUser.ts
+++ b/apps/store/src/features/retargeting/useRedirectUser.ts
@@ -1,0 +1,40 @@
+import { datadogLogs } from '@datadog/browser-logs'
+import { useRouter } from 'next/router'
+import { useShopSessionQuery } from '@/services/apollo/generated'
+import { useCurrentLocale } from '@/utils/l10n/useCurrentLocale'
+import { PageLink } from '@/utils/PageLink'
+import { QueryParam } from './retargeting.constants'
+
+export const useRedirectUser = () => {
+  const { routingLocale } = useCurrentLocale()
+  const router = useRouter()
+
+  const shopSessionId = useShopSessionId()
+  useShopSessionQuery({
+    skip: !shopSessionId,
+    variables: shopSessionId ? { shopSessionId } : undefined,
+    onCompleted(data) {
+      if (data.shopSession.cart.entries.length > 0) {
+        router.push(
+          PageLink.session({
+            shopSessionId: data.shopSession.id,
+            next: PageLink.cart({ locale: routingLocale }),
+          }),
+        )
+      }
+    },
+    onError(error) {
+      datadogLogs.logger.warn('Retargeting | Failed to fetch shop session', {
+        shopSessionId,
+        error,
+      })
+      router.push(PageLink.store({ locale: routingLocale }))
+    },
+  })
+}
+
+const useShopSessionId = () => {
+  const router = useRouter()
+  const shopSessionId = router.query[QueryParam.ShopSession]
+  return typeof shopSessionId === 'string' ? shopSessionId : undefined
+}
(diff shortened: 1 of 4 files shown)


apps/store/src/features/retargeting/validateUrl.ts
```diff
@@ -0,0 +1,18 @@
...
+import { QueryParam } from './retargeting.constants'
+
+export const validateUrl = () => {
+  Router.ready(() => {
+    const searchParams = new URLSearchParams(window.location.search)
+    const shopSessionId = searchParams.get(QueryParam.ShopSession)
+    if (!shopSessionId) {
+      datadogLogs.logger.warn('Retargeting | Missing shop session in url')
+      const { routingLocale } = getLocaleOrFallback(Router.locale as UiLocale)
+      window.location.assign(PageLink.store({ locale: routingLocale }))
```
alebedev, line 15: Can we do it at caller side?  I wouldn't expect `validateX` function to have side effects

Or we can rename it if side effect here is the best way to organize code

apps/store/src/features/retargeting/useRedirectUser.ts
```diff
@@ -0,0 +1,40 @@
...
+  const { routingLocale } = useCurrentLocale()
+  const router = useRouter()
+
+  const shopSessionId = useShopSessionId()
+  useShopSessionQuery({
+    skip: !shopSessionId,
+    variables: shopSessionId ? { shopSessionId } : undefined,
+    onCompleted(data) {
+      if (data.shopSession.cart.entries.length > 0) {
+        router.push(
```
alebedev, line 18: Would a log message here help with debugging?

alebedev (approved):
Looks good, minor questions

apps/store/src/features/retargeting/useRedirectUser.ts
robinandeer: Yeah that's true -- you mean through DD, right?

Perhaps we can cover the other cases first and then add logging on top -- I will add a ticket for it 🙂

apps/store/src/features/retargeting/validateUrl.ts
robinandeer: Ahhhh yeah I was thinking of naming it something else - would be an improvement

The caller side just derives state from the URL so there's no natural place to do something imperative like navigating to another place. However, this logic is kind of duplicated so perhaps I can refactor it a bit.

guilhermespopolin: Or maybe it could be placed inside `useRedirectUser`. I mean, at the end the reason we're validating the URL is to redirect the user to store page in case it's invalid so it can maybe fall under the same bucket.

robinandeer: You mean in module scope then or call it in a use effect?

guilhermespopolin: I guess both could work but I was thinking more inside `useRedirectUser` so a `useEffect`. I wonder if there's other way around `useEffect` tho 🤔

robinandeer: I feel like there would be risk we would make multiple redirects in that case...

robinandeer:
### Merge Activity

* **Aug 29, 3:54 PM**: @robinandeer started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3086).

PR #3108, opened 2023-08-31
Retargeting: add information about feature in README

robinandeer:
## Describe your changes

- Add information about retargeting feature to the README.md file

- Provide brief feature documentation

- Should we co-locate this information to the feature/code as well? 🤔

### DIFF
--- a/apps/store/README.md
+++ b/apps/store/README.md
@@ -142,3 +142,17 @@ We use a few custom headers when communicating with the API. These are:
 ## Page Debug Menu
 
 To show the page debug menu, press `Ctrl + d`. The menu will show the current shop session ID and other helpful information based on the page your are viewing.
+
+## Feature: Retargeting
+
+Entry point: `/{LOCALE}/session/resume?shopSessionId={SHOP_SESSION_ID}`
+
+This feature is used to re-target users who have started a shop session but not completed it. Users are sent here via CRM (email). The page shows a loading state while determining the next step. The next step is determined as follows:
+
+- If the user has a non-empty cart, they are redirected to the cart page.
+- If the user has confirmed a single price intent, they are redirected to the relevant product page.
+- If the user has confirmed multiple price intents, one offer per price intent is added, and the user is redirected to the cart page.
+
+Additional query parameters:
+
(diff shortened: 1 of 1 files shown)


robinandeer:
* main
  * **PR #3108** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3108" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3108?utm_source=stack-comment).

apps/store/README.md
```diff
@@ -142,3 +142,17 @@ We use a few custom headers when communicating with the API. These are:
...
+
+## Feature: Retargeting
+
+Entry point: `/{LOCALE}/session/resume?shopSessionId={SHOP_SESSION_ID}`
+
+This feature is used to re-target users who have started a shop session but not completed it. Users are sent here via CRM (email). The page shows a loading state while determining the next step. The next step is determined as follows:
+
+- If the user has a non-empty cart, they are redirected to the cart page.
+- If the user has confirmed a single price intent, they are redirected to the relevant product page.
+- If the user has confirmed multiple price intents, one offer per price intent is added, and the user is redirected to the cart page.
```
guilhermespopolin, line 154: I wonder if it would be relevant also add the reasoning behind the selection of one _offer per price intent_ 🤔 
But I guess its better not included. I can see that changing and we forgetting to update the document accordingly.

guilhermespopolin (approved):
After a quick look, it seems we'd need some tooling to condense multiple md files into a single output. Maybe the effort to co-locate that info under its feature folder is not worth it.

apps/store/README.md
robinandeer: Ahh yes, I think we haven't really worked out that logic so perhaps better be a bit vague for the time being

robinandeer:
> After a quick look, it seems we'd need some tooling to condense multiple md files into a single output. Maybe the effort to co-locate that info under its feature folder is not worth it.

Right, yeah I wouldn't want to invest in that

I was more thinking that we could have a summary README in the root that would point to more feature-specific README files elsewhere in the project

PR #3129, opened 2023-09-04
GEN-1065 - refact(PageLink.cart): return an URL object instead of a string

guilhermespopolin:
## Describe your changes

- Changes `PageLink.cart` return type: `string` --> `URL`

### DIFF
--- a/apps/store/src/pages/api/session/create.ts
+++ b/apps/store/src/pages/api/session/create.ts
@@ -18,7 +18,7 @@ import {
 } from '@/services/priceIntent/PriceIntentService'
 import { setupShopSessionServiceServerSide } from '@/services/shopSession/ShopSession.helpers'
 import { RoutingLocale } from '@/utils/l10n/types'
-import { ORIGIN_URL, PageLink } from '@/utils/PageLink'
+import { PageLink } from '@/utils/PageLink'
 
 const TEST_SSN = '199808302393'
 const productNames = ['SE_APARTMENT_RENT', 'SE_ACCIDENT'] as const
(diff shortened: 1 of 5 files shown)


guilhermespopolin:
* main
  * **PR #3119** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3119" target="_blank">Graphite</a> 
    * **PR #3120** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3120" target="_blank">Graphite</a> 
      * **PR #3127** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3127" target="_blank">Graphite</a> 
        * **PR #3129** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3129" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3129?utm_source=stack-comment).

apps/store/src/pages/api/session/create.ts
```diff
@@ -66,7 +66,7 @@ const handler = async (req: NextApiRequest, res: NextApiResponse) => {
   }
 
   const nextURL = new URL(ORIGIN_URL)
-  nextURL.pathname = PageLink.cart({ locale: DEFAULT_LOCALE })
+  nextURL.pathname = PageLink.cart({ locale: DEFAULT_LOCALE }).pathname
   const destination = nextURL.toString()
```
robinandeer, line 70: this is awkward --- seems like we can just do:

```suggestion
  const destination = PageLink.cart({ locale: DEFAULT_LOCALE }).href
```

robinandeer (approved):
Nice work!

apps/store/src/pages/api/session/create.ts
guilhermespopolin: True. Changing it.

PR #3130, opened 2023-09-04
GEN-1065 - refact(PageLink.checkout): return an URL object instead of a string

guilhermespopolin:
## Describe your changes

- Changes `PageLink.checkout` return type: `string` --> `URL`


guilhermespopolin:
* main
  * **PR #3119** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3119" target="_blank">Graphite</a> 
    * **PR #3120** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3120" target="_blank">Graphite</a> 
      * **PR #3127** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3127" target="_blank">Graphite</a> 
        * **PR #3129** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3129" target="_blank">Graphite</a> 
          * **PR #3130** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3130" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3130?utm_source=stack-comment).

apps/store/src/components/ProductPage/PurchaseForm/OfferPresenter.tsx
```diff
@@ -90,7 +90,9 @@ export const OfferPresenter = (props: Props) => {
       tracking.reportAddToCart(addedProductOffer, 'store')
       if (addToCartRedirect === AddToCartRedirect.Checkout) {
         tracking.reportBeginCheckout(shopSession.cart)
-        nextUrl = PageLink.checkout({ expandCart: true })
+        const checkoutUrl = PageLink.checkout({ expandCart: true })
+        const checkoutRelativeUrl = checkoutUrl.pathname + checkoutUrl.search
+        nextUrl = checkoutRelativeUrl
```
guilhermespopolin, line 95: This is an example of how it would look if we'd like to use relative paths. `nextUrl = PageLink.checkout({ expandCart: true })` would also works.

apps/store/src/components/CheckoutHeader/CheckoutHeader.helpers.ts
```diff
@@ -45,7 +45,7 @@ type GetCheckoutStepLinkParams = {
 export const getCheckoutStepLink = ({ step, shopSessionId, locale }: GetCheckoutStepLinkParams) => {
   switch (step) {
     case CheckoutStep.Checkout:
-      return PageLink.checkout()
+      return PageLink.checkout().toString()
```
robinandeer, line 48: should we try to be consistent and just use `.href`?

apps/store/src/components/ProductPage/PurchaseForm/OfferPresenter.tsx
robinandeer: I wonder if we should extend the built-in type to add method for this

See here for how to:
https://stackoverflow.com/a/58266306

Would be for a different PR tho.

robinandeer (approved):
Good stuff, let's just test this carefully so we don't mess up anything related to automatic locale handling!

apps/store/src/components/ProductPage/PurchaseForm/OfferPresenter.tsx
guilhermespopolin: That can work as well

PR #3145, opened 2023-09-04
GEN-1065 - refact(PageLink.paymentConnectLegacySucess): return an URL object instead of a string

guilhermespopolin:
## Describe your changes

- Changes `PageLink.paymentConnectLegacySucess` return type: `string` --> `URL`

### DIFF
--- a/apps/store/src/pages/api/adyen-callback/[locale].ts
+++ b/apps/store/src/pages/api/adyen-callback/[locale].ts
@@ -20,7 +20,7 @@ const handler = async (req: NextApiRequest, res: NextApiResponse) => {
     return
   }
 
-  res.redirect(302, PageLink.paymentConnectLegacySuccess({ locale: req.query.locale }))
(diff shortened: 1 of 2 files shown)


apps/store/src/pages/api/adyen-callback/[locale].ts
```diff
@@ -20,7 +20,7 @@ const handler = async (req: NextApiRequest, res: NextApiResponse) => {
     return
   }
 
-  res.redirect(302, PageLink.paymentConnectLegacySuccess({ locale: req.query.locale }))
+  res.redirect(302, PageLink.paymentConnectLegacySuccess({ locale: req.query.locale }).pathname)
```
Comment (robinandeer) [inline apps/store/src/pages/api/adyen-callback/[locale].ts:23]: do we need to return pathname here? Or should we change to `.href`?

robinandeer (approved):

apps/store/src/pages/api/adyen-callback/[locale].ts
Comment (guilhermespopolin) [reply apps/store/src/pages/api/adyen-callback/[locale].ts:23]: `href` should be fine.

PR #3187, opened 2023-09-12
GEN-1103 | Fix: correctly redirect 'default' locale

robinandeer:
## Describe your changes

- Redirect "default" locale by cookie preference OR use fallback (se)

- Apply to ALL routes (not just /)

- Remove geo redirect

- Let unknown locales render 404

- We noticed that we don't redirect `/cart` to `/se/cart`. This is better UX compared to just rendering 404.

- We don't need to support other markets so I removed the geo redirect

- I couldn't reproduce the issue where Next.js doesn't filter out API routes etc. so trying it out for all routes instead of just /.

### DIFF
--- a/apps/store/src/middleware.ts
+++ b/apps/store/src/middleware.ts
@@ -4,7 +4,7 @@ import type { NextRequest } from 'next/server'
 import { experimentMiddleware } from '@/services/Tracking/experimentMiddleware'
 import { countries } from '@/utils/l10n/countries'
 import { LOCALE_COOKIE_KEY } from '@/utils/l10n/locales'
-import { isRoutingLocale, toRoutingLocale } from '@/utils/l10n/localeUtils'
+import { toRoutingLocale } from '@/utils/l10n/localeUtils'
 
 export const config = {
   matcher: [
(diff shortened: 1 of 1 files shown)


robinandeer:
* main
  * **PR #3187** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3187" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3187?utm_source=stack-comment).

guilhermespopolin (approved):

apps/store/src/middleware.ts
```diff
@@ -15,12 +15,8 @@ export const config = {
 }
 
 export async function middleware(req: NextRequest) {
-  if (!isRoutingLocale(req.nextUrl.locale)) {
-    // Workaround for Vercel edge middleware matching requests for static resources despite config specifying not to do so
```
robinandeer, line 19: I've tested the site and can't reproduce it -- perhaps it's fixed on vercel / nextjs

PR #3194, opened 2023-09-13
Fix: locale redirect matching stuff it shouldn't

robinandeer:
## Describe your changes

- Use browser API to navigate on to API route

- Update middleware matching to exclude paths that contain a dot/period (.)

- Next.js router navigation triggers the middleware for some unknown reason

- We have an issue where we add locale to static files in the public folder. Excluding paths with a dot/period (.) will prevent this.


alebedev (approved):
Great debugging!

guilhermespopolin (approved):

PR #3211, opened 2023-09-15
GEN-1138 | BankSignering: fix logging opening link

robinandeer:
## Describe your changes

- Explicit logging of bank signering open action

- ~I can't find the action in DD, so I'm guessing the "data" attribute doesn't work as expected~

- Datadog doesn't automatically track clicks that lead to a new page being loaded (https://docs.datadoghq.com/real_user_monitoring/browser/tracking_user_actions/?tab=npm#track-user-interactions)

### DIFF
--- a/apps/store/src/components/ConfirmationPage/ContractCard.tsx
+++ b/apps/store/src/components/ConfirmationPage/ContractCard.tsx
@@ -46,11 +46,15 @@ const PendingContractCard = (props: ContractCardProps) => {
     },
   })
 
-  const handleClick = (contractId: string) => () => {
+  const handleClickInitiate = (contractId: string) => () => {
     datadogRum.addAction(BankSigneringEvent.Initiated, { contractId })
     // Logger context used in mutation result handlers
     BANK_SIGNERING_LOGGER.addContext('contractId', contractId)
     initiateBankSignering({ variables: { contractId } })
+  }
+
+  const handleClickOpen = (contractId: string) => () => {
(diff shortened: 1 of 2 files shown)


robinandeer:
* main
  * **PR #3211** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3211" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3211?utm_source=stack-comment).

alebedev (commented):
Can we reproduce the problem elsewhere?  It's very nice when attribute-based approach works, so let's try to see where we can rely on it

Maybe it has something to do with `target="_blank"` and the link being external?

robinandeer:
> Can we reproduce the problem elsewhere?  It's very nice when attribute-based approach works, so let's try to see where we can rely on it
> 
> Maybe it has something to do with `target="_blank"` and the link being external?

Very good point! It seems my conclusion was wrong -- data-attributes work as expected. However, in this case we should not expect it to track anything since the click "lead to a new page being loaded" (https://docs.datadoghq.com/real_user_monitoring/browser/tracking_user_actions/?tab=npm#track-user-interactions)

So in this case I think it's correct to use a custom action 🙂

alebedev (approved):
Great, now we know when we need to do custom actions

guilhermespopolin (approved):
Nice. Good to know about this detail with dd actions.

PR #3233, opened 2023-09-19
GEN-1081 - fix(c1): make it work with lazy load strategy

guilhermespopolin:
## Describe your changes

Make usage of `createWidget` API to append _customer1_ chat iframe into the DOM. With it we can better control when to load it, in contrast with the current setup where its loading process was being done on an _onload_ event handler they attach when _c1_ script get's parsed.

Screenshot 2023-09-19 at 09 46 48

**Disclaimers**

1. With this change we're appending the chat iframe into the DOM after _c1_ gets loaded with `next/Script` `onLoad` prop. That's because right now we just want fix the issue that originated this ticket where chat iframe might not get loaded with lazy loading approach. However we have different plans for the future where we'd condition chat iframe addition into the DOM as a result of a button click. (check [here](https://www.notion.so/hedviginsurance/Customer1-create-launch-button-desktop-114e9f7239dc4b38a45800006c6a822c?pvs=4)).
2. I could verify that even with the addition of `createWidget` on their API, their script still appends chat iframe into the DOM in a `onload` event handler (check image below). That's good because without it chat wouldn't be available in prod today, at least not before getting this merged. My idea however is to get this merged and then ask them to remove that.

Screenshot 2023-09-19 at 10 25 31

So _c1_ works properly when lazy loading its script.

### DIFF
--- a/apps/store/index.d.ts
+++ b/apps/store/index.d.ts
@@ -47,6 +47,7 @@ declare global {
 
     // Customer First (C1)
     customerFirstAPI?: {
+      createWidget(): void
       openWidget(): void
       closeWidget(): void
     }
--- a/apps/store/src/services/CustomerFirst.tsx
+++ b/apps/store/src/services/CustomerFirst.tsx
@@ -22,7 +22,11 @@ export const CustomerFirstScript = ({ hideChat = false }: Props) => {
   return (
     <>
       <Global styles={{ '#chat-iframe': { display: showLauncher ? 'initial' : 'none' } }} />
-      <Script strategy="afterInteractive" src={chatWidgetSrc} />
+      <Script
+        strategy="afterInteractive"
+        src={chatWidgetSrc}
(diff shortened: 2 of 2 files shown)


guilhermespopolin:
* main
  * **PR #3233** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3233" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3233?utm_source=stack-comment).

robinandeer (approved):
Looks good

guilhermespopolin:
### Merge Activity

* **Sep 19, 6:00 AM**: @guilhermespopolin started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3233).

PR #3242, opened 2023-09-20
GEN-975 - fix(TextBlock): proper render rich text content

guilhermespopolin:
## Describe your changes

- Fix text formatting for `TextBlock`'s rich text field - `body`

Screenshot 2023-09-20 at 10 07 13

`TextBlock` was accepting a rich text field but none formatting was being applied. Now the same one we use for `RichTextBlock` is being applied for `TextBlock` as well.

### DIFF
--- a/apps/store/src/blocks/RichTextBlock/RichTextBlock.tsx
+++ b/apps/store/src/blocks/RichTextBlock/RichTextBlock.tsx
@@ -1,5 +1,6 @@
 import { ISbRichtext, storyblokEditable } from '@storyblok/react'
 import Link from 'next/link'
+import { useMemo, type ReactNode } from 'react'
 import { render, RenderOptions, MARK_LINK } from 'storyblok-rich-text-react-renderer'
 import { GridLayout } from '@/components/GridLayout/GridLayout'
 import { RichText } from '@/components/RichText/RichText'
@@ -12,7 +13,23 @@ export type RichTextBlockProps = SbBaseBlockProps<{
   largeText?: boolean
 }>
 
-const richTextRenderOptions: RenderOptions = {
+export const RichTextBlock = ({ blok }: RichTextBlockProps) => {
+  const content = useMemo(() => renderRichText(blok.content), [blok.content])
+
+  return (
+    <GridLayout.Root {...storyblokEditable(blok)}>
+      <GridLayout.Content
+        width={blok.layout?.widths ?? { base: '1', md: '2/3', xl: '1/2' }}
+        align={blok.layout?.alignment ?? 'center'}
+      >
+        <RichText largeText={blok.largeText}>{content}</RichText>
+      </GridLayout.Content>
+    </GridLayout.Root>
+  )
+}
+RichTextBlock.blockName = 'richText'
+
+const RENDER_OPTIONS: RenderOptions = {
   blokResolvers: {
     image: (props) => <ImageBlock blok={props as ImageBlockProps['blok']} nested={true} />,
   },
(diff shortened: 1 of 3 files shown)


guilhermespopolin:
* main
  * **PR #3242** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3242" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3242?utm_source=stack-comment).

apps/store/src/blocks/TextBlock.tsx
```diff
@@ -17,25 +18,46 @@ export type TextBlockProps = SbBaseBlockProps<{
...
+    color: 'inherit',
+    fontSize: 'inherit',
+  },
+  'h2, h3, h4': {
+    color: 'inherit',
+    fontSize: 'inherit',
+  },
+  'ul li': {
+    '&::before': {
+      backgroundColor: theme.colors[color],
```
robinandeer, line 53: nit: should we use `getColor` here? Not sure if it makes a big difference

apps/store/src/blocks/RichTextBlock/RichTextBlock.tsx
```diff
@@ -55,18 +72,7 @@ const isExternalLink = (href?: string) => href?.match(/^(https?:)?\/\//)
...
-      >
-        <RichText largeText={blok.largeText}>
-          {render(blok.content, richTextRenderOptions)}
-        </RichText>
-      </GridLayout.Content>
-    </GridLayout.Root>
-  )
-}
-RichTextBlock.blockName = 'richText'
+export const renderRichText = (
```
robinandeer, line 75: nit: I wonder if we should place this inside `RichText` component module instead?

That would make it clearer that it's a utility used by multiple blocks 🙂

robinandeer (approved):
Great work, will make editors happy!

Added via Giphy

apps/store/src/blocks/RichTextBlock/RichTextBlock.tsx
guilhermespopolin: Thought the same. What stopped me was that this `renderRichText` function is specific for storyblok related content - it relies on `storyblok-rich-text-react-renderer` dependency and the content's type it can handle is `ISbRichText` (`@storyblok/react`). So I thought about don't "spilling" storyblok related stuff into "regular" components like `RichText` is suppose to be. Do you think that makes sense?

apps/store/src/blocks/TextBlock.tsx
guilhermespopolin: I like it. Changing it :)

apps/store/src/blocks/RichTextBlock/RichTextBlock.tsx
robinandeer: right that makes sense, let's see if we use it in a third place -- then we can make it a utility

guilhermespopolin:
### Merge Activity

* **Sep 20, 9:58 AM**: @guilhermespopolin started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3242).

apps/store/src/blocks/TextBlock.tsx
```diff
@@ -17,25 +18,46 @@ export type TextBlockProps = SbBaseBlockProps<{
...
+    color: 'inherit',
+    fontSize: 'inherit',
+  },
+  'h2, h3, h4': {
+    color: 'inherit',
+    fontSize: 'inherit',
+  },
+  'ul li': {
+    '&::before': {
+      backgroundColor: getColor(color),
```
gustaveen, line 53: Do we need to pass down a color prop to this styled component for this case? I suppose you tried `inherit` but otherwise `currentColor` should work right?

gustaveen (commented):
Good job! 🙌 Just want us to be a bit cautious if we need too much overriding off the RichText styles in this case. If we only need parts of the styling for generalising text styling, maybe we can break it out in smaller chunks in `RichText.stlyes.tsx` and only import that and then add Block specific styling.

apps/store/src/blocks/TextBlock.tsx
guilhermespopolin: Yeah, I tried `inherit` but it didn't work. Haven't tried `currentColor` tho. If that works it will be awesome. Gonna try that out.

guilhermespopolin:
> Good job! 🙌 Just want us to be a bit cautious if we need too much overriding off the RichText styles in this case. If we only need parts of the styling for generalising text styling, maybe we can break it out in smaller chunks in `RichText.stlyes.tsx` and only import that and then add Block specific styling.

I hear ya. This is even more important now that Peter pointed out that those changes messed up with the layout a little bit and we'd need to override even more cases: https://github.com/HedvigInsurance/racoon/pull/3250
Gonna try to extract pagagraph styles into some style chunk as you've suggested.

PR #3250, opened 2023-09-20
fix(TextBlock): fix paragraphs margins

guilhermespopolin:
## Describe your changes

- Fix margin for `p` elements added by rich text rendering.

### DIFF
--- a/apps/store/src/components/RichText/RichText.styles.tsx
+++ b/apps/store/src/components/RichText/RichText.styles.tsx
@@ -59,6 +59,10 @@ export const listStyles = css({
     marginBlock: theme.space.xs,
   },
 
+  'li > p': {
+    marginBlock: 0,
+  },
+
   'ul li': {
     position: 'relative',
     paddingLeft: theme.space.lg,
@@ -72,7 +76,7 @@ export const listStyles = css({
       width: 15,
       height: 15,
       borderRadius: '100%',
-      backgroundColor: theme.colors.textSecondary,
+      backgroundColor: 'var(--color, currentColor)',
     },
 
     ul: {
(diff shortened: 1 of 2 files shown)


robinandeer (commented):
Hmm... instead of overriding, can we apply the margin styles only to the places that need them? 🤔

guilhermespopolin:
> Hmm... instead of overriding, can we apply the margin styles only to the places that need them? 🤔

The issue is that rich text styles are applied in a broad way: https://github.com/HedvigInsurance/racoon/blob/main/apps/store/src/components/RichText/RichText.styles.tsx#L131
I don't se how we can be more specific on the selector there based on the way it works 🤔

apps/store/src/components/RichText/RichText.styles.tsx
```diff
@@ -72,7 +72,7 @@ export const listStyles = css({
       width: 15,
       height: 15,
       borderRadius: '100%',
-      backgroundColor: theme.colors.textSecondary,
+      backgroundColor: 'var(--color, currentColor)',
```
guilhermespopolin, line 75: Thanks for the tip with `currentColor` @gustaveen 🙌🏻

apps/store/src/components/RichText/RichText.styles.tsx
```diff
@@ -116,43 +116,53 @@ export const listStyles = css({
...
+  lineHeight: 1.32,
+})
 
-    '.preamble': {
-      display: 'block',
-      marginBottom: theme.space.xl,
-      fontSize: theme.fontSizes.xl,
-      color: theme.colors.textPrimary,
-    },
+export const paragraphyStyles = css({
```
robinandeer, line 124: nit: spelling mistake -> `paragraphStyles `

robinandeer (approved):
This looks reasonable :D

apps/store/src/components/RichText/RichText.styles.tsx
guilhermespopolin: Thanks

PR #3263, opened 2023-09-25
GEN-1189 - refact(CarTrialExtensionBlock): query real data from shopSession and priceItent

guilhermespopolin:
## Describe your changes

* Adds a `CarTrialExtension` query: at the moment it only fetches info about `shopSession` and `priceIntent`. TrialExtension information is still mocked;
* Update fixtures so it satisfied data request data format;

Get closer to the real feature while we don't have the API configured.

### DIFF
--- a/apps/store/src/features/carDealership/CarTrialExtension.graphql
+++ b/apps/store/src/features/carDealership/CarTrialExtension.graphql
@@ -1,0 +1,77 @@
+query CarTrialExtension($shopSessionId: UUID!) {
+  shopSession(id: $shopSessionId) {
+    id
+    customer {
+      ssn
+      authenticationStatus
+    }
+    cart {
+      id
+      entries {
+        id
+      }
+    }
+    priceIntents {
+      id
+      data
+      offers {
+        id
+        startDate
+        cost {
+          net {
+            amount
+            currencyCode
+          }
+          gross {
+            amount
+            currencyCode
+          }
+          discount {
+            amount
+            currencyCode
+          }
+        }
+        displayItems {
+          key
+          value
+          displayTitle
+          displayValue
+        }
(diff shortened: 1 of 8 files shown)


guilhermespopolin:
* main
  * **PR #3263** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3263" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3263?utm_source=stack-comment).

apps/store/src/features/carDealership/CarTrialExtension.graphql
```diff
@@ -0,0 +1,77 @@
...
+    cart {
+      id
+      entries {
+        id
+      }
+    }
+    priceIntents {
+      id
+      data
+      offers {
```
guilhermespopolin, line 17: I'd like to get your opinions on this. We could request `ProductOffer` fragment here but we'd be requesting some fields that we don't use here like `priceMatch` and `cancellation` so I though about just requesting what we actually gonna use, while having to type more as the tradeoff. Any opnions?

robinandeer: Yeah I like to separate as much as we can from the rest of the app -- the feature should be sort of self contained so I like this approach

robinandeer (approved):
Great stuff, let's try it out!

guilhermespopolin:
### Merge Activity

* **Sep 25, 8:07 AM**: @guilhermespopolin started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3263).

PR #3273, opened 2023-09-25
GEN-1200  - fix(useSignAndPay): show error dialog if something goes wrong while signing

guilhermespopolin:
## Describe your changes

- `useSignAndPay` - Show error dialog if somethings goes wrong while signing

Better UX

### DIFF
--- a/apps/store/src/features/carDealership/useSignAndPay.ts
+++ b/apps/store/src/features/carDealership/useSignAndPay.ts
@@ -1,5 +1,6 @@
 import { datadogLogs } from '@datadog/browser-logs'
 import { useRouter } from 'next/router'
+import { useEffect } from 'react'
 import {
   useCarDealershipRemoveAddMutation,
   useCurrentMemberLazyQuery,
@@ -17,11 +18,17 @@ type Params = {
 }
 
 export const useSignAndPay = (params: Params) => {
-  const { startCheckoutSign } = useBankIdContext()
+  const { startCheckoutSign, currentOperation } = useBankIdContext()
   const router = useRouter()
   const { showError } = useAppErrorHandleContext()
-
   const [getCurrentMember] = useCurrentMemberLazyQuery()
+
+  useEffect(() => {
+    const error = currentOperation?.error
+    if (error instanceof Error) {
+      showError(error)
+    }
+  }, [showError, currentOperation])
 
   const performSign = () => {
     const { ssn, authenticationStatus } = params.shopSession.customer ?? {}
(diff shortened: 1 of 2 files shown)


apps/store/src/features/carDealership/useSignAndPay.ts
```diff
@@ -46,16 +49,23 @@ export const useSignAndPay = (params: Params) => {
...
+      startCheckoutSign({
+        shopSessionId: params.shopSession.id,
+        customerAuthenticationStatus: authenticationStatus,
+        ssn,
+        onSuccess() {
+          console.info('Successfully signed shop session')
+          getCurrentMember()
+        },
+      })
+    } catch (error) {
```
robinandeer, line 62: hmm what are we catching here? we are not awaiting "start checkout sign" so not sure we can catch the error

apps/store/src/features/carDealership/useSignAndPay.ts
```diff
@@ -46,16 +49,23 @@ export const useSignAndPay = (params: Params) => {
...
+      let message: string | null = null
+      if (error instanceof Error && isApolloError(error)) {
+        message = error.extraInfo?.userError?.message
+      }
+      showError(new Error(message ?? t('UNKNOWN_ERROR_MESSAGE')))
```
robinandeer, line 67: I learned that this function is pretty smart so you can just pass the whole error like:

```
showError(error)
```

and it will pick out the user error message or use a fallback!

robinandeer (commented):
Just provided some comments

robinandeer (commented):
I figured one way to accomplish this would be this:

```
useEffect(() => {
    const error = currentOperation?.error
    if (error instanceof Error) {
      showError(error)
    }
  }, [showError, currentOperation?.error])
```

Not super clean but gets the work done hehe

apps/store/src/features/carDealership/useSignAndPay.ts
guilhermespopolin: Nice. Have update id.

apps/store/src/services/bankId/useBankIdCheckoutSign.ts
```diff
@@ -76,7 +76,7 @@ type SignOptions = {
 export const useBankIdCheckoutSignApi = ({ dispatch }: Options) => {
   const [fetchSigning, signingResult] = useShopSessionSigningLazyQuery({})
 
-  const [starSignMutate] = useShopSessionStartSignMutation()
+  const [startSignMutate] = useShopSessionStartSignMutation()
```
guilhermespopolin, line 79: Not related: just found this typo while going through sign with BankId logic.

robinandeer (approved):
Very neat

guilhermespopolin:
### Merge Activity

* **Sep 28, 5:12 AM**: @guilhermespopolin started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3273).

PR #3275, opened 2023-09-26
GEN-1201 | Avoid closing BankID dialog before navigating

robinandeer:
## Describe your changes

- Convert onSuccess to async function to prevent dialog from closing before navigation

- We don't want page to show before navigation is complete


guilhermespopolin (approved):
Nice catch 🕵🏻

PR #3282, opened 2023-09-28
GEN-1216 | Display attention card after removing extension offer

robinandeer:
## Describe your changes

- Display attention card with message after removing car trial offer extension

- Implement component to replace some placeholder inside a text string

- Updates to offer page design

- The new design actually uses a "ToastBar", however we haven't implemented that yet, so I've just used an attention card for now


robinandeer:
* main
  * **PR #3282** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3282" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3282?utm_source=stack-comment).

apps/store/src/features/carDealership/TrialExtensionForm.tsx
```diff
@@ -111,6 +122,23 @@ export const TrialExtensionForm = (props: Props) => {
   )
 }
 
+type ReplaceTextProps = TextProps & {
+  text: string
+  children: string
```
alebedev, line 127: Nit: Only accepting string children vs another string prop - what is more obvious for the caller?  No strong opinion from me

apps/store/src/features/carDealership/TrialExtensionForm.tsx
```diff
@@ -111,6 +122,23 @@ export const TrialExtensionForm = (props: Props) => {
...
+}
+
+const ReplaceText = ({ text, children, ...props }: ReplaceTextProps) => {
+  const placeholder = '{}'
+  const [before, after] = children.split(placeholder, 2)
+  return (
+    <>
+      {before}
+      <Text {...props}>{text}</Text>
```
alebedev, line 136: Do we want replacement text to have different style?  Not clear why we wrap it separately

alebedev (approved):
Looks good, few questions

apps/store/src/features/carDealership/TrialExtensionForm.tsx
robinandeer: ahaaa, right yeah that is a good point -- might be overloading children-prop a little too much

This is kind of a hack -- let's see if it survives when we refactor the code

apps/store/src/features/carDealership/TrialExtensionForm.tsx
robinandeer: So what I want to achieve is some substring to be styled differently from the rest of the text.

The rest of the text should pick up styles from a different parent component.

robinandeer:
### Merge Activity

* **Sep 28, 2:31 PM**: @robinandeer started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3282).

PR #3316, opened 2023-10-03
GEN-1245 | Show a warning dialog when user proceeds without extension

robinandeer:
## Describe your changes

- Display dialog with warning message when user continues without extension offer

- Remove unused translations and components (no longer part of the design)

- Adpat to the new design

### DIFF
--- a/apps/store/src/features/carDealership/TrialExtensionForm.tsx
+++ b/apps/store/src/features/carDealership/TrialExtensionForm.tsx
@@ -1,22 +1,21 @@
 import { datadogRum } from '@datadog/browser-rum'
-import styled from '@emotion/styled'
 import { useTranslation } from 'next-i18next'
 import { useRouter } from 'next/router'
 import { useState, useMemo } from 'react'
-import { TextProps } from 'ui/src/components/Text/Text'
-import { Space, Button, RestartIcon, Text, BankIdIcon } from 'ui'
+import { Space, Button, Text, BankIdIcon } from 'ui'
 import { ProductItemContainer } from '@/components/ProductItem/ProductItemContainer'
 import { TotalAmount } from '@/components/ShopBreakdown/TotalAmount'
 import { SpaceFlex } from '@/components/SpaceFlex/SpaceFlex'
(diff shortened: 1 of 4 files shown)


robinandeer:
* main
  * **PR #3316** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3316" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3316?utm_source=stack-comment).

apps/store/src/features/carDealership/TrialExtensionForm.tsx
```diff
@@ -96,34 +92,18 @@ export const TrialExtensionForm = (props: Props) => {
...
+          <ConfirmPayWithoutExtensionDialog onConfirm={handleConfirmPay}>
+            <ConfirmTrigger>
+              <Button variant="primary">
+                <SpaceFlex space={0.5} align="center">
+                  <BankIdIcon />
+                  {t('CONNECT_PAYMENT_BUTTON')}
+                </SpaceFlex>
+              </Button>
+            </ConfirmTrigger>
+          </ConfirmPayWithoutExtensionDialog>
```
guilhermespopolin, line 106: Nit:
I think it would be easier to follow if `ConfirmPayWithoutExtensionDialog` exported a button component that accept a `onClick` and inside of it we render the button alongside the dialog. Right now it's flexible as it accepts a `children` but if the `children` doesn't get constructed with `ConfirmTrigger` it will not work, right?! Besides there's no prop/info defined here that will change how the button is going to work so we can easily construct it inside `ConfirmPayWithoutExtensionDialog` without having to pass any additional prop. What do you think?

guilhermespopolin (approved):
Nice. Just left a comment.

apps/store/src/features/carDealership/TrialExtensionForm.tsx
robinandeer: Very good point, let's co-locate as much as possible!

PR #3321, opened 2023-10-03
GEN-1236 / Add toggle for car dealership extension

gustaveen:
## Describe your changes
- Add ToggleCard to toggle extension offer 
- Add translations

https://github.com/HedvigInsurance/racoon/assets/6661511/baf063fb-393c-45db-89fa-14d1eb604d19

Next step: add icon and adjust styling

Extension offer won't be displayed per default

### DIFF
--- a/apps/store/src/features/carDealership/TrialExtensionForm.tsx
+++ b/apps/store/src/features/carDealership/TrialExtensionForm.tsx
@@ -8,7 +8,8 @@ import { TotalAmount } from '@/components/ShopBreakdown/TotalAmount'
 import { SpaceFlex } from '@/components/SpaceFlex/SpaceFlex'
 import { TextWithLink } from '@/components/TextWithLink'
 import { WithLink } from '@/components/TextWithLink'
-import { InfoToastBar } from '@/components/ToastBar/ToastBar'
+import { AttentionToastBar, InfoToastBar } from '@/components/ToastBar/ToastBar'
+import { ToggleCard } from '@/components/ToggleCard/ToggleCard'
 import { useBankIdContext } from '@/services/bankId/BankIdContext'
 import { convertToDate } from '@/utils/date'
 import { useCurrentLocale } from '@/utils/l10n/useCurrentLocale'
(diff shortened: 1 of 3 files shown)


apps/store/src/features/carDealership/TrialExtensionForm.tsx
```diff
@@ -29,6 +30,7 @@ type Props = {
 export const TrialExtensionForm = (props: Props) => {
   const { t } = useTranslation(['carDealership', 'checkout'])
   const [userWantsExtension] = useState(true)
+  const [isExtensionOpen, setIsExtensionOpen] = useState(false)
   const { routingLocale } = useCurrentLocale()
```
robinandeer, line 34: Hmmm what is the difference between these states? 🤔 

Aren't they representing the same thing?

apps/store/src/features/carDealership/TrialExtensionForm.tsx
```diff
@@ -106,17 +108,26 @@ export const TrialExtensionForm = (props: Props) => {
 
       <Space y={2}>
         <Space y={1}>
-          <ProductItemContainer
-            offer={selectedOffer}
-            defaultExpanded={true}
-            variant={props.requirePaymentConnection ? 'green' : undefined}
+          <ToggleCard
+            label={t('TOGGLE_EXTENSION_LABEL')}
+            defaultChecked={false}
```
guilhermespopolin, line 113: We should check that but I think it should be checked by default.

apps/store/src/features/carDealership/TrialExtensionForm.tsx
guilhermespopolin: As Robin said, we could rely only on `userWrantsExtension` (need to grab the setter now). That would also require some changes on the markup but that's fine.

apps/store/src/features/carDealership/TrialExtensionForm.tsx
robinandeer: yeah the design was updated yesterday - now it's "on" by default

apps/store/src/features/carDealership/TrialExtensionForm.tsx
gustaveen: Ah I misunderstood that state, though it would be based on some other logic :D I'll use that 🙌

apps/store/src/features/carDealership/TrialExtensionForm.tsx
gustaveen: Right! 🙌

robinandeer (approved):
[goforit](https://app.graphite.dev/macros/HedvigInsurance)

PR #3386, opened 2023-10-09
GEN-1265 - fix(car-dealership): fix tier level order while editing - lowest price to highest price

guilhermespopolin:
## Describe your changes

Show car tier level properly sorted as we do for _OfferPresenter_ (from lowest to highest price) by:
* Promoting `getOffersByPrice` into a utility function;
* Updating `getOffersByPrice` signature so it accepts a generic type with the only constraint of having a `cost.net` field. We don't actually care about any other fields so doing this will make this function more easy to use.

Screenshot 2023-10-09 at 15 41 18

Question: Should we do this on BE already?

It makes more sense that way.


guilhermespopolin:
* main
  * **PR #3386** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3386" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3386?utm_source=stack-comment).

robinandeer (approved):
Why does this work on the offer page? Can we reuse something?

guilhermespopolin:
> Why does this work on the offer page? Can we reuse something?

We do the same there.

> Why does this work on the offer page? Can we reuse something?

Updated 👍🏻

PR #3426, opened 2023-10-13
Payment Link: convert error -> warning logs

robinandeer:
## Describe your changes

Small refactor of Payment Connect page to log warnings instead of errors

## Justify why they are needed

These are not client errors so we should only log warnings since we handle them gracefully

## Checklist before requesting a review

- [ ] I have performed a self-review of my code

PR #3427, opened 2023-10-13
GEN-1302 | Refactor Payment Link flow

robinandeer:
## Describe your changes

- Refactor Payment Link feature

- Create unique endpoints for the different steps (start, ready, success)

- Create Trustly URL server-side

- Allow fallback to login if authorization code has expired

- Make the feature more robust so members can retry and login themselves instead of having to wait for IEX to send them a new link.

- I left the old implementation in place so we can test this new one in parallel.

### DIFF
--- a/apps/store/src/pages/payment/connect/ready.tsx
+++ b/apps/store/src/pages/payment/connect/ready.tsx
@@ -1,0 +1,93 @@
+import { datadogLogs } from '@datadog/browser-logs'
+import { datadogRum } from '@datadog/browser-rum'
+import { type GetServerSideProps } from 'next'
+import { useTranslation } from 'next-i18next'
+import { serverSideTranslations } from 'next-i18next/serverSideTranslations'
+import { useRouter } from 'next/router'
+import { useState } from 'react'
+import { Space, Text } from 'ui'
+import { ErrorDialog } from '@/components/PaymentConnectPage/ErrorDialog'
+import { Layout } from '@/components/PaymentConnectPage/Layout'
+import { initializeApolloServerSide } from '@/services/apollo/client'
+import { getAccessToken } from '@/services/authApi/persist'
+import { createTrustlyUrl } from '@/services/trustly/createTrustlyUrl'
+import { TrustlyIframe } from '@/services/trustly/TrustlyIframe'
+import { isRoutingLocale } from '@/utils/l10n/localeUtils'
+import { PageLink } from '@/utils/PageLink'
+
+const LOGGER = datadogLogs.createLogger('PaymentConnect')
+
+type Props = {
+  trustlyUrl: string
+}
+
+export const getServerSideProps: GetServerSideProps<Props> = async (context) => {
+  if (!isRoutingLocale(context.locale)) return { notFound: true }
+
+  const isLoggedIn = !!getAccessToken({ req: context.req })
+  if (!isLoggedIn) {
+    const redirectUrl = PageLink.memberLogin({
+      locale: context.locale,
+      next: context.resolvedUrl,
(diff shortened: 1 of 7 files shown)


robinandeer:
* main
  * **PR #3426** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3426" target="_blank">Graphite</a> 
    * **PR #3427** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3427" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3427?utm_source=stack-comment).

apps/store/src/services/trustly/TrustlyIframe.tsx
```diff
@@ -74,6 +74,7 @@ const Iframe = styled.iframe(trustlyIframeStyles, {
 
   '&[data-loading=true]': {
     backgroundColor: theme.colors.gray200,
-    animation: `${pulseAnimation} 2s infinite`,
+    animation: `${pulseAnimation} 2s`,
+    animationIterationCount: 3,
```
robinandeer, line 78: I noticed that we sometimes fail to remove the data attribute which causes the animation to run forever

apps/store/src/utils/PageLink.ts
```diff
@@ -193,6 +204,16 @@ export const PageLink = {
   fourOhFour: ({ locale }: BaseParams = {}) => {
     return new URL(`${localePrefix(locale)}/404`, ORIGIN_URL)
   },
+
+  memberLogin: (params: MemberLoginPage = {}) => {
```
alebedev, line 208: Nice!  Let's reuse in member area when this PR is merged

apps/store/src/pages/payment/connect/start.tsx
```diff
@@ -0,0 +1,87 @@
...
+        const reponse = await exchangeAuthorizationCode(authorizationCode)
+        saveAuthTokens(reponse)
+        return await router.push(PageLink.paymentConnectReady())
+      } catch (error) {
+        LOGGER.warn('Failed to exchange authorization code', { error })
+      }
+    }
+
+    const isLoggedIn = !!getAccessToken()
+    if (!authorizationCode && isLoggedIn) {
```
alebedev, line 50: Would it be easier to read as `else if` to previous block?

apps/store/src/pages/payment/connect/ready.tsx
```diff
@@ -0,0 +1,93 @@
...
+  const [isError, setIsError] = useState(false)
+
+  const router = useRouter()
+  const handleSuccess = async () => {
+    LOGGER.info('Successfully connected to Trustly')
+    await router.push(PageLink.paymentConnectSuccess())
+  }
+
+  const handleFail = () => {
+    LOGGER.warn('Failed to connect to Trustly')
```
alebedev, line 65: Should we do logging in `TrustlyIframe` or here?  No strong opinion

apps/store/src/pages/payment/connect/start.tsx
robinandeer: Yeah this one was not totally clear

I started with if else but in general I like early returns

Since we are planning to get rid of this anyways -- perhaps we can merge like this

apps/store/src/pages/payment/connect/ready.tsx
robinandeer: Ahh yes, I prefer to log as early as possible so I agree it should be in iframe -- I can fix that when I integrate

PR #3446, opened 2023-10-17
GEN-1301 - chore: added a custom 'contact us' button until we wait for c1

guilhermespopolin:
## Describe your changes

Adds a custom "contact us" button.

https://github.com/HedvigInsurance/racoon/assets/19200662/bc6812ef-0dca-4dc6-9335-8954e2f6ca19

It takes a long time to get something updated from c1 and we want better support for lazy loading. This can be used while chat is unavailable while makes the app perform better.

### DIFF
--- a/apps/store/src/components/ContactUs/ContactUs.tsx
+++ b/apps/store/src/components/ContactUs/ContactUs.tsx
@@ -1,0 +1,128 @@
+import { keyframes } from '@emotion/react'
+import styled from '@emotion/styled'
+import * as Popover from '@radix-ui/react-popover'
+import { useState, useEffect } from 'react'
+import { MinusIcon, CrossIcon, theme } from 'ui'
+import * as FullScreenDialog from '@/components/FullscreenDialog/FullscreenDialog'
+import { useBreakpoint } from '@/utils/useBreakpoint/useBreakpoint'
+import { ContactUsButton } from './ContactUsButton'
+import { Content } from './Content'
+import { Header } from './Header'
+
+const scaleIn = keyframes({
+  from: {
+    opacity: 0,
+    transform: 'scale(0)',
+  },
+  to: {
+    opacity: 1,
+    transform: 'scale(1)',
+  },
+})
+
+const scaleOut = keyframes({
+  from: {
+    opacity: 1,
+    transform: 'scale(1)',
+  },
+  to: {
+    opacity: 0,
+    transform: 'scale(0)',
+  },
+})
+
+export const ContactUs = () => {
+  const hasMounted = useHasMounted()
+  const isDesktop = useBreakpoint('xs')
+  const [open, setOpen] = useState(false)
+
+  // There's no way to determine how this component should look on the server
+  if (!hasMounted) return null
+
+  const handleClickPhoneLink = () => {
+    setOpen(false)
+  }
+
+  if (isDesktop) {
+    return (
+      <Popover.Root open={open} onOpenChange={setOpen}>
+        <Popover.Trigger asChild={true}>
+          <ContactUsButton />
+        </Popover.Trigger>
+
+        <Popover.Portal>
+          <Popover.Content side="bottom" align="end" sideOffset={8} asChild={true}>
+            <ChatWindow>
+              <Header
+                CloseBtn={
+                  <Popover.Close asChild={true}>
+                    <CloseButton>
+                      <MinusIcon size="18px" />
+                    </CloseButton>
+                  </Popover.Close>
+                }
+              />
+
+              <Content onClickPhoneLink={handleClickPhoneLink} />
+            </ChatWindow>
+          </Popover.Content>
+        </Popover.Portal>
+      </Popover.Root>
+    )
+  }
+  return (
+    <FullScreenDialog.Root open={open} onOpenChange={setOpen}>
+      <FullScreenDialog.Trigger asChild={true}>
+        <ContactUsButton />
+      </FullScreenDialog.Trigger>
+
+      <FullScreenDialog.Modal
+        Header={
+          <Header
+            CloseBtn={
+              <FullScreenDialog.Close asChild={true}>
+                <CloseButton>
+                  <CrossIcon size="24px" />
+                </CloseButton>
+              </FullScreenDialog.Close>
+            }
+          />
+        }
+      >
+        <Content onClickPhoneLink={handleClickPhoneLink} />
+      </FullScreenDialog.Modal>
+    </FullScreenDialog.Root>
+  )
+}
+
+const useHasMounted = () => {
+  const [hasMounted, setHasMounted] = useState(false)
+
+  useEffect(() => {
+    setHasMounted(true)
+  }, [])
+
+  return hasMounted
+}
+
+const ChatWindow = styled.div({
+  display: 'grid',
+  gridTemplateRows: '55px 1fr',
+  height: 500,
+  aspectRatio: '4 / 5',
+  backgroundColor: theme.colors.offWhite,
+  borderRadius: theme.radius.md,
+  transformOrigin: 'var(--radix-popover-content-transform-origin)',
+  filter: 'drop-shadow(0px 1px 1px hsl(0deg 0% 0% / 0.15))',
+
+  '&[data-state=open]': {
+    animation: `${scaleIn} 250ms ${theme.transitions.css.easeInCubic}`,
+  },
+  '&[data-state=closed]': {
+    animation: `${scaleOut} 250ms ${theme.transitions.css.easeOutCubic}`,
+  },
+})
+
+const CloseButton = styled.button({
+  cursor: 'pointer',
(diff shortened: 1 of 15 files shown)


guilhermespopolin:
* main
  * **PR #3446** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3446" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3446?utm_source=stack-comment).

apps/store/src/components/ContactUs/ContactUs.tsx
```diff
@@ -0,0 +1,238 @@
...
+  },
+})
+
+export const ContactUs = () => {
+  const { t } = useTranslation('contact-us')
+  const isDesktop = useBreakpoint('lg')
+  const { routingLocale } = useCurrentLocale()
+  const [open, setOpen] = useState(false)
+
+  if (!isDesktop) {
```
alebedev, line 51: We have single caller, can we move it there?  I feel like using `return null` to render conditionally usually makes code harder to reason about.  Perhaps we should avoid it where practical

apps/store/src/components/ContactUs/ContactUs.tsx
```diff
@@ -0,0 +1,238 @@
...
+    return null
+  }
+
+  const handleOpenChange = (open: boolean) => setOpen(open)
+
+  return (
+    <Popover.Root open={open} onOpenChange={handleOpenChange}>
+      <Popover.Trigger asChild={true}>
+        <FixedButton>
```
robinandeer, line 60: Why not using a `styled(Popover.Trigger)`?

apps/store/src/components/ContactUs/ContactUs.tsx
```diff
@@ -0,0 +1,238 @@
...
+  const [open, setOpen] = useState(false)
+
+  if (!isDesktop) {
+    return null
+  }
+
+  const handleOpenChange = (open: boolean) => setOpen(open)
+
+  return (
+    <Popover.Root open={open} onOpenChange={handleOpenChange}>
```
alebedev, line 58: Nit: just pass `setOpen` directly if you're not transforming it in any way

apps/store/src/components/ContactUs/ContactUs.tsx
```diff
@@ -0,0 +1,238 @@
...
+              <CloseButton>
+                <MinusIcon size="16px" />
+              </CloseButton>
+            </Popover.Close>
+          </Header>
+
+          <Content y={2}>
+            <Paper>
+              <Space y={1}>
+                <Heading as="h1" variant="standard.18" align="center">
```
robinandeer, line 80: what is the correct heading level in this case? A page should only have one "h1" 🤔

apps/store/src/components/ContactUs/ContactUs.tsx
```diff
@@ -0,0 +1,238 @@
...
+}
+
+const FixedButton = styled.button({
+  position: 'fixed',
+  right: theme.space.lg,
+  bottom: theme.space.lg,
+  display: 'inline-flex',
+  alignItems: 'center',
+  gap: theme.space.sm,
+  zIndex: 999,
```
alebedev, line 143: Is this the case for another zIndex in theme or it's OK to keep it as one-off thing?

apps/store/src/components/ContactUs/ContactUs.tsx
```diff
@@ -0,0 +1,238 @@
...
+  aspectRatio: '1 / 1',
+  borderRadius: '50%',
+  backgroundColor: theme.colors.green600,
+})
+
+const PopoverContent = styled(Popover.Content)({
+  display: 'grid',
+  gridTemplateRows: '55px 1fr',
+  height: 500,
+  aspectRatio: '0.80 / 1',
```
alebedev, line 165: Is `4 / 5` the same thing?  It's probably easier to read if so

apps/store/src/components/ContactUs/ContactUs.tsx
```diff
@@ -0,0 +1,238 @@
...
+  backgroundColor: theme.colors.white,
+  borderRadius: theme.radius.md,
+  transformOrigin: 'var(--radix-popover-content-transform-origin)',
+
+  '&[data-state=open]': {
+    animation: `${scaleIn} 350ms cubic-bezier(0.65,0.05,0.36,1)`,
```
alebedev, line 171: Why not `transtions.framer.easeInCubic / easeOutCubic` ?

packages/ui/src/components/Button/Button.tsx
```diff
@@ -24,7 +25,13 @@ export const Button = forwardRef<HTMLButtonElement, Props>((props, ref) => {
...
   const buttonChildren = (
     <>
-      <span style={{ opacity: loading ? 0 : 1 }}>{children}</span>
+      {React.Children.map(children, (child) =>
+        React.isValidElement<{ style?: React.CSSProperties }>(child) ? (
+          React.cloneElement(child, { style: { ...child.props.style, opacity: loading ? 0 : 1 } })
+        ) : (
+          <span style={{ opacity: loading ? 0 : 1 }}>{child}</span>
+        ),
+      )}
```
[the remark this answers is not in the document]
alebedev: Looks good, can you check performance effects of this?

robinandeer (commented):
The button lacks drop shadow, no?

The animation feels a little slow - maybe we can speed it up a bit?

apps/store/src/components/ContactUs/ContactUs.tsx
alebedev: Can we convert same values to CSS format to avoid copy-pasting them?

apps/store/src/components/ContactUs/ContactUs.tsx
gustaveen: Feels like a it could be a good case for a theme z value :)

apps/store/src/components/ContactUs/ContactUs.tsx
guilhermespopolin: I agree with you. Based on recent discussions though we might want to still display the button on mobile screen size so this conditional rendering is gonna to remove all together.

apps/store/src/components/ContactUs/ContactUs.tsx
guilhermespopolin: I see. I though about the popover as a separate context, being ok to add a new h1 heading. However it seems that might be an issue. Gonna change it to h2.

gustaveen (commented):
Great work! @guilhermespopolin 🥳

Yes, maybe we can speed up the animation a bit, kind of like [Intercom](https://www.intercom.com/drlp/platform)

apps/store/src/components/ContactUs/ContactUs.tsx
guilhermespopolin: 🤔 I guess we might wan to move forward with a custom button. In that case I think it makes sense to add a zIndex for it in theme.

apps/store/src/components/ContactUs/ContactUs.tsx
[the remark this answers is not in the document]
guilhermespopolin: I can give it a try. On their [API](https://mui.com/material-ui/api/button/#Button-prop-children) children can be a ReactNode and we still need to hide button's content when loading 🤔. If we get a `startIcon`/`endIcon` prop then I think we would still need to wrap those into something that contains the "hide" feature. E.g:

```
// Input
<Button startIcon={<Icon />}>Label</Button>

// Output
<button>
  <span style={{ opacity: loading ? 0 : 1 }}>
    <svg></svg>
    Label
  </span>
</button>
```

packages/ui/src/components/Button/Button.tsx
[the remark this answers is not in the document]
guilhermespopolin: I can give it a try. On their [API](https://mui.com/material-ui/api/button/#Button-prop-children) children can be a ReactNode and we still need to hide button's content when loading 🤔. If we get a `startIcon`/`endIcon` prop then I think we would still need to wrap those into something that contains the "hide" feature. E.g:

```
// Input
<Button startIcon={<Icon />}>Label</Button>

// Output
<button>
  <span style={{ opacity: loading ? 0 : 1 }}>
    <svg></svg>
    Label
  </span>
</button>
```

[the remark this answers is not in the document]
guilhermespopolin: Gonna address that down in the stack.

guilhermespopolin:
### Merge activity

* **Oct 24, 7:57 AM**: @guilhermespopolin started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3446).

PR #3459, opened 2023-10-18
Make sure sitemap exludes pages with noindex

gustaveen:
## Describe your changes
Refactor sitemap generation

Since the `link` endpoint unfortunately don't support the [filter queries](https://www.storyblok.com/docs/api/content-delivery/v2#filter-queries/overview) or contain story information, we need to use the `stories` endpoint.

So what we filter on is excluding all stories with `robots: noindex`

Only concern is if we would hit the API rate limit

We need to exclude `noindex` pages from the sitemap

### DIFF
--- a/apps/store/src/pages/new-sitemap-pending.xml.js
+++ b/apps/store/src/pages/new-sitemap-pending.xml.js
@@ -1,4 +1,4 @@
-import { getFilteredPageLinks } from '@/services/storyblok/storyblok'
+import { fetchStories } from '@/services/storyblok/storyblok'
 import { ORIGIN_URL } from '@/utils/PageLink'
 
 const generateSiteMap = (pages) => {
@@ -8,7 +8,7 @@ const generateSiteMap = (pages) => {
        .map((page) => {
          return `
           <url>
-            <loc>${`${ORIGIN_URL}/${page.link.slug}`}</loc>
+            <loc>${`${ORIGIN_URL}/${page.full_slug}`}</loc>
           </url>
         `
        })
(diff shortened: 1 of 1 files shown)


apps/store/src/pages/new-sitemap-pending.xml.js
```diff
@@ -21,8 +21,37 @@ const SiteMap = () => {
   // getServerSideProps will do the heavy lifting
 }
 
+const getFilteredPages = async () => {
+  /**
+   * Sitemap should exlude pages that are:
```
guilhermespopolin, line 26: extreme nit:
```suggestion
   * Sitemap should exclude pages that are:
```

apps/store/src/pages/new-sitemap-pending.xml.js
```diff
@@ -21,8 +21,37 @@ const SiteMap = () => {
...
+    paginating = response.headers.total <= currentPage * perPage
+    currentPage++
+  }
+
+  return filteredPages
+}
+
 export const getServerSideProps = async ({ res }) => {
-  const pageLinks = await getFilteredPageLinks()
+  const pageLinks = await getFilteredPages()
```
guilhermespopolin, line 54: Wouldn't be the case of updating `getFiletedPageLinks` function so it covers that requirement? 🤔

guilhermespopolin (approved):
Nice. just left some comments

apps/store/src/pages/new-sitemap-pending.xml.js
gustaveen: I wish, but since that function gets `links` we have no access to the story data and can therefore not filter on `noindex`. Also all other use cases we want to get the links for pages that has `noindex`, it's just for the sitemap we'd like to exclude them :) @guilhermespopolin

gustaveen:
### Merge activity

* **Oct 19, 5:22 AM**: @gustaveen merged this pull request with [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3459).

PR #3461, opened 2023-10-19
Set cache version on sitemap request

gustaveen:
## Describe your changes
Set cache version to avoid rate limit on requests

Does this look valid to you? :)

### DIFF
--- a/apps/store/src/pages/new-sitemap-pending.xml.js
+++ b/apps/store/src/pages/new-sitemap-pending.xml.js
@@ -21,6 +21,7 @@ const SiteMap = () => {
   // getServerSideProps will do the heavy lifting
 }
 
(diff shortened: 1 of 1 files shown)


apps/store/src/pages/new-sitemap-pending.xml.js
```diff
@@ -35,12 +36,17 @@ const getFilteredPages = async () => {
   let perPage = 100
   let paginating = false
 
+  const cacheVersion = STORYBLOK_CACHE_VERSION ? parseInt(STORYBLOK_CACHE_VERSION) : NaN
```
guilhermespopolin, line 39: Maybe we can do this?

```suggestion
  const cacheVersion = parseInt(STORYBLOK_CACHE_VERSION)
```

robinandeer (approved):
Looks good to me!

guilhermespopolin (approved):

PR #3462, opened 2023-10-19
Fix error when getting locale from URL

robinandeer:
## Describe your changes

- Fix error in `getUrlLocale` function

- Get the second part of the url which is the locale

- Add tests for the function

- The function was not working properly which caused us to detect a country change when it was not the case. We never noticed this since we were only like in Sweden which is already the default market.


robinandeer:
* main
  * **PR #3462** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3462" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3462?utm_source=stack-comment).

guilhermespopolin (approved):

PR #3469, opened 2023-10-20
GEN-1369 | Fix bug causing us to create duplicated price intents

robinandeer:
## Describe your changes

- Reuse existing price intent service instance instead of creating new ones

- Update how to compare input params when creating new price intents

- We deduplicate requests to create price intents, however, we store the promise on the class instance. We also create new instances of the class for each request. This means that we never reuse the saved promise. This PR fixes that by reusing the existing instance.

- We also compare the input params to determine if we should create a new price intent. However, the compare method was quite crude so I changed it to be more specific and only compare the fields that are relevant to the price intent.

### DIFF
--- a/apps/store/src/services/priceIntent/PriceIntentService.ts
+++ b/apps/store/src/services/priceIntent/PriceIntentService.ts
@@ -82,8 +82,12 @@ export class PriceIntentService {
     }
 
     // Deduplicate mutation, Apollo won't do this for us
-    const paramsEquals = (a: unknown, b: unknown) => JSON.stringify(a) === JSON.stringify(b)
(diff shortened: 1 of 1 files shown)


robinandeer:
* main
  * **PR #3469** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3469" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3469?utm_source=stack-comment).

alebedev (approved):
Nice, let's do it

guilhermespopolin (approved):
Nice catch :)

PR #3479, opened 2023-10-24
Render regular a link for links of type url

gustaveen:


gustaveen:
* main
  * **PR #3479** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3479" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3479?utm_source=stack-comment).

apps/store/src/components/Header/NavigationLink.tsx
```diff
@@ -32,13 +32,22 @@ const ProductNavigationLinkCard = styled(Space)({
...
     pillowImageSrc?: string
     label?: string
+    absoluteUrl?: boolean
+  }
+
+export const NavigationLink = ({ href, children, absoluteUrl, ...rest }: NavigationLinkProps) => {
+  if (absoluteUrl) {
+    return (
+      <a href={href}>
+        <StyledNavigationLink {...rest}>{children}</StyledNavigationLink>
```
robinandeer, line 46: isn't this already an anchor tag?

apps/store/src/components/Header/NavigationLink.tsx
```diff
@@ -71,7 +80,7 @@ export const ProductNavigationLink = ({
   label,
 }: NavigationLinkProps) => {
   return (
-    <Link href={href}>
+    <a href={href}>
```
robinandeer, line 83: Why do we need to change this?

apps/store/src/components/Header/NavigationLink.tsx
```diff
@@ -32,13 +32,22 @@ const ProductNavigationLinkCard = styled(Space)({
...
 
-type NavigationLinkProps = Pick<LinkProps, 'href'> &
+type NavigationLinkProps = Pick<HTMLAnchorElement, 'href'> &
   Omit<NavigationMenuPrimitive.NavigationMenuLinkProps, 'href'> & {
     pillowImageSrc?: string
     label?: string
+    absoluteUrl?: boolean
+  }
+
+export const NavigationLink = ({ href, children, absoluteUrl, ...rest }: NavigationLinkProps) => {
```
robinandeer, line 42: Would it make more sense to check inside this component if we are linking with absolute URL?

What does it mean that `absoluteUrl = false` and `href = "https://www.google.com"` for example? That seems very ambiguous - should we not just derive it from the href?

apps/store/src/components/Header/NavigationLink.tsx
gustaveen: Oops, yes it is :D

apps/store/src/components/Header/NavigationLink.tsx
gustaveen: It's a next link, and that's for client-side navigation between next routes right? So to cause a full reload on navigation we need to use a regular link was my though. Or am I missing something :)

apps/store/src/components/Header/NavigationLink.tsx
gustaveen: I wanted to derive it from the `linktype`: `url` and not just the `href`. Or are we sure a `linktype`: `story` always is a relative path? Maybe it is then I think it's fine. Otherwise we can pass linktype then

robinandeer: We never tell Storyblok anything about which domain we use for the website so the only way it could provide a absolute URL is if an editor added it explicitly, no?

We could validate this but that would be my guess

On the other hand, I think I can just type in `/se/about` or any relative URL in `linktype: url`, no? :)

apps/store/src/components/Header/NavigationLink.tsx
robinandeer: Hmm yes that's correct, however, I guess I couldn't find that we even render these in non-SE markets

So why couldn't we use client side routing? :/

apps/store/src/components/Header/NavigationLink.tsx
gustaveen: For sure, `linktype: url` could definitely be relative path if someone wrote it. Just though it would make sense as a distinction on what link to render :) But you're right, stories should always be relative 👍

alebedev (approved):
Went over it in Slack, let's go!

gustaveen:
### Merge activity

* **Oct 25, 5:07 AM**: @gustaveen merged this pull request with [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3479).

PR #3498, opened 2023-10-27
GEN-1384 - feat: added widgetFlow content type

guilhermespopolin:
## Describe your changes

* Added a `widget-flow` Content type so editors can configure some aspects of a widget flow.

Screenshot 2023-10-27 at 14 46 43

### DIFF
--- a/apps/store/src/blocks/WidgetFlowBlock.tsx
+++ b/apps/store/src/blocks/WidgetFlowBlock.tsx
@@ -1,0 +1,74 @@
+import styled from '@emotion/styled'
+import { StoryblokComponent, storyblokEditable, type SbBlokData } from '@storyblok/react'
+import { Space } from 'ui'
+import { GridLayout } from '@/components/GridLayout/GridLayout'
+import { ProductItemSkeleton } from '@/components/ProductItem/ProductItem'
+import { Skeleton } from '@/components/Skeleton'
+import { SpaceFlex } from '@/components/SpaceFlex/SpaceFlex'
+import { SbBaseBlockProps } from '@/services/storyblok/storyblok'
+
+type Props = SbBaseBlockProps<{
+  checkoutPageContent: Array<SbBlokData>
+}>
+
+export const WidgetFlowBlock = ({ blok }: Props) => {
+  return (
+    <GridLayout.Root {...storyblokEditable(blok)}>
+      <GridLayout.Content width="1/3" align="center">
+        <CheckoutContentPlaceholder />
+      </GridLayout.Content>
+
+      <GridLayout.Content width="1" align="center">
+        {blok.checkoutPageContent.map((nestedBlock) => (
+          <StoryblokComponent key={nestedBlock._uid} blok={nestedBlock} />
+        ))}
+      </GridLayout.Content>
+    </GridLayout.Root>
+  )
+}
+WidgetFlowBlock.blockName = 'widget-flow'
+
+const RecoomendationSkeleton = styled(Skeleton)({
+  height: '16rem',
+})
+
+const SignButtonSkeleton = styled(Skeleton)({
+  height: '3.5rem',
+})
+
+const StepSkeleton = styled(Skeleton)({
+  flex: 1,
+  height: '2.125rem',
+})
+
+const BreadCrumbSkeleton = () => {
+  return (
+    <SpaceFlex direction="horizontal" space={1}>
+      <StepSkeleton />
+      <StepSkeleton />
+      <StepSkeleton />
+    </SpaceFlex>
+  )
+}
+
+const HeadingSkeleton = styled(Skeleton)({
+  height: '5.625rem',
+})
+
+const CheckoutContentPlaceholder = () => {
+  return (
+    <Space style={{ marginTop: '1rem' }} y={4}>
+      <BreadCrumbSkeleton />
+
+      <HeadingSkeleton />
+
+      <Space y={1}>
+        <Space y={2}>
+          <ProductItemSkeleton />
+          <RecoomendationSkeleton />
+        </Space>
+        <SignButtonSkeleton />
+      </Space>
+    </Space>
+  )
+}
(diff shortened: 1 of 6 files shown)


guilhermespopolin:
* main
  * **PR #3498** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3498" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3498?utm_source=stack-comment).

apps/store/src/pages/widget/[[...slug]].tsx
```diff
@@ -0,0 +1,108 @@
...
+
+const NextWidgetContentPage = (props: NextWidgetContentPageProps) => {
+  const story = useStoryblokState(props[STORY_PROP_NAME])
+
+  if (!story) {
+    return null
+  }
+
+  return <StoryblokComponent blok={story.content} />
+}
```
Comment (guilhermespopolin) [inline apps/store/src/pages/widget/[[...slug]].tsx:56]: So those two are exactly the same but I was having a type issue if I try to initialise `useStoryblokState` with something that can assume two different types that are incompatible. Any ideas?

Comment (alebedev) [reply apps/store/src/pages/widget/[[...slug]].tsx:56]: Did you try having single Typescript type with `  type: 'content' | 'flow'`?  Maybe having single type to feed into `useStoryblokState` solves it

apps/store/src/pages/widget/[[...slug]].tsx
```diff
@@ -0,0 +1,108 @@
...
+
+export const getStaticProps: GetStaticProps<PageProps, StoryblokQueryParams> = async (context) => {
+  const { locale, params, draftMode } = context
+
+  if (!isRoutingLocale(locale)) {
+    return { notFound: true }
+  }
+
+  const slug = `${STORYBLOK_WIDGET_FOLDER_SLUG}/${(params?.slug ?? []).join('/')}`
```
Comment (alebedev) [inline apps/store/src/pages/widget/[[...slug]].tsx:65]: Can we take `widget` here from full URL instead of making it another constant that we need to keep in sync between code and Storyblok?

apps/store/src/services/storyblok/Storyblok.helpers.ts
```diff
@@ -85,6 +85,10 @@ export const isProductStory = (story: ISbStoryData): story is ProductStory => {
   return story.content.component === 'product'
 }
 
+export const isWidgetFlowStory = (story: ISbStoryData): story is WidgetFlowStory => {
```
alebedev, line 88: General questions
- Do you need to change anything with `getStaticPaths` on generic CMS page to avoid including duplicates there?
- Do we need to change sitemap generation?

alebedev (commented):
Good start here!  Posted some high-level questions, will look deeper when we have clarity on those

apps/store/src/blocks/WidgetFlowBlock.tsx
```diff
@@ -0,0 +1,74 @@
...
+        {blok.checkoutPageContent.map((nestedBlock) => (
+          <StoryblokComponent key={nestedBlock._uid} blok={nestedBlock} />
+        ))}
+      </GridLayout.Content>
+    </GridLayout.Root>
+  )
+}
+WidgetFlowBlock.blockName = 'widget-flow'
+
+const RecoomendationSkeleton = styled(Skeleton)({
```
gustaveen, line 31:
```suggestion
const RecommendationSkeleton = styled(Skeleton)({
```

apps/store/src/services/storyblok/Storyblok.helpers.ts
gustaveen: Yes, we should exclude it from the sitemap 👍

gustaveen (commented):
Great start with this! 🙌

apps/store/src/blocks/WidgetFlowBlock.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+      <GridLayout.Content width="1" align="center">
+        {blok.checkoutPageContent.map((nestedBlock) => (
+          <StoryblokComponent key={nestedBlock._uid} blok={nestedBlock} />
+        ))}
+      </GridLayout.Content>
+    </GridLayout.Root>
+  )
+}
+WidgetFlowBlock.blockName = 'widget-flow'
```
gustaveen, line 29: We should probably stick to camelCase naming for bloks :)

apps/store/src/blocks/WidgetFlowBlock.tsx
guilhermespopolin: Thanks.

apps/store/src/services/storyblok/Storyblok.helpers.ts
guilhermespopolin: * Since we're excluding widget folder content in `getFilteredLinks` function I think we'll not be duplicating those.
* Right, gonna exclude it from the sitemap

apps/store/src/pages/widget/[[...slug]].tsx
Comment (guilhermespopolin) [reply apps/store/src/pages/widget/[[...slug]].tsx:65]: The think is we don't have access to full URL inside `getStaticProps`, or maybe I misunderstood what you're proposing? 🤔

apps/store/src/pages/widget/[[...slug]].tsx
Comment (guilhermespopolin) [reply apps/store/src/pages/widget/[[...slug]].tsx:56]: I doesn't work 😢 
The issue is with the story type. Keeping a single type would make `PageProps` be defined like this:

```
type PageProps = {
  type: 'content' | 'flow'
  [STORY_PROP_NAME]: PageStory | WidgetFlowStory
  hideChat?: boolean
}
```
The problem is that `PageStory` and `WidgetFlowStory` are incompatible types. E.g: `body` field is required for `PageStory` but not for `WidgetFlowStory`.

Comment (guilhermespopolin) [reply apps/store/src/pages/widget/[[...slug]].tsx:56]: I think I've found a way. The code is updated it with.

alebedev (approved):
Looks like a good foundation now.  We should probably improve things when we learn more about widget flows, but let's go ahead and merge this now

gustaveen (approved):
Nice, let's merge!

guilhermespopolin:
### Merge activity

* **Nov 1, 6:39 AM**: @guilhermespopolin started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3498).

PR #3507, opened 2023-11-01
Update query with new carTrial endpoint

gustaveen:


gustaveen:
* main
  * **PR #3507** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3507" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3507?utm_source=stack-comment).

alebedev (commented):
> customer.ssn can be undefined, how should we treat that case? just show loader or some sort of error?

Let's crash, this is invalid state where can cannot do anything useful

apps/store/src/features/carDealership/CarTrialExtensionBlock.tsx
```diff
@@ -87,15 +92,11 @@ const useCarTrialQuery = (params: UseCarTrialQueryParams) => {
...
-    return getTrialExtension(data)
-  }
-
-  return null
+  return data
 }
 
 type AddNotificationBannerOptions = {
-  data: TrialExtension
+  data: CarTrialExtension
```
guilhermespopolin, line 99: Right now `CarTrialExtension` type can assume the "actual data" | null | undefined.
It would be nice if we could right code here assuming `data: CarTrialExtension` is defined and non nullable and make sure we only call `addNotificationBanner` if we actually have some data because right now we need to right:

```
addBanner(
  t('REMAIN_INSURED_BANNER', {
    dueDate: `<b>${dateFull(new Date(data?.trialContract.terminationDate))}</b>`,
  }),
  'warning',
  { force: true },
)
```

and if `data` is nullable here we'll end up showing some weird info in that banner.

But maybe we can do that in another PR, as part of a refactoring.

```suggestion
  data: NonNullable<CarTrialExtension>
```

gustaveen: Yes, I had a bit of problems with the the banner and typing :) I was thinking of handling in a new PR but let's sync and we can try to figure it out 🙌 @guilhermespopolin

guilhermespopolin (approved):

PR #3520, opened 2023-11-06
Create reusable Radio Option List component

robinandeer:
## Describe your changes

- Create a reusable Radio Option List component (based on design system)

- Support the product option item

- Add storybook story

- Needed for the hedvig widget implementation

- We already use it in the price calculator. We should extract it from there and make it reusable. However, I thought this would be an easier first step to review.

### DIFF
--- a/apps/store/src/components/RadioOptionList/RadioOptionList.tsx
+++ b/apps/store/src/components/RadioOptionList/RadioOptionList.tsx
@@ -1,0 +1,82 @@
+import styled from '@emotion/styled'
+import * as RadioGroup from '@radix-ui/react-radio-group'
+import { MouseEventHandler, type ComponentProps } from 'react'
+import { Text, theme } from 'ui'
+import { Pillow } from '@/components/Pillow/Pillow'
+import { useHighlightAnimation } from '@/utils/useHighlightAnimation'
+
+type RootProps = RadioGroup.RadioGroupProps
+
+export const Root = (props: RootProps) => {
+  return <RadioGroup.Root {...props}>{props.children}</RadioGroup.Root>
+}
+
+type ProductOptionProps = RadioGroup.RadioGroupItemProps & {
+  title: string
+  subtitle: string
+  pillow: ComponentProps<typeof Pillow>
+}
+
+export const ProductOption = (props: ProductOptionProps) => {
+  const { highlight, animationProps } = useHighlightAnimation<HTMLLabelElement>()
+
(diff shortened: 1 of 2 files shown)


robinandeer:
* main
  * **PR #3520** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3520" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3520?utm_source=stack-comment).

apps/store/src/components/RadioOptionList/RadioOptionList.tsx
```diff
@@ -0,0 +1,82 @@
...
+type OptionProductProps = RadioGroup.RadioGroupItemProps & {
+  title: string
+  subtitle: string
+  pillow: ComponentProps<typeof Pillow>
+}
+
+export const ProductOption = (props: OptionProductProps) => {
+  const { highlight, animationProps } = useHighlightAnimation<HTMLLabelElement>()
+
+  const handleClick: MouseEventHandler<HTMLButtonElement> = (event) => {
```
alebedev, line 23: Nit: should we react on state change instead of click?  For example, user triggers selection from keyboard, we probably want the same animation, right?

alebedev (approved):
Nice!  One special effects comment

apps/store/src/components/RadioOptionList/RadioOptionList.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+export const Root = (props: RootProps) => {
+  return <RadioGroup.Root {...props}>{props.children}</RadioGroup.Root>
+}
+
+type OptionProductProps = RadioGroup.RadioGroupItemProps & {
+  title: string
+  subtitle: string
+  pillow: ComponentProps<typeof Pillow>
+}
```
guilhermespopolin, line 18: Maybe?!
```suggestion
type ProductOptionProps = RadioGroup.RadioGroupItemProps & {
  title: string
  subtitle: string
  pillow: ComponentProps<typeof Pillow>
}
```

guilhermespopolin (approved):
Nice job

apps/store/src/components/RadioOptionList/RadioOptionList.tsx
robinandeer: Do you have a suggestion how that would work? I would like to avoid making the component controlled just for the sake of animations

PR #3521, opened 2023-11-07
GEN-1385 | Add radio input to widget select product page

robinandeer:
## Describe your changes

- Add radio input for selecting product in widget flow

- Break out page and put inside feature folder

- Add link to next step in widget flow (calculate price)

- Add helpers to fetch correct price template based on product name

- We can't test it quite yet since we don't have the widget products in the backend yet, but we can test the UI. Just go to: `/se/widget/test/test/select`.

- I will continue to work on this in the next PRs and hook up to get correct product data from backend.

### DIFF
--- a/apps/store/src/features/widget/SelectProductPage.tsx
+++ b/apps/store/src/features/widget/SelectProductPage.tsx
@@ -1,0 +1,141 @@
+import { useApolloClient } from '@apollo/client'
+import styled from '@emotion/styled'
+import { useRouter } from 'next/router'
+import { useState, type ComponentProps, type FormEventHandler } from 'react'
+import { Button, Heading, HedvigLogo, Space, mq, theme } from 'ui'
+import { GridLayout } from '@/components/GridLayout/GridLayout'
+import { STYLES } from '@/components/GridLayout/GridLayout.helper'
+import * as ProgressIndicator from '@/components/ProgressIndicator/ProgressIndicator'
+import * as RadioOptionList from '@/components/RadioOptionList/RadioOptionList'
+import { priceIntentServiceInitClientSide } from '@/services/priceIntent/PriceIntentService'
+import { PageLink } from '@/utils/PageLink'
+import { type WidgetProductName, getPriceTemplate, isWidgetProductName } from './widget.helpers'
+
+const PRODUCTS: Array<{
+  name: WidgetProductName
+  title: string
+  subtitle: string
+  pillow: ComponentProps<typeof RadioOptionList.ProductOption>['pillow']
+}> = [
+  {
+    name: 'SE_WIDGET_APARTMENT_RENT',
+    title: 'Hyresrätt',
+    subtitle: 'För dig som hyr bostad',
+    pillow: {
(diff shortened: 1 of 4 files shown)


robinandeer:
* main
  * **PR #3520** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3520" target="_blank">Graphite</a> 
    * **PR #3521** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3521" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3521?utm_source=stack-comment).

alebedev (approved):
Looks good!

apps/store/src/features/widget/SelectProductPage.tsx
```diff
@@ -0,0 +1,141 @@
...
+                        subtitle={item.subtitle}
+                        pillow={item.pillow}
+                      />
+                    ))}
+                  </Space>
+                </RadioOptionList.Root>
+
+                <CustomButton type="submit" variant="primary" data-disabled={!productName}>
+                  Continue
+                </CustomButton>
```
guilhermespopolin, line 110: I saw that the idea is to keep the button clickable while it appears disabled when `productName` is `undefined`. Is that really necessary? I mean, by clicking on it while disabled makes the app break which is too harsh I think. 
If we could only prevent it to be clickable while disabled and use that state for applying styles, things would be simpler no? We can also add `aria-disabled` so assistive technologies can know about that. Do you think that would solve the A11y issues behind the addition of `data-disabled`?

```suggestion
                <CustomButton type="submit" variant="primary" disabled={"productName} aria-disabled={!productName}>
                  Continue
                </CustomButton>
```

guilhermespopolin (approved):
Nice. Just left a comment :)

apps/store/src/features/widget/SelectProductPage.tsx
robinandeer: > by clicking on it while disabled makes the app break which is too harsh I think

Not sure what you mean here. If you click the button without selecting a product, we just show native form validation error.

> ...prevent it to be clickable while disabled and use that state for applying styles, things would be simpler no?

A disabled button is not clickable. However, in this case I want the button to remain clickable so assistive technologies can access it and support native error handling.

Aria labels is probably a good idea tho!

gustaveen (approved):
GJ! 🥳

apps/store/src/features/widget/SelectProductPage.tsx
guilhermespopolin: Jaha, I wrote these because when testing this branch locally clicking on the button I as getting the following:

image

But doesn't happen anymore due the recent changes. Well, never mind then :)

robinandeer: oh that's strange indeed -- then I totally understand your comment

robinandeer:
### Merge activity

* **Nov 8, 10:08 AM**: @robinandeer started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3521).

PR #3528, opened 2023-11-08
Add PriceBreakdown to PayForTrial component

gustaveen:
## Describe your changes
Add PriceBreakdown to PayForTrial component

Screenshot 2023-11-08 at 15 39 54

## Checklist before requesting a review

- [ ] I have performed a self-review of my code

PR #3524, opened 2023-11-08
Add new PriceBreakdown component for car dealership

gustaveen:
## Describe your changes
A new component for explaining the cost for trial and extension separately.  
I didn't want to use the `Price` component where we send in a reduced price. The logic is the other way around in this case. The trial price is the fixed price (`amount`) and the default price that we want to strike through is the price for the same tier in full price (`defaultAmount`). 

Screenshot 2023-11-08 at 11 53 11

## Justify why they are needed
It doesn't make sense to show a total price for explaining the price the member will pay for 2 months trial and then the extension

## Checklist before requesting a review

- [ ] I have performed a self-review of my code

PR #3529, opened 2023-11-08
Add check for crossed over price to catch cases where offer would be cheaper than trial

gustaveen:
## Describe your changes
Make sure we never show crossed over price that's cheaper than the trial amount. Fred mentioned rare cases where offer would be cheaper than trial and this also would happen if default offer was not set to match trial tier.

Alos adressed comments from first PR

Avoid this case:
Screenshot 2023-11-08 at 15 58 59

### DIFF
--- a/apps/store/src/features/carDealership/PriceBreakdown.tsx
+++ b/apps/store/src/features/carDealership/PriceBreakdown.tsx
@@ -5,7 +5,7 @@ import { useFormatter } from '@/utils/useFormatter'
 
 type Props = {
   amount: number
-  defaultAmount?: number
+  crossedOverAmount?: number
   currencyCode: CurrencyCode
   title: string
   subTitle: string
@@ -14,23 +14,23 @@ type Props = {
 
 export const PriceBreakdown = ({
   amount,
-  defaultAmount,
+  crossedOverAmount,
   currencyCode,
   title,
   subTitle,
   priceExplanation,
 }: Props) => {
   const formatter = useFormatter()
-
+  const showCrossedOverAmount = crossedOverAmount !== undefined && crossedOverAmount > amount
   return (
     <div>
       <Row>
         <Text>{title}</Text>
         <Wrapper>
-          {defaultAmount !== undefined && (
+          {showCrossedOverAmount && (
             <Text as="p" size="md" strikethrough={true} color="textSecondary">
               {formatter.monthlyPrice({
-                amount: defaultAmount,
+                amount: crossedOverAmount,
                 currencyCode: currencyCode,
               })}
             </Text>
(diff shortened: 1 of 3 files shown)


gustaveen:
* main
  * **PR #3524** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3524" target="_blank">Graphite</a> 
    * **PR #3528** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3528" target="_blank">Graphite</a> 
      * **PR #3529** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3529" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3529?utm_source=stack-comment).

apps/store/src/features/carDealership/PriceBreakdown.tsx
```diff
@@ -14,23 +14,23 @@ type Props = {
...
-  defaultAmount,
+  crossedOverAmount,
   currencyCode,
   title,
   subTitle,
   priceExplanation,
 }: Props) => {
   const formatter = useFormatter()
-
+  const showCrossedOverAmount = crossedOverAmount !== undefined && crossedOverAmount > amount
```
robinandeer, line 24: Should this be a presentational component? In that case, perhaps it's better to make this check when rendering `PriceBreakdown`

apps/store/src/features/carDealership/TrialExtensionForm.tsx
```diff
@@ -167,6 +166,19 @@ const UspWrapper = styled.div({
   gap: theme.space.xs,
 })
 
+export const Divider = () => {
```
robinandeer, line 169: does this need to be exported?

robinandeer (approved):
Nice improvements -- just had a comment about presentational component logic

Added via Giphy

apps/store/src/features/carDealership/PriceBreakdown.tsx
gustaveen: Good point! Generally I think it makes sense to move that logic outside this component. But for now I would need to make that check in two places and this is a quite specific use case for now.

apps/store/src/features/carDealership/TrialExtensionForm.tsx
gustaveen: oops! :)

gustaveen:
### Merge activity

* **Nov 9, 3:17 AM**: [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3529) rebased this pull request after merging its parent, because this pull request is set to merge when ready.

PR #3553, opened 2023-11-13
Implement simple API endpoint to test Widget flow

robinandeer:
## Checklist before requesting a review

- [ ] I have performed a self-review of my code

PR #3555, opened 2023-11-13
GEN-1455 | Initialize Widget from flow route

robinandeer:
Co-authored by @alebedev

## Describe your changes

- Setup new route to initialize widget from a CMS "widgetFlow" story

- Use new "shopSessionCreatePartner" mutation

- Fix `isWidgetFlowStory` helper

- Define single entry point for widget initialization

- Landing pages will link to this route to initialize the widget

- This overrides the other slug-route but we will address that in the next PR

### DIFF
--- a/apps/store/src/pages/widget/flows/[...slug].tsx
+++ b/apps/store/src/pages/widget/flows/[...slug].tsx
@@ -1,0 +1,61 @@
+import { GetServerSideProps } from 'next'
+import { STORYBLOK_WIDGET_FOLDER_SLUG } from '@/features/widget/widget.constants'
+import { initializeApolloServerSide } from '@/services/apollo/client'
+import {
+  ShopSessionCreatePartnerDocument,
+  ShopSessionCreatePartnerMutation,
+  ShopSessionCreatePartnerMutationVariables,
+} from '@/services/apollo/generated'
+import { PageStory, WidgetFlowStory, getStoryBySlug } from '@/services/storyblok/storyblok'
+import { isWidgetFlowStory } from '@/services/storyblok/Storyblok.helpers'
+import { getCountryByLocale } from '@/utils/l10n/countryUtils'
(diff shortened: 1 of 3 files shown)


robinandeer:
* main
  * **PR #3553** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3553?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #3555** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3555?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3555?utm_source=stack-comment).

apps/store/src/pages/widget/flows/[...slug].tsx
```diff
@@ -0,0 +1,61 @@
...
+  const story = await getStoryBySlug<PageStory | WidgetFlowStory>(slug, {
+    version: context.draftMode ? 'draft' : undefined,
+    locale: context.locale,
+  })
+  const flow = String(story.id)
+
+  if (!isWidgetFlowStory(story)) {
+    console.warn('Story is not a widget flow story:', story.slug)
+    return { notFound: true }
+  }
```
Comment (guilhermespopolin) [inline apps/store/src/pages/widget/flows/[...slug].tsx:32]: It's configured in Storyblok that only `WidgetFlow` stories can be placed under `widget/flows` folder but I guess it better being safe and verify this here as well, right?!

image

guilhermespopolin (approved):

apps/store/src/pages/widget/flows/[...slug].tsx
Comment (robinandeer) [reply apps/store/src/pages/widget/flows/[...slug].tsx:32]: Aha yes I agree - I think we should treat it as an external system so we should validate it in code before we use it :)

PR #3589, opened 2023-11-17
CAR-41 - refact(carDealership): added car dealership landing page with extension

guilhermespopolin:
## Describe your changes

- extract shared CMS page props retrieval into a function: `getStoryblokPageProps`
- create a `car-buyer/[contractId]` landing page to be linked by CRM

Current URL for car dearlership landing pages is: `[locale]/car-buyer?:id=[contractId]` while the CRM uses `[locale]/car-buyer/[contractId]`. Ideally we want the `contractId` as a required URL segment so this PR is about doing that. However car dealership landing pages are regular CMS pages, which are currently being handled by `pages/[[...slug]]` page. So what we want is to use the same _retrieve props for CMS pages logic_ used in `pages/[[...slug]]` for `car-buyer/[contractId]/index.tsx` page while preserving the `contractId` in the URL.


guilhermespopolin:
* main
  * **PR #3589** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3589?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈
    * **PR #3590** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3590?utm_source=stack-comment-icon" target="_blank">Graphite</a> 

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3589?utm_source=stack-comment).

apps/store/src/pages/[[...slug]].tsx
```diff
@@ -84,78 +79,55 @@ export const getStaticProps: GetStaticProps<PageProps, StoryblokQueryParams> = a
...
-    breadcrumbs,
-    trustpilot,
-    hideChat: story.content.hideChat ?? false,
-  }
-
-  if (isProductStory(story)) {
-    const priceTemplate = fetchPriceTemplate(story.content.priceFormTemplateId)
-    if (priceTemplate === undefined) {
-      throw new Error(`Unknown price template: ${story.content.priceFormTemplateId}`)
+    const props = await getStoryblokPageProps({ context, slug, locale, draftMode })
```
Comment (guilhermespopolin) [inline apps/store/src/pages/[[...slug]].tsx:83]: This `getStaticProps` function will fail typescript here because `getLayoutWithMenuProps` can return the props or `null`. So whenever we call it we do the checking:

```
if (layoutWithMenuProps === null) return { notFound: true }
```

However we just want to spread those props inside `getStoryblokPageProps` which makes it difficult to check if layout with menu props was retrieved sucessfully while calling `getStoryblokPageProps`. So I thought about changing the signature of `getLayoutWithMenuProps` to throw an Error when it fails instead of `null` (next PR on the stack). Let me know what you think,

apps/store/src/pages/car-buyer/[contractId]/index.tsx
```diff
@@ -0,0 +1,26 @@
...
+type Params = {
+  contractId: string
+}
+
+export const getServerSideProps: GetServerSideProps<Props, Params> = async (context) => {
+  const { params, locale, draftMode = false } = context
+  if (!params) throw new Error('Missing params')
+  if (!isRoutingLocale(locale)) return { notFound: true }
+
+  // TODO: we should probably validate if the contractId is a valid UUID
```
Comment (alebedev) [inline apps/store/src/pages/car-buyer/[contractId]/index.tsx:17]: No need, let backend worry about it

apps/store/src/pages/car-buyer/[contractId]/index.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+  const props = await getStoryblokPageProps({ context, slug: 'car-buyer', locale, draftMode })
```
Comment (alebedev) [inline apps/store/src/pages/car-buyer/[contractId]/index.tsx:19]: Do we want a constant for it?

alebedev (approved):
Very good!

PR #3609, opened 2023-11-21
Fix pre-rendering widget landing pages

robinandeer:
## Describe your changes

- Only pre-render static pages under the "widget" folder

- Add optional prop to "getPageLinks" to filter by path prefix

- Remove unused function "getFilteredProductLinks"

- Fix bug when pre-rendering static widget landing pages

- Before it was trying to render all the static pages although it should be restricted to pages under the "widget" folder

### DIFF
--- a/apps/store/src/pages/widget/[[...slug]].tsx
+++ b/apps/store/src/pages/widget/[[...slug]].tsx
@@ -10,9 +10,9 @@ import {
   getStoryBySlug,
   type StoryblokPageProps,
   type StoryblokQueryParams,
-  getFilteredPageLinks,
   type PageStory,
   getRevalidate,
+  getPageLinks,
 } from '@/services/storyblok/storyblok'
 import { STORY_PROP_NAME } from '@/services/storyblok/Storyblok.constant'
 import { isRoutingLocale } from '@/utils/l10n/localeUtils'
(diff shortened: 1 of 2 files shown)


apps/store/src/pages/widget/[[...slug]].tsx
```diff
@@ -67,15 +67,26 @@ export const getStaticProps: GetStaticProps<PageProps, StoryblokQueryParams> = a
...
+    (context.locales ?? []).map((locale) =>
+      getPageLinks({ startsWith: `${locale}/${STORYBLOK_WIDGET_FOLDER_SLUG}` }),
+    ),
+  )
+
   return {
-    paths: pageLinks.map((item) => ({ params: { slug: item.slugParts }, locale: item.locale })),
+    paths: nestedPageLinks
+      .flat()
+      .filter((item) => !item.slugParts.includes('flows'))
```
Comment (alebedev) [inline apps/store/src/pages/widget/[[...slug]].tsx:85]: Should we filter out by content type instead?

apps/store/src/pages/widget/[[...slug]].tsx
```diff
@@ -67,15 +67,26 @@ export const getStaticProps: GetStaticProps<PageProps, StoryblokQueryParams> = a
...
+      .map((item) => ({
+        params: { slug: item.slugParts.slice(1) },
```
Comment (alebedev) [inline apps/store/src/pages/widget/[[...slug]].tsx:87]: Should we have better name for it?  Is it slug parts without locale or something else?

alebedev (approved):
Good fix, but let's think if we can follow up with something more reliable

apps/store/src/pages/widget/[[...slug]].tsx
Comment (robinandeer) [reply apps/store/src/pages/widget/[[...slug]].tsx:85]: Not possible unfortunately since we are fetching "links" instead of "stories"

But perhaps we should just fetch minimal stories instead...

apps/store/src/pages/widget/[[...slug]].tsx
Comment (robinandeer) [reply apps/store/src/pages/widget/[[...slug]].tsx:87]: It's the slug minus the "widget" part since that's already included in the Next pages folder -- makes sense?

But yes perhaps I can come up with a name for it

robinandeer:
### Merge activity

* **Nov 21, 3:58 PM**: @robinandeer merged this pull request with [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3609).

PR #3671, opened 2023-11-29
Widget | Hide close button in loading price dialog

robinandeer:
## Describe your changes

- Refactor fullscreen modal to be able to override the header

- In the fullscreen price loader dialog, hide the header

- User can't close this dialog, so the header is not needed

### DIFF
--- a/apps/store/src/features/widget/CalculatePricePage.tsx
+++ b/apps/store/src/features/widget/CalculatePricePage.tsx
@@ -114,7 +114,7 @@ export const CalculatePricePage = (props: Props) => {
       </Space>
 
       <FullscreenDialog.Root open={loading}>
-        <FullscreenDialog.Modal center={true}>
+        <FullscreenDialog.Modal center={true} Header={null}>
           <PriceLoaderWrapper>
             <PriceLoader />
           </PriceLoaderWrapper>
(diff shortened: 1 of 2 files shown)


apps/store/src/features/widget/CalculatePricePage.tsx
```diff
@@ -114,7 +114,7 @@ export const CalculatePricePage = (props: Props) => {
       </Space>
 
       <FullscreenDialog.Root open={loading}>
-        <FullscreenDialog.Modal center={true}>
+        <FullscreenDialog.Modal center={true} Header={null}>
```
gustaveen, line 117: We pass in `null` but check for `undefined`? Can't we just skip passing Header at all or will that break?

gustaveen (approved):
Nice, just a comment!

apps/store/src/features/widget/CalculatePricePage.tsx
robinandeer: The default behavior is to show the header with the close button - this is what you get if you pass `undefined`

In this case I want to override the default behavior but at the same time render nothing.I know it's a little confusing but I feel like it still makes sense.

robinandeer:
### Merge activity

* **Nov 30, 9:44 AM**: @robinandeer started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3671).

PR #3687, opened 2023-12-04
Hydrate all atoms to explicit stores

robinandeer:
## Describe your changes

- Add an explicit global Jotai store to the app

- Hydrate global data to the global store

- We have an issue where global product metadata is not available on PDP

- This is because the atom loads data from a different store (provider) than the one that is used to hydrate the global data

- With this approach, we can hydrate the global data to the global store and use it in the atom

- I didn't change anything about atoms that are never hydrated but we might see similar issues there in the future - let's discuss this!

### DIFF
--- a/apps/store/src/components/LayoutWithMenu/ProductMetadataContext.tsx
+++ b/apps/store/src/components/LayoutWithMenu/ProductMetadataContext.tsx
@@ -1,13 +1,14 @@
 import { atom, useAtomValue } from 'jotai'
 import { useHydrateAtoms } from 'jotai/utils'
+import { globalStore } from '@/utils/globalStore'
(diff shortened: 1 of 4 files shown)


guilhermespopolin (approved):

gustaveen (approved):
Makes sense to create an explicit store! 🙌 Just wondering when you wouldn't want/need to create specific stores, when you only have one store in the app or what would the use case be?

robinandeer:
> Makes sense to create an explicit store! 🙌 Just wondering when you wouldn't want/need to create specific stores, when you only have one store in the app or what would the use case be?
> 
> 

If you only rely on a single global store you can just use the default one (implied)

It's when you have multiple (especially nested) providers that you need to specify the store. However, I find the API a little awkward to work with I have to say...

PR #3692, opened 2023-12-05
GEN-1538 | Hide chat on all Widget pages

robinandeer:
## Describe your changes

- Hide customer first chat on all widget pages

- Refactor "hide chat" page prop

- Rename page prop to `_hideChat` to avoid naming collisions

- Wrap reading and "writing" of page prop in utility functions

- Remove explicit page prop typing of "hide chat"

- This logic was starting to get spread out all over the code base so I wanted to consolidate it into a single place

- Since the page prop can be added to any page, I think it is not necessary to explicitly type it

### DIFF
--- a/apps/store/src/pages/car-buyer/validation/[dealerId].tsx
+++ b/apps/store/src/pages/car-buyer/validation/[dealerId].tsx
@@ -3,6 +3,7 @@ import Head from 'next/head'
 import { type ComponentProps } from 'react'
 import { Heading } from 'ui'
 import { CarBuyerValidationPage } from '@/features/carDealership/CarBuyerValidationPage'
+import { hideChatOnPage } from '@/services/CustomerFirst'
 
 type Props = ComponentProps<typeof CarBuyerValidationPage> & { authenticated: boolean }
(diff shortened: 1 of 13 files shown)


apps/store/src/pages/car-buyer/validation/[dealerId].tsx
```diff
@@ -21,7 +22,7 @@ export const getServerSideProps: GetServerSideProps<Props, Params> = (context) =
     props: {
       authenticated: true,
       dealerId: context.params.dealerId,
-      hideChat: true,
+      ...hideChatOnPage(),
```
Comment (alebedev) [inline apps/store/src/pages/car-buyer/validation/[dealerId].tsx:25]: Did you consider inverting it?  `props.showChat` enables chat and for some cases we can pass default

Comment (robinandeer) [reply apps/store/src/pages/car-buyer/validation/[dealerId].tsx:25]: Yes, I tried a few options.

However, since we already pass it as "hideChat" from the CMS it was even harder to read the code.

It also matches the current setup of "enabled by default".

However, I think we can build on this new approach -- I do agree that setting "show chat" is more intuitive than "hide chat" 🙂

PR #3702, opened 2023-12-07
GEN-1559 | Parse student pre-fill query parameter

robinandeer:
## Describe your changes

- Parse "student" query parameter from the URL

- Deserialize data value into a string for all radio inputs

- Remove unused query parameters like city, firstName, lastName, etc.

- Update tests

- We want to be able to pre-fill the "is student" input

### DIFF
--- a/apps/store/src/features/widget/parseSearchParams.test.ts
+++ b/apps/store/src/features/widget/parseSearchParams.test.ts
@@ -18,11 +18,10 @@ describe('parsePriceIntentDataSearchParams', () => {
     expect(priceIntentData).toEqual({
       street: '123 Main St',
       zipCode: '12345',
-      city: 'New York',
       livingSpace: 100,
       numberCoInsured: 2,
     })
-    expect(remainingSearchParams.toString()).toBe('')
+    expect(remainingSearchParams.toString()).toBe('city=New+York')
   })
 
   it('should handle empty search params', () => {
@@ -51,5 +50,18 @@ describe('parsePriceIntentDataSearchParams', () => {
     // Assert
     expect(priceIntentData).toEqual({ street: '123 Main St' })
     expect(remainingSearchParams.toString()).toBe('coInsured=xyz&livingSpace=abc&unknown=xyz')
+  })
+
+  it('should handle boolean values', () => {
+    // Arrange
+    const searchParams = new URLSearchParams()
+    searchParams.set('student', 'yes')
+
+    // Act
+    const [priceIntentData, remainingSearchParams] = parsePriceIntentDataSearchParams(searchParams)
+
+    // Assert
+    expect(priceIntentData).toEqual({ isStudent: true })
+    expect(remainingSearchParams.toString()).toBe('')
   })
 })
--- a/apps/store/src/features/widget/parseSearchParams.ts
+++ b/apps/store/src/features/widget/parseSearchParams.ts
@@ -18,16 +18,12 @@ enum SearchParam {
   // Price Intent Data
   StreetAddress = 'street',
   ZipCode = 'zipCode',
-  City = 'city',
   LivingSpace = 'livingSpace',
   NumberCoInsured = 'coInsured',
+  Student = 'student',
 
   // No longer used?
   PhoneNumber = 'phoneNumber',
-  BirthDate = 'birthDate',
-  FirstName = 'firstName',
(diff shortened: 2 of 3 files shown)


robinandeer:
* main
  * **PR #3702** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3702?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3702?utm_source=stack-comment).

apps/store/src/features/widget/parseSearchParams.ts
```diff
@@ -18,16 +18,12 @@ enum SearchParam {
   // Price Intent Data
   StreetAddress = 'street',
   ZipCode = 'zipCode',
-  City = 'city',
   LivingSpace = 'livingSpace',
   NumberCoInsured = 'coInsured',
+  Student = 'student',
```
alebedev, line 23: It's called `isStudent` in the priceIntent data, should we use the same name?

apps/store/src/features/widget/parseSearchParams.test.ts
```diff
@@ -18,11 +18,10 @@ describe('parsePriceIntentDataSearchParams', () => {
     expect(priceIntentData).toEqual({
       street: '123 Main St',
       zipCode: '12345',
-      city: 'New York',
       livingSpace: 100,
       numberCoInsured: 2,
     })
-    expect(remainingSearchParams.toString()).toBe('')
+    expect(remainingSearchParams.toString()).toBe('city=New+York')
```
gustaveen, line 24: Just curious, do we use the city param for something?

apps/store/src/features/widget/parseSearchParams.ts
```diff
@@ -123,6 +123,17 @@ const parseNumber = (value: string): number | undefined => {
...
+  switch (value) {
+    case 'yes':
+      return true
+    case 'no':
+      return false
+    default:
+      return undefined
+  }
+}
+
```
gustaveen, line 136: Do we use `yes` for other boolean query params as well? Just so we don't mix with setting `1` or `0` or maybe this is the only one?

alebedev (approved):
Looks good, 2 questions inside

gustaveen (approved):
Nice!

apps/store/src/features/widget/parseSearchParams.ts
robinandeer: true but this is what the API looks like from before

we can deprecate this query param and also support "isStudent" perhaps?

apps/store/src/features/widget/parseSearchParams.test.ts
robinandeer: we have used in the past but we no longer care about it

so this is just an example of a query param that is sent but that we don't do anything with - it could be anything we don't know about like gtm parameters etc.

apps/store/src/features/widget/parseSearchParams.ts
robinandeer: we don't have any other boolean params in the widget that i am aware of

i am just trying to match the current api so we don't complicate things for the partners

however, i agree that it would be great to migrate over to 1/0 for example

apps/store/src/features/widget/parseSearchParams.test.ts
```diff
@@ -52,4 +51,17 @@ describe('parsePriceIntentDataSearchParams', () => {
     expect(priceIntentData).toEqual({ street: '123 Main St' })
     expect(remainingSearchParams.toString()).toBe('coInsured=xyz&livingSpace=abc&unknown=xyz')
   })
+
+  it('should handle boolean values', () => {
+    // Arrange
+    const searchParams = new URLSearchParams()
+    searchParams.set('student', 'yes')
```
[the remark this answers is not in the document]
alebedev: That's probably enough

apps/store/src/features/widget/parseSearchParams.ts
alebedev: As long as you send `isStudent` to storefront, it should be fine.  If we have it documented, let's plan slow deprecation after we're live with new widget

PR #3705, opened 2023-12-07
GEN-1541 | Show partner specific product data in widget flow

robinandeer:
## Describe your changes

- Pass partner name from widget flow story when fetching product data

- We want to display the partner specific product data in the widget (for AVY)

### DIFF
--- a/apps/store/src/components/ProductData/ProductData.graphql
+++ b/apps/store/src/components/ProductData/ProductData.graphql
@@ -1,5 +1,5 @@
-query ProductData($productName: String!) {
-  product(productName: $productName) {
+query ProductData($productName: String!, $partnerName: String) {
+  product(productName: $productName, partnerName: $partnerName) {
     id
     name
     displayNameFull
@@ -30,5 +30,3 @@ query ProductData($productName: String!) {
     }
   }
 }
-
-
--- a/apps/store/src/components/ProductData/fetchProductData.ts
+++ b/apps/store/src/components/ProductData/fetchProductData.ts
@@ -6,19 +6,21 @@ import {
 } from '@/services/apollo/generated'
 import { ProductData } from './ProductData.types'
 
-type Params = {
+type Params = ProductDataQueryVariables & {
   apolloClient: ApolloClient<unknown>
-  productName: string
 }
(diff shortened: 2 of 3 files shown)


robinandeer:
* main
  * **PR #3705** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3705?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3705?utm_source=stack-comment).

apps/store/src/components/ProductData/ProductData.graphql
```diff
@@ -1,5 +1,5 @@
-query ProductData($productName: String!) {
-  product(productName: $productName) {
```
guilhermespopolin, line 2: Just curious:
Since adding more fields to the query is not part of the changes, which one(s) would take `partnerName` into account? `documents`? 🤔

guilhermespopolin (approved):
Looks correct :)
Just left a small question

apps/store/src/components/ProductData/ProductData.graphql
robinandeer: I think the way to think about it is that "partnerName" is like a filter for product variants

we have specific product variants for AVY so anything that we fetch on product variant will be affected

guilhermespopolin: Aha, got it. In that case should we use "field arguments"? Something like this?

```graphql
query ProductData($productName: String!, $partnerName: String) {
  product(productName: $productName, partnerName: $partnerName) {
    ...
    variants(partnerName: $partnerName) {
      typeOfContract
      displayName
      displayNameSubtype
      perils {
        ...Peril
      }
      ...
  }
}
```

robinandeer: ahaaa i understand -- that might be more correct right now

but this gives us the option to resolve other fields in a partner-specific way for the product

PR #3714, opened 2023-12-11
Parse API date without converting between time zones

robinandeer:
## Describe your changes

- Use date-fns to convert between Date and API date string

- We noticed issues earlier where someone would pick a day in the date picker only to see the previous day selected.

- Previously we converted using `toISOString` which means that JS would convert the date to UTC and then convert it to a string. This is not what we want, we want to convert the date to a string in the local timezone.

### DIFF
--- a/apps/store/src/utils/date.ts
+++ b/apps/store/src/utils/date.ts
@@ -1,3 +1,5 @@
+import { format } from 'date-fns'
+
 export const convertToDate = (value: unknown): Date | null => {
   if (typeof value === 'string') {
     const date = new Date(value)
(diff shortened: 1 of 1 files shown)


robinandeer:
* main
  * **PR #3714** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3714?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3714?utm_source=stack-comment).

gustaveen (approved):
👏

PR #3717, opened 2023-12-11
GEN-1528 / Display main offer title and pillow on QuickAdd component

gustaveen:
## Describe your changes
- Add new size for Pillow component
- Add prop for passing main offer pillow and display multiple pillows

Screenshot 2023-12-11 at 14 13 57

Update accident x-sell project

### DIFF
--- a/apps/store/src/components/QuickAdd/QuickAdd.tsx
+++ b/apps/store/src/components/QuickAdd/QuickAdd.tsx
@@ -11,6 +11,7 @@ type Props = {
   title: string
   subtitle: string
   pillow: ComponentProps<typeof Pillow>
+  mainOfferPillow?: ComponentProps<typeof Pillow>
   badge?: ComponentProps<typeof Badge>
   href: string
   price?: ComponentProps<typeof Price>
@@ -24,7 +25,14 @@ export const QuickAdd = (props: Props) => {
     <Card>
       <Space y={1}>
         <SpaceFlex space={1} align="center">
-          <Pillow size="small" {...props.pillow} />
+          {props.mainOfferPillow ? (
+            <PillowWrapper>
+              <Pillow size="mini" {...props.mainOfferPillow} />
+              <Pillow size="mini" {...props.pillow} />
+            </PillowWrapper>
+          ) : (
+            <Pillow size="small" {...props.pillow} />
+          )}
           <div>
             <StyledLink href={props.href}>
               <Text as="span" color="textTranslucentPrimary">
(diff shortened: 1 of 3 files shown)


apps/store/src/components/QuickAdd/QuickAdd.tsx
```diff
@@ -80,6 +88,16 @@ const StyledLink = styled(Link)({
...
 })
 
+const PillowWrapper = styled.div({
+  display: 'flex',
+  width: '3rem',
+  height: '3rem',
+
+  '& > *:last-of-type': {
+    marginTop: 'auto',
+  },
```
robinandeer, line 98: what does this rule do?

apps/store/src/components/QuickAdd/QuickAdd.tsx
```diff
@@ -11,6 +11,7 @@ type Props = {
   title: string
   subtitle: string
   pillow: ComponentProps<typeof Pillow>
+  mainOfferPillow?: ComponentProps<typeof Pillow>
```
robinandeer, line 14: another idea would be to pass in a list of pillows - just a thought

robinandeer (approved):
Added via Giphy

apps/store/src/components/QuickAdd/QuickAdd.tsx
gustaveen: Indeed @robinandeer  👍, I started with that solution but felt I wanted it more restrictive so you don't pass in more than two pillows and the order of the pillows also matter. But I'm up for either if we feel like it makes more sense with a list 🙌

apps/store/src/components/QuickAdd/QuickAdd.tsx
gustaveen: It places the second pillow at the bottom. `margin-top|left: auto` is my go to for placing a single item at the other side of a container :) But I realise I can simplify it, with `margin-left: auto` we don't even need `flex` 🥳

apps/store/src/components/QuickAdd/QuickAdd.tsx
guilhermespopolin: Another nit/suggestion thing about it if I may: should we also rename `pillow` to something like `additionalOfferPillow`? My concern is being difficult to figure out what prop to use when using `QuickAdd` with only one offer without having to look at `QuickAdd` implementation. No that I think about it, since `pillow` is required while `mainOfferPillow`, shouldn't `pillow` contain the pillow for the main offer? I mean, we always have a "main offer" (required) while we my have an additional offer...Does that make sense? 🤔

guilhermespopolin (approved):
Well done 👏🏻 
Just left a small comment.

gustaveen:
### Merge activity

* **Dec 14, 4:58 AM**: @@gustaveen started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3717).

PR #3718, opened 2023-12-11
Don't allow picking a start date in the past

robinandeer:
## Describe your changes

Set a minimim start of today in the start date input.

I feel like we discussed this before but it seems very reasonable


gustaveen (approved):
Ah, good catch! 👌

PR #3719, opened 2023-12-11
GEN-968 | Add custom date input

robinandeer:
## Describe your changes

- Implement a custom date picker

- Update API for `InputDate` to use a change callback that returns a Date object

- Using the native date picker is problematic, especially on desktop where the behavior is hard to customize and sometimes doesn't work as intended.

- This is just an experiment, let's see if we want to merge it

- I was considering using the native picker on iOS but I think this might be easier to maintain

### DIFF
--- a/apps/store/src/components/InputDay/InputDay.tsx
+++ b/apps/store/src/components/InputDay/InputDay.tsx
@@ -1,0 +1,227 @@
+import { keyframes } from '@emotion/react'
+import styled from '@emotion/styled'
+import * as Popover from '@radix-ui/react-popover'
+import { sv } from 'date-fns/locale'
+import { useTranslation } from 'next-i18next'
+import { useId, useState } from 'react'
+import { useInput, DayPicker, type SelectSingleEventHandler } from 'react-day-picker'
+import { ChevronIcon, theme, Text, LockIcon } from 'ui'
+import { convertToDate } from '@/utils/date'
+import { Language } from '@/utils/l10n/types'
+import { useCurrentLocale } from '@/utils/l10n/useCurrentLocale'
+import { useFormatter } from '@/utils/useFormatter'
+import { useHighlightAnimation } from '@/utils/useHighlightAnimation'
+
+import 'react-day-picker/dist/style.css'
+
+type Props = {
+  label: string
+
+  id?: string
+  name?: string
+  selected?: Date
+  defaultSelected?: Date
+  onSelect?: (day: Date) => void
+  fromDate?: Date
+  toDate?: Date
+  disabled?: boolean
+  backgroundColor?: 'default' | 'light'
+  required?: boolean
+}
+
+export const InputDay = (props: Props) => {
+  const { t } = useTranslation()
+  const formatter = useFormatter()
+  const { language } = useCurrentLocale()
+  const { inputProps, dayPickerProps } = useInput({
+    locale: language === Language.Sv ? sv : undefined,
+    format: 'yyyy-MM-dd',
+    required: props.required,
+    defaultSelected: props.selected ?? props.defaultSelected,
+    fromDate: props.fromDate,
+    toDate: props.toDate,
+  })
+
+  const [open, setOpen] = useState(false)
+
+  const autoIdentifier = useId()
+  const inputId = props.id ?? autoIdentifier
+
+  const backgroundColor =
+    props.backgroundColor === 'light'
+      ? theme.colors.backgroundFrostedGlass
+      : theme.colors.translucent1
+
+  const { highlight, animationProps } = useHighlightAnimation<HTMLDivElement>({
+    defaultColor: backgroundColor,
+  })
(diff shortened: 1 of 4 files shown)


robinandeer:
* main
  * **PR #3714** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3714?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #3718** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3718?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
      * **PR #3719** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3719?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3719?utm_source=stack-comment).

alebedev (commented):
We probably want some visual tweaks before it's merged, right?
For example
- isn't shadow under popover too small?
- should month name be capitalized?
- do we need/want year selector?  Eg. birth date for 10-year old dog would not be easy to enter right now

alebedev:
Looks like there's built-in solution for year picker - using dropdowns for month and year

https://react-day-picker.js.org/basics/navigation#choosing-a-caption-layout

apps/store/src/components/InputDate/InputDate.tsx
```diff
@@ -64,30 +60,80 @@ export const InputDate = (props: Props) => {
...
+              <LockIcon size="1rem" color={theme.colors.textSecondary} />
+            ) : (
+              <StyledChevronIcon size="1rem" />
+            )}
+          </InnerWrapper>
+        </Wrapper>
+      </Popover.Trigger>
+
+      <Popover.Portal>
+        <PopoverContent side="bottom" align="start" sideOffset={-4} alignOffset={-4}>
```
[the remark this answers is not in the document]
robinandeer: ahh I can see if it is supported to use rems

apps/store/src/components/InputDate/InputDate.tsx
```diff
@@ -64,30 +60,80 @@ export const InputDate = (props: Props) => {
...
+          <StyledDayPicker
+            id={inputId}
+            locale={language === Language.Sv ? sv : undefined}
```
gustaveen, line 88: Is default language en?

gustaveen (commented):
Great, feels like solid date picker! 🙌

So let's just validate the year/month dropdown works as mentioned. Styling wise it looks like it's flexible for making it fit with our design 👌 Does it have support for a clear or/and today button?

gustaveen:
nit: saw that the story https://62ea2a095003502af2e247b8-rollyjuewp.chromatic.com/?path=/story/inputs-date--default don't update the date when picking date, should we add a handler to reflect changes?

guilhermespopolin (commented):
Great work
I did some testing on Android and it works fine :)
The only reaming issue is that behaviour during date selection that comes from the timezone difference: The actual selected date is always 2 days before the one I have clicked as one can see on the video below:

https://github.com/HedvigInsurance/racoon/assets/19200662/0243c95d-80e6-46ef-a833-9ed503fe1210

But that is not something introduced with the custom date picker. It was also happening with the native one.

apps/store/src/components/InputDate/InputDate.tsx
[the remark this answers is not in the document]
robinandeer: not possible...

apps/store/src/components/InputDate/InputDate.tsx
robinandeer: yes it is!

apps/store/src/components/InputDay/InputDay.tsx
```diff
@@ -0,0 +1,227 @@
...
+    animationName: slideUpAndFadeAnimation,
+  },
+
+  '&[data-side="bottom"]': {
+    animationName: slideDownAndFadeAnimation,
+  },
+})
+
+const StyledDayPicker = styled(DayPicker)({
+  // Docs: https://react-day-picker.js.org/basics/styling
```
robinandeer, line 150: I tried to use `css().name` in combination with `modifierClassNames` which does add a generated class name to the element in the calendar

however, the styles don't get included on the page still

not sure where the original problem is

still we will have to override some styles using this approach so better to keep it all here

alebedev (approved):

gustaveen (approved):
Wohoo, great work!

robinandeer:
### Merge activity

* **Dec 13, 11:30 AM**: @@robinandeer started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3719).

PR #3808, opened 2024-01-04
Bug: send correct currency code to Adyen

robinandeer:
## Describe your changes

- Store currency code by country

- Send correct currency code to Adyen

- For some reason we sent locale and this caused errors

### DIFF
--- a/apps/store/src/services/adyen/useGetAdyenCheckout.ts
+++ b/apps/store/src/services/adyen/useGetAdyenCheckout.ts
@@ -12,7 +12,7 @@ import { useAdyenTranslations } from './useAdyenTranslations'
 import { useTokenizePaymentDetails } from './useTokenizePaymentDetails'
 
 export const useGetAdyenCheckout = (): ((options: CoreOptions) => AdyenCheckoutAPI) => {
-  const { countryCode } = useCurrentCountry()
+  const { countryCode, currencyCode } = useCurrentCountry()
   const { payButton, locale } = useAdyenTranslations()
   const handleSubmit = useHandleSubmit()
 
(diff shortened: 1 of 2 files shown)


robinandeer:
* `main`
  * **PR #3808** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3808?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3808?utm_source=stack-comment).

apps/store/src/services/adyen/useGetAdyenCheckout.ts
```diff
@@ -31,11 +31,11 @@ export const useGetAdyenCheckout = (): ((options: CoreOptions) => AdyenCheckoutA
           'applepay',
         ],
 
-        paymentMethodsConfiguration: getPaymentMethodsConfiguration(locale, countryCode),
+        paymentMethodsConfiguration: getPaymentMethodsConfiguration(currencyCode, countryCode),
```
guilhermespopolin, line 34: Nice catch. I guess that wouldn't happen if we had a object/enum with the possible/supported currency right? Do you think adding that is worth it?

guilhermespopolin (approved):
Nice catch

apps/store/src/services/adyen/useGetAdyenCheckout.ts
robinandeer: nah not really, adyen is being removed in a few months also

PR #3837, opened 2024-01-15
GEN-1696 - styles: custom styles for scrollbar. Make appear roughly the same on all browsers.

guilhermespopolin:
## Describe your changes

Customise scrollbar so they appear roughly the same in all browsers.
Note: CSS specification includes two properties that can be used for that purpose: `scrollbar-width` and `scrollbar-color` but at the time only _Firefox_ implements it. That's why we need declarations for `-webkit` based browsers like _Chrome_ and _Safari_. 

**Chrome**
https://github.com/HedvigInsurance/racoon/assets/19200662/24e4df7e-42ed-4e70-8afe-6b1e5d73c14e

**Safari**
https://github.com/HedvigInsurance/racoon/assets/19200662/69ca0cc0-9cbd-4e24-9d3b-d473668faab4

**Firefox**
https://github.com/HedvigInsurance/racoon/assets/19200662/fd7b75ec-4948-4877-ae9d-2ce42e04ade9

Please, give it a try :)

We wanted to move away from how PC scrollbar looks like, while customising into something that blends in more naturally in our app.

### DIFF
--- a/apps/store/src/components/ProductReviews/ReviewsDialog.tsx
+++ b/apps/store/src/components/ProductReviews/ReviewsDialog.tsx
@@ -116,13 +116,32 @@ const CloseButton = styled.button({
 })
 
 const DialogWindow = styled(Dialog.Window)({
+  '--scrollbar-thumb-color': theme.colors.gray500,
+  '--scrollbar-thumb-color-hover': theme.colors.gray600,
+  '--scrollbar-track-width': '8px',
+
   paddingInline: theme.space.md,
   paddingBlock: theme.space.xxxl,
   overflowY: 'auto',
-  // Hide scrollbar in Dialog
-  scrollbarWidth: 'none',
-  '&::-webkit-scrollbar': {
-    display: 'none',
+
+  // Scrollbar styles
+  [mq.sm]: {
+    // Firefox
+    scrollbarColor: 'var(--scrollbar-thumb-color)',
+    scrollbarWidth: 'thin',
+
+    // Webkit based browsers
+    '::-webkit-scrollbar': {
+      width: 'var(--scrollbar-track-width)',
+      height: 'var(--scrollbar-track-width)',
+    },
+    '::-webkit-scrollbar-thumb': {
+      borderRadius: '1000px',
+      backgroundColor: 'var(--scrollbar-thumb-color)',
+    },
+    '::-webkit-scrollbar-thumb:hover': {
+      backgroundColor: 'var(--scrollbar-thumb-color-hover)',
+    },
   },
 
   [mq.md]: {


guilhermespopolin:
* `main`
  * **PR #3837** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3837?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3837?utm_source=stack-comment).

gustaveen (commented):
Nice, looks good! 🙌
Had some few considerations though:
- The scrollbar is always visible in desktop. Is it possible to make it disappear when you stop scrolling like the native behaviour? It seems to be working in iPhone though on safari.
- Would be nice to have some space in desktop at least to the right of the scrollbar like the native one.

alebedev:
Did you consider only changing dialog scrollbar?  There having something small instead of nothing is clearly better.  On the other hand, scrollbar styling is extra tricky to get right and maybe we should keep default scrollbar for main content?

gustaveen:
> Did you consider only changing dialog scrollbar? There having something small instead of nothing is clearly better. On the other hand, scrollbar styling is extra tricky to get right and maybe we should keep default scrollbar for main content?

Yeah I think that could be a good way forward to only use it in the dialog if it's hard to make it feel more native on the rest of the site 👍

guilhermespopolin:
> > Did you consider only changing dialog scrollbar? There having something small instead of nothing is clearly better. On the other hand, scrollbar styling is extra tricky to get right and maybe we should keep default scrollbar for main content?
> 
> Yeah I think that could be a good way forward to only use it in the dialog if it's hard to make it feel more native on the rest of the site 👍

Yes, we can do that. After some more investigation, it seems as soon as you style scrollbar browsers/OS default behaviour for it will be overiden. That's why I can't "respect" MacOS setting for it after adding custom styles - meaning it will be always visible despite your config. That only happens for webkit based browsers though. It works expected in Firefox.

More research has show that applying custom styles for scrollbar in mobile might not be a good idea as you'd doing more harm than good as scrollbar mobile styles are fairly inobtrusive. So here's what I'm gonna do:

* Apply custom styles only for the Dialog
* Make sure it doesn't get applicable on mobile devices.

gustaveen (approved):
Good job investigating this! 👌

guilhermespopolin:
### Merge activity

* **Jan 17, 8:38 AM**: @guilhermespopolin started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3837).

PR #3838, opened 2024-01-15
Update storyblok/js and storyblok/react to consistent versions

alebedev:
## Describe your changes

Update @storyblok/js and @storyblok/react to versions that work with each other.  Renovate tried to do it separately and failed due to type errors from conflicting versions


gustaveen (approved):
Good catch! 👌

PR #3858, opened 2024-01-19
fix(hydration-errors): make sure to import useTranslation from next-i18next

guilhermespopolin:
## Describe your changes

* Fix hydration errors on _Checkout_ page by changing from where `RemoveActionButton` gets `useTranslation` hook: `next-i18next` instead of `react-i18next`.

https://github.com/HedvigInsurance/racoon/assets/19200662/24f3ec28-ef2c-48d6-a445-e71b5938487c

Also took the liberty of updating it in other places of the codebase as well.
I could say (at lest personally) the major reason why that mistake insists on existing in your codebase is because intelisense only suggest auto importing `useTranslation` from `react-i18next`.  As a consequence I've added a eslint rule to help us to remember.

image

I didn't fully understand the issue but here's the context where it happens:

* `i18n` was not being initialised for `react-i18next` during build

In any case we should be using `useTranslation` hook from `next-i18next` so...

That hypothesis can be verified by adding the following to `RemoveActionButton`:

```
  const { t, ready } = useTranslation('cart')
  console.log('ready', ready)
  console.log('t', t('REMOVE_ENTRY_BUTTON'))
```

That would print `false` and `REMOVE_ENTRY_BUTTON` on the server but `true` and `Ta bort` on the client.


guilhermespopolin:
* `main`
  * **PR #3858** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3858?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3858?utm_source=stack-comment).

alebedev (approved):
Nice!  If this helps, can you looks into adding eslint rule forbidding direct imports from react-i18next?

gustaveen (approved):
Great find @guilhermespopolin 🥳
This is annoying yes, the auto import in VScode chooses `react-i18next` be default so a eslint rule would be nice 🙌

guilhermespopolin:
> Nice! If this helps, can you looks into adding eslint rule forbidding direct imports from react-i18next?

Thanks for the suggestion. I've updated it by including a eslint config.

PR #3861, opened 2024-01-22
GEN-1715 - fix(ProductCardBlock): hydration errors on landing page (se-en)

guilhermespopolin:
## Describe your changes

* Make sure we hydrate   _product metadata atom_ during build/SSR, avoiding wrong HTML being generated as a result of the wrong data being stored in _product metadata atom_.

**Context**

I've noticed some hydration errors involving `ProductCardBlock` and the English version of the landing page (`se-en`).

https://github.com/HedvigInsurance/racoon/assets/19200662/cbb9ebb9-016a-4c3d-9c1d-e21381aed5fc

The reason we get hydration errors here is because, on the server, we end up entering in an `if` statement that avoids rendering `ProductCardBlock` while linking into a story that doesn't belong to a product or a product category.

image

**The reason why that happens is because the `productMetada` used for `getLinkType` function holds the swedish product metadata, despite the fact we're loading the english version.** As a result, we get the wrong HTML:

Screenshot 2024-01-22 at 17 36 44

What's interesting tho is that the correct data is being sent from the server to the client:

Screenshot 2024-01-22 at 17 40 38

Which results in a different page while hydrating the app and then causes the warnings 😢 

We've added the `globalStore` [here](https://github.com/HedvigInsurance/racoon/pull/3687) as a fix for an issue where product metadata was not accessible on product pages. The side effect tho is that the same `globalStore` is used for all locales. By default, atoms are hydrated only once per store so if, during the building process, swedish pages are generated first, the swedish version of product metadata will be used while generating pages in english.

**The solution**

I could think on 2 solutions:

1. Having a global store per routing locale. This one is more explicit but evolves having a bunch stores. Doesn't look like future proof while thinking on adding new locales.
2. Hydrating product metadata atom more then once while on the server; Doesn't include multiple stores but it's more implicit.

What do you guys think?

### DIFF
--- a/apps/store/src/components/LayoutWithMenu/ProductMetadataContext.tsx
+++ b/apps/store/src/components/LayoutWithMenu/ProductMetadataContext.tsx
@@ -1,14 +1,25 @@
 import { atom, useAtomValue } from 'jotai'
-import { useHydrateAtoms } from 'jotai/utils'
+import { atomFamily, useHydrateAtoms } from 'jotai/utils'
 import { globalStore } from '@/utils/globalStore'
-import { GlobalProductMetadata } from './fetchProductMetadata'
+import { type RoutingLocale } from '@/utils/l10n/types'
+import { useCurrentLocale } from '@/utils/l10n/useCurrentLocale'
+import { type GlobalProductMetadata } from './fetchProductMetadata'
 
-const productsAtom = atom<GlobalProductMetadata | null>(null)
+// We don't actually neeed _routineLocale for the atom creation. It will be used
+// to create a new atom for each locale by default.
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+const productsMetadataAtom = atomFamily((_routingLocale: RoutingLocale) =>
+  atom<GlobalProductMetadata | null>(null),
+)
 
 export const useProductMetadata = () => {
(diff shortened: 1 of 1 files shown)


guilhermespopolin:
* `main`
  * **PR #3861** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3861?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3861?utm_source=stack-comment).

alebedev (approved):
Great investigation and good temporary solution.  Up to you if you want to merge this first or investigate long-term solution before doing that

gustaveen (approved):
Cudos for taking the time to get to the bottom of this bug! 🥳
As discussed on stand up, let's merge this and see if we can find a long term solution 🙌

guilhermespopolin:
### Merge activity

* **Jan 23, 7:59 AM**: @guilhermespopolin started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3861).

PR #3862, opened 2024-01-23
GEN-1692 / Go back to previous step to edit reg nr when clicking edit in warning dialog

gustaveen:
## Describe your changes
When clicking "Edit" button, the logic is that you will go back to the previous step in the price form where you can edit reg number.

It made more sense to go back rather then block continuing to the next step since the happy path is smoother that way when you click "I understand".

https://github.com/HedvigInsurance/racoon/assets/6661511/e58b7f88-f70a-4f7b-bbdb-1197debc927d

## Justify why they are needed
User shoulkd be able to go back and edit reg number

## Checklist before requesting a review

- [ ] I have performed a self-review of my code

PR #3851, opened 2024-01-16
GEN-1690 / Add WarningPromt to be used inside PurchaseForm

gustaveen:
## Describe your changes
- Add new WarningPrompt that we can use in PurchaseForm to display warnings from a PriceIntent

Screenshot 2024-01-16 at 15 46 47

## Justify why they are needed
To be able to display decomisioned/not owner of car warning 

## Checklist before requesting a review

- [ ] I have performed a self-review of my code

PR #3866, opened 2024-01-23
GEN-1692 / Use atom to toggle PriceIntentWarning state

gustaveen:
## Describe your changes
Use jotai atom to control when to show PriceIntentWarning

Make sure we show potential warning after user go back and edit reg nr

### DIFF
--- a/apps/store/src/components/PriceCalculator/PriceCalculator.tsx
+++ b/apps/store/src/components/PriceCalculator/PriceCalculator.tsx
@@ -1,4 +1,5 @@
 import { datadogLogs } from '@datadog/browser-logs'
+import { useSetAtom } from 'jotai'
 import { useMemo, useState } from 'react'
 import {
   prefillData,
@@ -16,6 +17,7 @@ import { PriceCalculatorAccordion } from './PriceCalculatorAccordion'
 import { PriceCalculatorSection } from './PriceCalculatorSection'
 import { useShowFetchInsurance } from './useFetchInsurance'
 import { useHandleSubmitPriceCalculator } from './useHandleSubmitPriceCalculator'
+import { showPriceIntentWarningAtom } from './Warning/showPriceIntentWarningAtom'
 import { Warning } from './Warning/Warning'
 
 type Props = {
@@ -48,6 +50,7 @@ export const PriceCalculator = (props: Props) => {
   })
 
   const showFetchInsurance = useShowFetchInsurance({ priceIntentId: priceIntent.id })
+  const showPriceIntentWarning = useSetAtom(showPriceIntentWarningAtom)
   const [handleSubmit, handleSubmitSection, isLoading] = useHandleSubmitPriceCalculator({
     shopSession,
     priceIntent,
@@ -61,6 +64,7 @@ export const PriceCalculator = (props: Props) => {
         onConfirm()
       } else {
         if (priceIntent.externalInsurer) showFetchInsurance()
+        if (priceIntent.warning) showPriceIntentWarning(true)
 
         setActiveSectionId((prevSectionId) => {
           const currentSectionIndex = form.sections.findIndex(({ id }) => id === prevSectionId)
--- a/apps/store/src/components/PriceCalculator/Warning/Warning.tsx
+++ b/apps/store/src/components/PriceCalculator/Warning/Warning.tsx
@@ -1,8 +1,9 @@
 import styled from '@emotion/styled'
-import { useState } from 'react'
+import { useAtom } from 'jotai'
 import { Dialog, theme } from 'ui'
 import { PriceIntentWarning } from '@/services/graphql/generated'
 import { Features } from '@/utils/Features'
+import { showPriceIntentWarningAtom } from './showPriceIntentWarningAtom'
 import { WarningPrompt } from './WarningPrompt/WarningPrompt'
 
 type Props = {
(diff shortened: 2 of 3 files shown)


gustaveen:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3866?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* `main`
  * **PR #3851** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3851?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #3855** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3855?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
      * **PR #3862** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3862?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
        * **PR #3866** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3866?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3866?utm_source=stack-comment).

apps/store/src/components/PriceCalculator/PriceCalculator.tsx
```diff
@@ -48,6 +49,7 @@ export const PriceCalculator = (props: Props) => {
   })
 
   const showFetchInsurance = useShowFetchInsurance({ priceIntentId: priceIntent.id })
+  const [, setShowPriceIntentWarning] = useShowPriceIntentWarning()
```
alebedev, line 52: As an alternative, you can export atom and then you'll have `useAtomValue / useSetAtom` to selectively work with only read or only write.  Recent example here in Hope: https://app.graphite.dev/github/pr/HedvigInsurance/hedgehog/1205/Jotai-atom-demo-convert-non-task-chat-visibility-into-global-atom-backed-by-localStorage

Up to you if this is makes sense here

apps/store/src/components/PriceCalculator/PriceCalculator.tsx
```diff
@@ -61,7 +63,7 @@ export const PriceCalculator = (props: Props) => {
         onConfirm()
       } else {
         if (priceIntent.externalInsurer) showFetchInsurance()
-
+        if (priceIntent.warning) setShowPriceIntentWarning(true)
```
alebedev, line 66: Maybe we want to do it before Insurely (above)?

apps/store/src/components/PriceCalculator/useShowPriceIntentWarning.ts
```diff
@@ -0,0 +1,7 @@
+import { atom, useAtom } from 'jotai'
+
+const SHOW_PRICE_INTENT_WARNING_ATOM = atom(true)
```
alebedev, line 3: Nit: what do you think about naming for atoms?  CONSTANT_CASE here vs camelCase as if it were normal value

alebedev (approved):

apps/store/src/components/PriceCalculator/PriceCalculator.tsx
gustaveen: I wonder if it would actually stack on top of the Insurely Dialog since it would be appended after? For current situation we don't show Insurely if there's a warning since you're either not the owner of the car or it's decommissioned. But could be good to test if we add other warnings that could be rendered when you're eligible for Insurely

apps/store/src/components/PriceCalculator/useShowPriceIntentWarning.ts
gustaveen: I like camelCase better 👍 Saw now that we use both, will update this

apps/store/src/components/PriceCalculator/PriceCalculator.tsx
gustaveen: Ah, that makes sense! I created the hook thinking I needed to handle more logic than just updating the atom value so makes sense to export the atom instead 👍

gustaveen:
### Merge activity

* **Jan 24, 4:41 AM**: [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3866) rebased this pull request after merging its parent, because this pull request is set to merge when ready.

apps/store/src/components/PriceCalculator/PriceCalculator.tsx
```diff
@@ -61,6 +64,7 @@ export const PriceCalculator = (props: Props) => {
         onConfirm()
       } else {
         if (priceIntent.externalInsurer) showFetchInsurance()
+        if (priceIntent.warning) showPriceIntentWarning(true)
```
guilhermespopolin, line 67: Could we return early if that condition is met, avoiding changing `activeSectionId`? In that way I think we could make things simpler as we wouldn't need a `goBackToPreviousSection` callback.

guilhermespopolin (commented):
Well done :)
Just left a comment.

PR #2330, opened 2023-05-10
Fix duplicated strucutred data

gustaveen:
## Describe your changes
Avoid structured data script being rendered twice in <head>. There's seem to be a bug with next/head since it should [not render duplicates if you add a key prop](https://nextjs.org/docs/pages/api-reference/components/head#:~:text=To%20avoid%20duplicate,the%20following%20example%3A)

Seems to be an hydration issue so this makes sure it's only rendered once, validated with [Rich Result Test](https://search.google.com/test/rich-results/result?id=NtRRH13hlxCG3pPyGxsCFg)

Screenshot 2023-05-10 at 17 43 43
 

## Justify why they are needed
Warning from Google Search Console that content is duplicated

Screenshot 2023-05-10 at 17 47 23

## Checklist before requesting a review

- [ ] I have performed a self-review of my code

PR #3875, opened 2024-01-26
GEN-1750 - fix: added organization structured data for landing pages only

guilhermespopolin:
## Describe your changes

* Avoid structured data being added twice to the same document.

Same fix was used [here](https://github.com/HedvigInsurance/racoon/pull/2330)

Tried to abstract it into a shred component but it doesn't work as `next/head` has some restrictions about what can be used as its children.

Also spent some time trying to figure it out why is that happening but could not find any explanation. It works fine locally but it's possible to reproduce on staging.


guilhermespopolin:
* `main`
  * **PR #3875** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3875?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3875?utm_source=stack-comment).

gustaveen (approved):
Great! 🙌 Let's ping Peter to validate

PR #3879, opened 2024-01-30
Vanilla-extract support

alebedev:
## Describe your changes

- Support vanilla-extract in Next config and storybook
- Demo of using design tokens and responsive styles
- Example of extending vanilla-based component with emotion

First steps towards new CSS-in-JS solution


packages/ui/src/components/Dialog/Dialog.tsx
```diff
@@ -55,34 +16,39 @@ export const Content = (props: ContentProps) => {
...
 export const Trigger = DialogPrimitive.Trigger
 export const Close = DialogPrimitive.Close
+
+// Example of exposing base component for further styling by either vanilla-extract or emotion
+export const Window = ({
+  className,
+  ...forwardedProps
+}: { className?: string } & PropsWithChildren) => (
+  <div className={clsx(dialogWindow, className)} {...forwardedProps} />
+)
```
guilhermespopolin, line 54: 👍🏻

packages/ui/src/components/Dialog/Dialog.css.ts
```diff
@@ -0,0 +1,51 @@
...
+      [animationAllowed]: {
+        animation: `${overlayShow} 500ms cubic-bezier(0.16, 1, 0.3, 1) forwards`,
+      },
+    },
+  },
+  frosted: {
+    backgroundColor: theme.colors.backgroundFrostedGlass,
+    backdropFilter: 'blur(64px)',
+  },
+})
```
guilhermespopolin, line 24: [Discussion]
What do you think about the other notation where we add `@media` rules at the same level of all styles? Like we do for emotion styled components today. Should we set a common preference?

```
// theme
export const animationAllowed = '@media (prefers-reduced-motion: no-preference)'
```

```suggestion
export const overlay = styleVariants({
  base: {
    backgroundColor: 'rgba(0, 0, 0, 0.15)',
    position: 'fixed',
    inset: 0,
    [animationAllowed]: {
      animation: `${overlayShow} 500ms cubic-bezier(0.16, 1, 0.3, 1) forwards`,
    },
  },
  frosted: {
    backgroundColor: theme.colors.backgroundFrostedGlass,
    backdropFilter: 'blur(64px)',
  },
})
```

apps/store/src/components/ProductCard/ProductCard.tsx
```diff
@@ -4,7 +4,7 @@ import Link from 'next/link'
 import { useRouter } from 'next/router'
 import { useTranslation } from 'next-i18next'
 import { useState } from 'react'
-import { Button, Space, theme, mq } from 'ui'
```
guilhermespopolin, line 7: I guess I didn't follow this change? 🤔

guilhermespopolin (approved):
Nice work with this 💪🏻 .
It feels nice seeing regular HTML tags again. Looking forward to starting working with it.

packages/ui/src/components/Dialog/Dialog.css.ts
alebedev: Started with that, decided against it for 2 reasons
- type safety.  Not compatible with vanilla types unless you specifically downcast media rules to generic `string` which makes code more awkward
- compilation results are different.  Both variants work in Chrome, but I would not be 100% sure about all browsers

alebedev:
### Merge activity

* **Feb 8, 9:54 AM EST**: @alebedev started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3879).

PR #3918, opened 2024-02-13
Fix blog category links

alebedev:
## Describe your changes

Convert CMS href to absolute paths for blog categories


guilhermespopolin (commented):
It seems it's recommended to use relative paths for internal links with `next/link` for the following reasons:

> **Consistency**: Relative paths ensure consistency across your project, as they are relative to the current file or directory structure.
**Portability**: Relative paths make your code more portable. If you move a file around, relative paths will still work correctly, whereas absolute paths might break.
**Simplicity**: Relative paths are simpler and easier to manage, particularly in large projects with deep nesting.

Would it be possible to solve the issue we're having it while still using relative paths? 🤔

gustaveen:
> Would it be possible to solve the issue we're having it while still using relative paths? 🤔

Yes, when using `next/link` we shouldn't use absolute urls. What change was it that caused this to fail? @alebedev

PR #3926, opened 2024-02-14
GEN-1817 - styles(BankIdV6Dialog): hold sucess screen for a moment so users can see it

guilhermespopolin:
## Describe your changes

- Hold _Sucess Screen_ for a moment after sucessfuly login so users can see it.

https://github.com/HedvigInsurance/racoon/assets/19200662/5e3186e3-00c4-43bd-9fad-c516d9573d7e

13.02.mov…

Otherwise they will have no feedback about the login process.


apps/store/src/utils/wait.ts
```diff
@@ -0,0 +1 @@
+export const wait = (ms = 1500) => new Promise<void>((resolve) => setTimeout(resolve, ms))
```
alebedev, line 1: Do we want a default value here?

alebedev (approved):

apps/store/src/utils/wait.ts
guilhermespopolin: 🤔 yeah, probably not. Removing it.

gustaveen (approved):

PR #3934, opened 2024-02-16
Convert Text component to use vanilla-extract with sprinkles

gustaveen:
## Describe your changes
First stab to see how we could convert UI components and still keep the same API 

This can be achieved with using `sprinkles` https://vanilla-extract.style/documentation/packages/sprinkles/. That way we can apply different classes depending on the props we pass to the component.

It also has a neat way of defining responsive styles. For the font size where we can pass in an object like `size={ _: 'sm', md: 'lg', lg: 'xxl' }` we can use `conditions` to get all the breakpoints we want and match it with properties, in this case `fontSizes`


gustaveen:
* `main`
  * **PR #3934** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3934?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3934?utm_source=stack-comment).

apps/store/src/blocks/TextBlock.tsx
```diff
@@ -27,7 +27,7 @@ export const TextBlock = ({ blok }: TextBlockProps) => {
     <StyledText
       {...storyblokEditable(blok)}
       as="div"
-      align={blok.textAlignment}
+      align={blok.textAlignment === '' ? undefined : blok.textAlignment}
```
gustaveen, line 30: Storyblok can pass empty string as value which would cause the Text component to break.

packages/ui/src/components/Text/Text.css.ts
```diff
@@ -0,0 +1,64 @@
...
+
+const responsiveProperties = defineProperties({
+  conditions: {
+    _: {},
+    xs: { '@media': 'screen and (min-width: 480px)' },
+    sm: { '@media': 'screen and (min-width: 640px)' },
+    md: { '@media': 'screen and (min-width: 768px)' },
+    lg: { '@media': 'screen and (min-width: 1024px)' },
+    xl: { '@media': 'screen and (min-width: 1280px)' },
+    xxl: { '@media': 'screen and (min-width: 1536px)' },
```
guilhermespopolin, line 54: Should we use our pre defined breakpoints here?  Sorry in advance if you've already mentioned about adding that...

packages/ui/src/components/Text/Text.tsx
```diff
@@ -1,61 +1,68 @@
...
-import { FontSizeProps, getFontSize, theme, UIColors } from '../../theme'
-import { Space } from '../Space'
+import { FontSizeProps } from '../../theme'
+import {
+  textBase,
+  textSprinkles,
+  textStrikethrough,
+  textUppercase,
+  type TextColor,
+  Sprinkles,
```
guilhermespopolin, line 11: nit
```suggestion
  type Sprinkles,
```

packages/ui/src/components/Text/Text.tsx
```diff
@@ -1,61 +1,68 @@
...
->
+type TextStyleProps = {
+  align?: Sprinkles['textAlign']
+  balance?: boolean
+  className?: string
+  color?: TextColor
+  size?: FontSizeProps
+  strikethrough?: boolean
+  uppercase?: boolean
+}
```
guilhermespopolin, line 22: nit:
A suggestion would be to change `color` and `size` type definitions to use `Sprinkles` type. With that we can remove `TextColor` type created inside `Text.css.ts` file and we'd making explicitly that those `props` are being handled by _sprinkles_.
```suggestion
type TextStyleProps = {
  align?: Sprinkles['textAlign']
  balance?: boolean
  className?: string
  color?: Sprinkles['color']
  size?: Sprinkles['fontSize']
  strikethrough?: boolean
  uppercase?: boolean
}
```

guilhermespopolin (approved):
Nice work with this one 👏🏻 
Nice to see how a more complex case can be handled with `@vanilla-extract/css`.
Just left some minor comments.

PR #3954, opened 2024-02-20
Error handling with app router

alebedev:
## Describe your changes

Add HTTP 500 page for app router

Provide examples of triggering uncaught errors client and server side in pages and app routers


apps/store/src/app/[locale]/layout.tsx
```diff
@@ -54,6 +81,32 @@ export const generateStaticParams = () => {
 
 export default Layout
 
+export const metadata = {
```
Comment (gustaveen) [inline apps/store/src/app/[locale]/layout.tsx:84]: We could use the `Metadata` type from `next` to validate right?
```suggestion
export const metadata: Metadata = {
```

apps/store/src/app/[locale]/layout.tsx
```diff
@@ -54,6 +81,32 @@ export const generateStaticParams = () => {
...
+  metadataBase: new URL(ORIGIN_URL),
+  twitter: { site: '@hedvigapp', card: 'summary_large_image' },
```
Comment (gustaveen) [inline apps/store/src/app/[locale]/layout.tsx:86]: Looks like we should be using `openGraph` instead

https://nextjs.org/docs/app/api-reference/functions/generate-metadata#opengraph

guilhermespopolin (approved):

PR #3976, opened 2024-02-27
Prevent full page rerender after initial shopSessionCustomerUpdate

alebedev:
## Describe your changes

Price calculator was particularly slow after SSN section submit

I tracked root cause to the way we subscribed to session change for tracking and discount banner display

We've run `useEffect` on top level of `ProductPageBlock` which led to full rerender every time customer or session changed even if it had nothing to do with direct children on a block.  I've isolated those effects into childless components.  Rerender after updating customer is not ~3x cheaper

Optimize slow interaction

### DIFF
--- a/apps/store/src/blocks/PageBlock.tsx
+++ b/apps/store/src/blocks/PageBlock.tsx
@@ -8,17 +8,24 @@ type PageBlockProps = SbBaseBlockProps<{
 }>
 
 export const PageBlock = ({ blok }: PageBlockProps) => {
-  useDiscountBanner()
   return (
-    <Main {...storyblokEditable(blok)}>
-      {blok.body.map((nestedBlock) => (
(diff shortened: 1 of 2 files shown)


gustaveen (approved):
Nice improvement! 

Just to understand, where did you see the performance bottle neck in the profiler? Where you using the react profiler for this, and how could you see it's 3x faster now?

alebedev:
@gustaveen I was using react profiler.  3x improvement is cost difference for most expensive renders during SSN section submit (there were 2 or 3 of them)

For every render during profiling you can see which components updated, why (with " Record why each component rendered while profiling." option) and total time it took

Based on that I'm trying to reduce the size of subtree we're rerendering.  Common problem is that some component near root of the tree runs effect on frequently changed value and ends up rerendering all of its children

PR #3985, opened 2024-02-28
Migrate some hooks from next/router to next/navigation

alebedev:
## Describe your changes

Use next/navigation for routing in few more places

Small standalone change extracted from supporting product pages in app router

### DIFF
--- a/apps/store/src/components/ProductPage/PurchaseForm/useOpenPriceCalculatorQueryParam.ts
+++ b/apps/store/src/components/ProductPage/PurchaseForm/useOpenPriceCalculatorQueryParam.ts
@@ -1,4 +1,4 @@
-import { useRouter } from 'next/router'
+import { useSearchParams } from 'next/navigation'
 import { useEffect, useRef } from 'react'
 
 export const OPEN_PRICE_CALCULATOR_QUERY_PARAM = 'openPriceCalculator'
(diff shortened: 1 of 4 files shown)


apps/store/src/components/ProductPage/PurchaseForm/useOpenPriceCalculatorQueryParam.ts
```diff
@@ -8,24 +8,21 @@ type Params = {
...
 
 export const useOpenPriceCalculatorQueryParam = ({ onQueryParamDetected }: Params) => {
-  const router = useRouter()
   const triggeredRef = useRef(false)
-
   const isPriceCalculatorExpanded = useIsPriceCalculatorExpanded()
 
   useEffect(() => {
-    if (router.isReady && isPriceCalculatorExpanded && !triggeredRef.current) {
+    if (isPriceCalculatorExpanded && !triggeredRef.current) {
```
gustaveen, line 15: What did `router.isReady` do?

gustaveen (approved):
Nice that we can start using `next/navigation` in app router

apps/store/src/components/ProductPage/PurchaseForm/useOpenPriceCalculatorQueryParam.ts
alebedev: With `useRouter()` you could get in-between state when you know some of values like path, but not searchParams.  With next/navigation it's no longer needed and `isReady` is removed from API

PR #3994, opened 2024-02-29
GEN-1860 - refact(product-reviews): foundation for fetching reviews client side

guilhermespopolin:
## Describe your changes

* Rename: `ReviewData` --> `ReviewMetadata`. I think it makes more sense since it will only hold _averageRating_ and _reviewsDistribution_.

```
export type ReviewsMetadata = {
  averageRating: Rating
  reviewsDistribution: ReviewsDistribution
}

```
* Rename:  `CompanyReviewsDataProvider` --> `CompanyReviewsMetadataProvider`
* Rename:  `ProductReviewsDataProvider` --> `ProductReviewsMetadataProvider`
* Remove `useReviews`. It was being used for holding state like the `selectedScore` and the `reviews` that should be displayed by the dialog. That functionality will be part of `ReviewsDialog` component since it will be the one responsible for fetching the reviews client side.

This is a series of PRs that's gonna be used as the foundation for getting the following request done by the usage of RSC:

* Being able to render `AverageRatingBanner` block with data about all the reviews (as we're already doing today) **or** a subset of it configured in _Storyblok_. That would allow Peter to use that block for category pages like _Pet Insurance_ as well, by showing average rating, reviews distribution and reviews of _Dog_ and _Cat_ insurance only. That data should also be used for structured data so it's important to have it done during build time.

image


apps/store/src/features/memberReviews/CompanyReviewsMetadataProvider.tsx
```diff
@@ -0,0 +1,20 @@
+import { createContext, useContext, type PropsWithChildren } from 'react'
+import type { ReviewsMetadata } from './memberReviews.types'
+
+const CompanyReviewsMetadataContext = createContext<ReviewsMetadata | null>(null)
```
alebedev, line 4: We should probably use jotai for state management in new features, not React context.  What do you think?

apps/store/src/features/memberReviews/memberReviews.ts
```diff
@@ -83,51 +82,49 @@ const fetchCompanyLatestReviews = async (): Promise<ReviewsData['reviewsByScore'
...
-            score: comment.score,
-            content: comment.content,
-            tag: comment.tag,
-          })),
-        },
-      }
-    },
-    {} as ReviewsData['reviewsByScore'],
-  )
+  return Object.entries(commentsByScore).reduce((acc, [score, data]) => {
```
alebedev, line 89: Can we use `Object.fromEntries` instead of reduce here?

alebedev (commented):
Good work!  Let's discuss state management solution, no other questions

apps/store/src/features/memberReviews/memberReviews.ts
guilhermespopolin: That one is used to transform key-value pair lists into objects so I wouldn't work here. I can't see a better array of transforming an object into another one except `Object.keys` + `reduce` or `Object.entries` + `reduce`. I could also use a `for...of` loop but it seems more complex and it doesn't play well with types.

PR #3996, opened 2024-02-29
GEN-1860 - refact(ReviewComment): move tag logic to component level

guilhermespopolin:
## Describe your changes

* Update the way _review tags_ are obtained. There is a translation key for each product type that follows the patter: `[productId]__REVIEW_COMMENT_TAG`. E.g: `SE_CAT_REVIEW_COMMENT_TAG`. Before I was storing this for each comment on _redis_ which it seems wrong. Now I have a `attributedTo` field for each comment which indicates which product it's connected to. I then build that translation key at the UI level inside `ReviewComment` comment.

image

It feels better this way.

**Not to myself**: Remember to remove `tag` deprecated field from comments stored on _redis_.

### DIFF
--- a/apps/store/src/components/ProductReviews/ReviewComment.tsx
+++ b/apps/store/src/components/ProductReviews/ReviewComment.tsx
@@ -12,16 +12,16 @@ type Props = Review & {
   className?: string
 }
 
-export const ReviewComment = ({ score, date, tag, content, className }: Props) => {
+export function ReviewComment({ score, date, content, attributedTo, className }: Props) {
   const { t } = useTranslation('reviews')
+  const tag = useTag(attributedTo)
   const formatter = useFormatter()
 
   return (
     <div className={clsx(wrapper, className)}>
       <div className={reviewHeader}>
         <Stars score={score} size="1rem" />
-        {/* @ts-expect-error couldn't find a way to tell TS that tag string content belongs to 'reviews' namespace */}
-        {tag && <span className={reviewTag}>{t(tag)}</span>}
+        {<span className={reviewTag}>{tag}</span>}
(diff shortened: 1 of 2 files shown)


apps/store/src/components/ProductReviews/ReviewComment.tsx
```diff
@@ -21,7 +23,7 @@ export const ReviewComment = ({ score, date, tag, content, className }: Props) =
       <div className={reviewHeader}>
         <Stars score={score} size="1rem" />
         {/* @ts-expect-error couldn't find a way to tell TS that tag string content belongs to 'reviews' namespace */}
-        {tag && <span className={reviewTag}>{t(tag)}</span>}
+        {<span className={reviewTag}>{t(`${attributedTo}${TAG_TRANSLATION_KEY_SUFFIX}`)}</span>}
```
alebedev, line 26: Building translation keys from parts breaks scanning for used keys workflow.  How many values for `attributedTo` do we expect?

alebedev: I think one big switch statement that returns a key from productId might be better, even if uglier

guilhermespopolin: We expect one for each product - 8 in total.
I can add that switch, no problem.

alebedev (approved):
Great!

guilhermespopolin:
### Merge activity

* **Mar 13, 11:49 AM EDT**: @guilhermespopolin started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/3996).

PR #4023, opened 2024-03-05
perf(bankid-dialog): defer BankIdDialog/BankIdV6Dialog loading

guilhermespopolin:
## Describe your changes

* Defer the loading of BankId's dialogs.

Better initial performance by not including JS for those modals in the initial bundle.

### DIFF
--- a/apps/store/src/components/BankIdDialogDynamic.tsx
+++ b/apps/store/src/components/BankIdDialogDynamic.tsx
@@ -1,0 +1,21 @@
+import dynamic from 'next/dynamic'
+import { Space } from 'ui'
+import * as FullscreenDialog from '@/components/FullscreenDialog/FullscreenDialog'
+import { Skeleton } from '@/components/Skeleton'
+
+export const BankIdDialogDynamic = dynamic({
+  loader: async () => {
+    const { BankIdDialog } = await import('./BankIdDialog')
+    return BankIdDialog
+  },
+  loading: () => (
+    <FullscreenDialog.Root open={true}>
+      <FullscreenDialog.Modal center={true}>
+        <Space y={0.5}>
+          <Skeleton style={{ height: 22, width: 180 }} />
+          <Skeleton style={{ height: 22, width: 200 }} />
+        </Space>
+      </FullscreenDialog.Modal>
+    </FullscreenDialog.Root>
+  ),
+})
--- a/apps/store/src/pages/_app.tsx
+++ b/apps/store/src/pages/_app.tsx
@@ -10,8 +10,8 @@ import { Provider as BalancerProvider } from 'react-wrap-balancer'
 import globalCss from 'ui/src/global.css'
 import { theme } from 'ui'
 import { AppErrorDialog } from '@/components/AppErrorDialog'
-import { BankIdDialog } from '@/components/BankIdDialog'
-import { BankIdV6Dialog } from '@/components/BankIdV6Dialog/BankIdV6Dialog'
+import { BankIdDialogDynamic } from '@/components/BankIdDialogDynamic'
+import { BankIdV6DialogDynamic } from '@/components/BankIdV6Dialog/BankIdV6DialogDynamic'
 import { ContactUs } from '@/components/ContactUs/ContactUs'
 import { CookieConsentLoader } from '@/components/CookieConsentLoader'
 import { GlobalLinkStyles } from '@/components/RichText/RichText.styles'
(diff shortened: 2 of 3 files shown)


guilhermespopolin:
* **#4023** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4023?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

apps/store/src/pages/_app.tsx
```diff
@@ -54,6 +52,17 @@ const DynamicGlobalBanner = dynamic(() => import('@/components/GlobalBanner/Glob
   ssr: false,
 })
 
+const DynamicBankIdV6Dialog = dynamic(
+  () => import('@/components/BankIdV6Dialog/BankIdV6Dialog').then((mod) => mod.BankIdV6Dialog),
+  {
+    ssr: false,
```
alebedev, line 58: Why no ssr for one but not the other?

apps/store/src/pages/_app.tsx
```diff
@@ -54,6 +52,17 @@ const DynamicGlobalBanner = dynamic(() => import('@/components/GlobalBanner/Glob
...
+  },
```
alebedev, line 59: With no loader there could be a delay between the moment dialog should open and actual opening.  Please check on emulated Fast 3G.  Maybe it would be better experience to have some loader (empty dialog with loading indicator in center, perhaps?)

apps/store/src/pages/_app.tsx
guilhermespopolin: I was basing myself from their [docs](https://nextjs.org/docs/pages/building-your-application/optimizing/lazy-loading#with-no-ssr). They say:

> This is useful if an external dependency or component relies on browser APIs like `window`.

Since `BankIdDialog` doesn't rely on external dependencies and/or browser APIs I didn't toggle off `ssr` flag. But thinking about it now, I think it wouldn't harm load it exclusively on the client as well. I mean, it would be less DOM nodes for parsing initially but I'm not sure if It would be a perceived gain.

apps/store/src/pages/_app.tsx
guilhermespopolin: Yeah, that's a good point. Gonna definitely check that. Thanks man

apps/store/src/pages/_app.tsx
alebedev: If dialog is closed by default, we have no DOM nodes anyway.  I checked with PriceCalculator and found no difference between ssr: false and default mode in either HTML size or rendering time

guilhermespopolin: That's true. I though DOM nodes for the dialog would be present for animation purposes but It isn't.

apps/store/src/pages/_app.tsx
guilhermespopolin: Have updated it by adding a loading state. I couldn't actually reproduce that scenario but I think it's a valid addition.  I guess at the end it's very unlikely a fast enough interaction where that dialog "gets requested by user action" while it JS doesn't load yet. I guess that would be easy to reproduce if dialog resources were loaded as a consequence of of an user interaction, which would be good for performance but doesn't actually fit our current implementation I'd say.

apps/store/src/components/BankIdDialogDynamic.tsx
```diff
@@ -0,0 +1,21 @@
...
+    <FullscreenDialog.Root open={true}>
+      <FullscreenDialog.Modal center={true}>
+        <Space y={0.5}>
+          <Skeleton style={{ height: 22, width: 180 }} />
+          <Skeleton style={{ height: 22, width: 200 }} />
+        </Space>
+      </FullscreenDialog.Modal>
+    </FullscreenDialog.Root>
+  ),
+})
```
guilhermespopolin, line 21: Also, keep default `ssr` config for this one; getting hydration errors otherwise. Since the idea is to replace this altogether with new BankId dialog in a couple of months I don't think it's worth it to spend more time trying to fix it.

alebedev (approved):

guilhermespopolin:
### Merge activity

* **Mar 6, 8:09 AM EST**: @guilhermespopolin started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4023).

PR #4031, opened 2024-03-07
fix: Set aspect ratio for video posters

Youakeem:
## Describe your changes
Make sure video poster images aspect ration is calculated

It's required to have both width and height in order for `getImageProps` to work correctly


guilhermespopolin (approved):
🙌🏻

PR #4032, opened 2024-03-07
fix(bankid): avoid showing an unexpected loading state while the dialog gets lazy loaded

guilhermespopolin:
## Describe your changes

I've noticed an unexpected loading dialog while loading home page. I've accidentally added that..

https://github.com/HedvigInsurance/racoon/assets/19200662/366988b9-f7f5-4a32-86b3-cfcb81a47401

What I was trying to do is having a loading state for bankid dialog in case it gets used before being loaded (it's been lazy loaded right now). The issue tho is that those dialogs doesn't get opened directly by a user interaction so it's hard to load it only when it's needed. So I removed the loading state as I think it will probably be always loaded/cached by the time it gets used so no issues here. Worst case scenario is some delay between clicking on login/sign and getting that dialog opened but I think that's gonna be rare as we trigger the loading of those dialogs code right after any page get's loaded.

### DIFF
--- a/apps/store/src/components/BankIdDialogDynamic.tsx
+++ b/apps/store/src/components/BankIdDialogDynamic.tsx
@@ -1,21 +1,12 @@
 import dynamic from 'next/dynamic'
-import { Space } from 'ui'
-import * as FullscreenDialog from '@/components/FullscreenDialog/FullscreenDialog'
-import { Skeleton } from '@/components/Skeleton'
 
+// By the time it gets used, BankIdDialog will probably be loaded/cached already so
+// no reason to add a 'loading' component here. At the end of the day it's also hard to
+// add a loading component as this dialog is not opened directly by the user
+// but as a consequence of having a BankId operation in progress.
 export const BankIdDialogDynamic = dynamic({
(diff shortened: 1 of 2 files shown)


guilhermespopolin:
* **#4032** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4032?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

alebedev (approved):
Good as quick fix.  Perhaps we can add some loader inside BankID dialog.  Sorry I missed this problem during earlier review

guilhermespopolin:
> Good as quick fix. Perhaps we can add some loader inside BankID dialog. Sorry I missed this problem during earlier review

Yes. Gonna merge this now while keeping thinking on an alternative solution

guilhermespopolin:
### Merge activity

* **Mar 7, 5:05 AM EST**: @guilhermespopolin started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4032).

PR #4035, opened 2024-03-07
Wrap ui/Dialog.Root open/close change into startTransition (performance)

alebedev:
## Describe your changes

Make Dialog open/close change a transition

This should improve responsiveness (INP) for most dialogs.  We still have 3 places where we use  '@radix-ui/react-dialog' directly, 1 already uses startTransition

### DIFF
--- a/packages/ui/src/components/Dialog/Dialog.tsx
+++ b/packages/ui/src/components/Dialog/Dialog.tsx
@@ -1,6 +1,13 @@
 import * as DialogPrimitive from '@radix-ui/react-dialog'
 import { clsx } from 'clsx'
-import { forwardRef, PropsWithChildren, ReactNode } from 'react'
+import {
+  ComponentProps,
+  forwardRef,
+  PropsWithChildren,
+  ReactNode,
+  startTransition,
+  useCallback,
+} from 'react'
 import { contentWrapper, dialogWindow, overlay } from './Dialog.css'
 
 type ContentProps = {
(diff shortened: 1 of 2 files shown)


alebedev:
* **#4035** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4035?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

packages/ui/src/components/Dialog/Dialog.tsx
```diff
@@ -41,7 +48,22 @@ const Overlay = forwardRef<HTMLDivElement, OverlayProps>(({ frosted }, ref) => {
...
 
-export const Root = DialogPrimitive.Root
+export const Root = ({
+  onOpenChange,
+  ...forwardedProps
+}: ComponentProps<typeof DialogPrimitive.Root>) => {
+  // Optimization: dialogs can get quite large, so it's good idea to treat opening/closing as transition
+  const handleOpenChange = useCallback(
+    (newValue: boolean) => {
+      startTransition(() => {
```
guilhermespopolin, line 58: So a confusion point to me is that by marking the state update that would result on the opening/displaying of a dialog as a non urgent (as a transition) wouldn't the time between the interaction between the trigger and the Dialog's rendering be significantly bigger? 🤔

guilhermespopolin (approved):

packages/ui/src/components/Dialog/Dialog.tsx
alebedev: Overhead of `startTransition` is pretty small, so usually not.  The only exception is where we have higher priority React renders that delay transition

PR #4045, opened 2024-03-11
Create main layout for app router routes

Youakeem:


apps/store/src/app/layout.tsx
```diff
@@ -1,11 +1,41 @@
...
+}: LocalizedLayoutProps) {
+  const apolloClient = getApolloClient({ locale })
+  const [{ resources }, productMetadata] = await Promise.all([
+    initTranslationsServerSide(locale),
+    fetchGlobalProductMetadata({ apolloClient }),
+  ])
+
+  return (
+    <RootLayout locale={locale}>
+      <TranslationsProvider locale={locale} resources={resources}>
```
alebedev, line 31: I suspect this will not work - `params` will be empty outside of `[locale]` segment

apps/store/src/components/Header/constants.ts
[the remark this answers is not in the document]
Youakeem: Good to know, will update! Thanks!

apps/store/src/app/layout.tsx
Youakeem: True, I pushed an update that should fix this.

apps/store/src/components/CheckoutHeader/CheckoutHeader.tsx
```diff
@@ -2,8 +2,8 @@ import styled from '@emotion/styled'
 import type { ReactNode } from 'react'
 import { mq, theme } from 'ui'
 import { GridLayout } from '@/components/GridLayout/GridLayout'
-import { MENU_BAR_HEIGHT_MOBILE, MENU_BAR_HEIGHT_DESKTOP } from '@/components/Header/HeaderStyles'
 import { LogoHomeLink } from '@/components/LogoHomeLink'
+import { MENU_BAR_HEIGHT_MOBILE, MENU_BAR_HEIGHT_DESKTOP } from '../Header/constants'
```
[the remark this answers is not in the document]
Youakeem: Good catch, it was auto imported!

apps/store/src/app/layout.tsx
```diff
@@ -1,11 +1,24 @@
...
+
+export default function RootAppLayout({ children }: Props) {
+  return (
+    <ApolloProvider>
+      <ShopSessionProvider>
+        <JotaiProvider>{children}</JotaiProvider>
+      </ShopSessionProvider>
+    </ApolloProvider>
+  )
+}
```
guilhermespopolin, line 21: I guess I'm a little bit confused with this one.
In the [docs](https://nextjs.org/docs/app/building-your-application/routing/pages-and-layouts#root-layout-required:~:text=This%20layout%20is,from%20the%20server) they say:

> [root layoyt] is required and must contain `html` and `body` tags, allowing you to modify the initial HTML returned from the server.

I can see those are defined inside `RootLayout` component so layout are nested by default wouldn't the result page for debugger for example be something like the following? 🤔 

```
    <ApolloProvider>
      <ShopSessionProvider>
        <JotaiProvider>
           <RootLayout>
             <div className={header}>
               <LogoHomeLink />
              </div>
              {children}
            </RootLayout>
        </JotaiProvider>
      </ShopSessionProvider>
    </ApolloProvider>
```

apps/store/src/app/layout.tsx
```diff
@@ -1,11 +1,24 @@
+import { Provider as JotaiProvider } from 'jotai'
 import { Metadata, Viewport } from 'next'
-import { PropsWithChildren } from 'react'
+import { ReactNode } from 'react'
 import { theme } from 'ui'
+import { ShopSessionProvider } from '@/services/shopSession/ShopSessionContext'
 import { ORIGIN_URL } from '@/utils/PageLink'
+import { ApolloProvider } from './providers/ApolloProvider'
```
alebedev, line 7: Folder structure - can / should we move it to `appComponents/providers` instead?

apps/store/src/app/debugger/layout.css.ts
```diff
@@ -0,0 +1,16 @@
...
+import { MENU_BAR_HEIGHT_MOBILE } from '@/components/Header/Header.constants'
+
+export const header = style({
+  display: 'flex',
+  paddingInline: theme.space.md,
+  alignItems: 'center',
+  height: MENU_BAR_HEIGHT_MOBILE,
+
+  '@media': {
+    'screen and (min-width: 768px)': {
```
alebedev, line 12: Maybe `[minWidth.lg]`?

PR #4051, opened 2024-03-13
Add debugger for car trial

gustaveen:
## Describe your changes
Add Car trial debugger that generates a link to the car trial page

Since the trial contract isn't created when the trial returns, we need to poll for it similar to how it's done in the storefront script: https://github.com/HedvigInsurance/storefront/blob/main/script/create-test-car-trial.py

https://github.com/HedvigInsurance/racoon/assets/6661511/aea7e962-16f2-4608-a8ca-624f2e4891f9

Next steps:
See if we should add more parameters to the form like dealerId

Easier debugging for car trials


gustaveen:
* **#4051** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4051?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

apps/store/src/features/carDealership/DebuggerCarTrial/DebuggerCarTrialPage.tsx
```diff
@@ -0,0 +1,174 @@
...
+      <Text as="p" color="signalAmberText">
+        {props.errorMessage}
+      </Text>
+    </Space>
+  </AttentionCard>
+)
+
+const Wrapper = styled.div({ paddingBottom: theme.space.xl })
+
+const ResultRow = styled.div({
```
alebedev, line 168: vanilla?

apps/store/src/features/carDealership/DebuggerCarTrial/DebuggerCarTrialPage.tsx
```diff
@@ -0,0 +1,174 @@
...
+import type { TextSprinkles } from 'ui/src/components/Text/Text.css'
+import { Button, Heading, Space, Text, theme } from 'ui'
+import * as GridLayout from '@/components/GridLayout/GridLayout'
+import { AttentionCard, CampaignCard } from '@/components/InfoCard/InfoCard'
+import { InputCarRegistrationNumber } from '@/components/InputCarRegistrationNumber/InputCarRegistrationNumber'
+import { InputSelect } from '@/components/InputSelect/InputSelect'
+import { PersonalNumberField } from '@/components/PersonalNumberField/PersonalNumberField'
+import { Field, useCarBuyerCreatePageState, createCarTrial } from './CarBuyerCreatePage.helpers'
+
+const TIER_OPTIONS = [
```
alebedev, line 11: Do we have it in dealership code somewhere?

apps/store/src/features/carDealership/DebuggerCarTrial/DebuggerCarTrialPage.tsx
```diff
@@ -0,0 +1,174 @@
...
+
+export const DebuggerCarTrialPage = () => {
+  const [state, setState] = useCarBuyerCreatePageState()
+
+  const handleSubmit = async (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault()
+
+    const form = event.currentTarget
+    const formData = new FormData(form)
+    const ssn = formData.get(Field.SSN) as string
```
alebedev, line 24: Should we add something like `getStringValueOrThrow(formData, key)` ?

apps/store/src/pages/debugger/car-trial.tsx
```diff
@@ -0,0 +1 @@
+export { DebuggerCarTrialPage as default } from '@/features/carDealership/DebuggerCarTrial/DebuggerCarTrialPage'
```
alebedev, line 1: Do we want to return 404 when in production?  It only works in staging anyway

Something like
```
if (process.env.VERCEL_ENV === 'production') throw notFound()
```

apps/store/src/features/carDealership/DebuggerCarTrial/createCarDealershipTrial.ts
```diff
@@ -0,0 +1,155 @@
...
+
+      if (json.trialUrl) {
+        return {
+          type: 'success',
+          trialUrl: json.trialUrl,
+        }
+      }
+
+      retries--
+      console.log('Retrying getting trial contract...')
```
alebedev, line 68: Nit: can you move log one line up?  Having `retries--` next to `wait()` is probably easier to read

apps/store/src/pages/debugger/car-trial.tsx
gustaveen: Yes, let's do that 👍

apps/store/src/features/carDealership/DebuggerCarTrial/CarBuyerCreatePage.helpers.tsx
```diff
@@ -0,0 +1,40 @@
...
+  } | null
+}
+
+export const useCarBuyerCreatePageState = () => {
+  const [state, setState] = useState<CartTrialCreateState>({
+    type: 'IDLE',
+    result: null,
+    parameters: null,
+  })
+  return [state, setState] as const
```
alebedev, line 25: What do you think about exporting `setError / setSuccess` and similar specific callbacks instead of generic `setState`?

alebedev (approved):
Great work!  Some comments inside, up to you which ones should be fixed before merge.  Also, good next step would be to work with @Youakeem on moving this to app router

Youakeem:
I am up to it! Hit me up if you wanna pair on it

apps/store/src/features/carDealership/DebuggerCarTrial/DebuggerCarTrialPage.tsx
gustaveen: We do not, feels ok to do this in the case of the debugger

apps/store/src/features/carDealership/DebuggerCarTrial/DebuggerCarTrialPage.tsx
gustaveen: hm, maybe not needed for the debugger I think, we have field level validation so you can't submit without a valid SSN or reg number

apps/store/src/features/carDealership/DebuggerCarTrial/CarBuyerCreatePage.helpers.tsx
gustaveen: yeah, I can refactor when moving to app router 👍

gustaveen:
> Great work! Some comments inside, up to you which ones should be fixed before merge. Also, good next step would be to work with @Youakeem on moving this to app router

Thanks! For sure! 👍

apps/store/src/features/carDealership/DebuggerCarTrial/DebuggerCarTrialPage.tsx
alebedev: What would happen if we ever allow another value?  Would be nice to keep debugger in sync with primary UI

PR #4052, opened 2024-03-13
Fix loading Storyblok redirects

alebedev:
## Describe your changes

We used to pass `cv: undefined` to storyblok API when loading redirects unless env variable with cache version was defined

Storyblok returned some old data for that case, while not passing `cv` correctly fetched latest data

### DIFF
--- a/apps/store/next.config.mjs
+++ b/apps/store/next.config.mjs
@@ -127,6 +127,7 @@ const config = {
     let storyblokRedirects = []
     if (process.env.NEXT_PUBLIC_FEATURE_STORYBLOK_REDIRECTS === 'true') {
       storyblokRedirects = await getStoryblokRedirects()
+      console.log(`Loaded ${storyblokRedirects.length} redirects from storyblok`)
     }
 
     return [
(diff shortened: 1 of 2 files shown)


alebedev:
* **#4052** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4052?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

apps/store/next.config.mjs
```diff
@@ -127,6 +127,7 @@ const config = {
     let storyblokRedirects = []
     if (process.env.NEXT_PUBLIC_FEATURE_STORYBLOK_REDIRECTS === 'true') {
       storyblokRedirects = await getStoryblokRedirects()
+      console.log(`Loaded ${storyblokRedirects.length} redirects from storyblok`)
```
Youakeem, line 130: Is this intentional or a debugging leftover?

Youakeem (approved):

apps/store/next.config.mjs
alebedev: This is intentional, I think this would be useful as a diagnostic log for every run

PR #4063, opened 2024-03-14
refactor: [RND-542] Implement customer session debugging logic in server actions

Youakeem:
## Describe your changes
Implement the session debugging logic in a server action.

Used services relied on the `req` and `res` objects to handle cookies, which isn't working in app router. The app router cookie persister is created in order to reuse the same services but with the `cookies` util from next.

### DIFF
--- a/apps/store/src/app/debugger/actions.ts
+++ b/apps/store/src/app/debugger/actions.ts
@@ -1,22 +1,69 @@
 'use server'
 
-import { cookies } from 'next/headers'
 import { redirect } from 'next/navigation'
+import { addProduct, updateCustomer } from '@/pages/api/session/create'
+import { getApolloClient } from '@/services/apollo/app-router/rscClient'
+import { CountryCode } from '@/services/graphql/graphql'
+import { setupPriceIntentService } from '@/services/priceIntent/app-router/PriceIntentService.utils'
+import { setupShopSession } from '@/services/shopSession/app-router/ShopSession.utils'
+import { RoutingLocale } from '@/utils/l10n/types'
 import { PageLink } from '@/utils/PageLink'
 
-export const create = async (formData: FormData) => {
-  const ssn = formData.get('ssn')
+const DEFAULT_LOCALE: RoutingLocale = 'se-en'
+const DEFAULT_COUNTRY_CODE = CountryCode.Se
+const TEST_SSN = '199808302393'
+const productNames = ['SE_APARTMENT_RENT', 'SE_ACCIDENT'] as const
 
-  if (typeof ssn !== 'string') {
-    return
+const getRandomEmailAddress = () => {
+  const randomId = Math.random().toString(36).substring(2, 5)
+  return `sven.svensson.${randomId}@hedvig.com`
+}
+
+export const createCustomerSession = async (formData: FormData) => {
+  try {
+    const ssn = formData.get('ssn') || TEST_SSN
+
+    if (typeof ssn !== 'string') {
+      return
+    }
+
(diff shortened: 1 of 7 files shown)


Youakeem:
* **#4063** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4063?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

apps/store/src/app/debugger/actions.ts
```diff
@@ -1,22 +1,161 @@
...
+import { PriceIntentService } from '@/services/priceIntent/PriceIntentService'
+import { setupShopSession } from '@/services/shopSession/app-router/ShopSession.utils'
+import { RoutingLocale } from '@/utils/l10n/types'
 import { PageLink } from '@/utils/PageLink'
 
-export const create = async (formData: FormData) => {
-  const ssn = formData.get('ssn')
+const DEFAULT_LOCALE: RoutingLocale = 'se-en'
+const DEFAULT_COUNTRY_CODE = CountryCode.Se
+const TEST_SSN = '199808302393'
```
alebedev, line 24: Since our fontend repo is public, we should not use real SSNs in code.  Maybe some generated test number or just empty value?

apps/store/src/app/debugger/actions.ts
```diff
@@ -1,22 +1,161 @@
...
+  const randomId = Math.random().toString(36).substring(2, 5)
+  return `sven.svensson.${randomId}@hedvig.com`
+}
+
+export const createCustomerSession = async (formData: FormData) => {
+  try {
+    const ssn = formData.get('ssn') || TEST_SSN
+
+    if (typeof ssn !== 'string') {
+      return
```
alebedev, line 37: Should we throw or return some error state here

alebedev (approved):
Looks good!  Should we remove old API too?

apps/store/src/app/debugger/actions.ts
Youakeem: Fair point. This was already used in the code. I am not familiar of how the test SSN works or why it's needed.

Would any number matching an SSN work? Do we even need to fallback to a test number?

apps/store/src/app/debugger/actions.ts
Youakeem: Looks like this is related to having a test ssn. If we fallback to a string then this condition will never run. However, if we decided to not do that, then ideally we should let the client know an SSN is needed.

Youakeem:
> Looks good!  Should we remove old API too?
> 
> 

Thanks! It's removed in #4070

apps/store/src/app/debugger/actions.ts
alebedev: I think the best way is to rely on input SSN without fallback

apps/store/src/app/debugger/actions.ts
alebedev: More generic question - how should we handle invalid form input in server actions when it's expected that it can happen?  Ideally we want some common patter we use everywhere.

apps/store/src/app/debugger/actions.ts
Youakeem: Sounds good!

apps/store/src/app/debugger/actions.ts
Youakeem: I would say return an error and then use that error on the frontend to highlights the invalid input, but we don't currently have that functionality on the frontend.

guilhermespopolin (approved):
Well done :)

PR #4071, opened 2024-03-18
GEN-1937 - fix(bankid): launching on IOS

guilhermespopolin:
## Describe your changes

Improves a little bit the experience for IOS users that **open up hedvig-com in another browser other than safari**. For the case you've open it on Safari while having another browser as the default one, _Bankid App_ will redirect to your default browser and I'm afraid there's nothing we can do about it. At least the Safari tab that will be on the current state so although not perfect, the user can still go back to hevig-com manually. 

Raised as an issue during a testing session. We can't solve the actual issue but we can at least make things a little bit better for other cases.

### DIFF
--- a/apps/store/src/components/BankIdV6Dialog/BankIdV6Dialog.tsx
+++ b/apps/store/src/components/BankIdV6Dialog/BankIdV6Dialog.tsx
@@ -5,7 +5,7 @@ import Link from 'next/link'
 import { useTranslation } from 'next-i18next'
 import { QRCodeSVG } from 'qrcode.react'
 import { useEffect, useRef, type ReactNode } from 'react'
-import { isMobile, isIOS } from 'react-device-detect'
+import { isMobile, isIOS, isChrome, isFirefox, isOpera } from 'react-device-detect'
 import {
   Button,
   type ButtonProps,
@@ -34,7 +34,7 @@ import {
   qrOnAnotherDeviceLink,
 } from './BankIdV6Dialog.css'
 
-export const BankIdV6Dialog = () => {
+export function BankIdV6Dialog() {
   const { t } = useTranslation('bankid')
   const { startLogin, cancelLogin, cancelCheckoutSign, currentOperation } = useBankIdContext()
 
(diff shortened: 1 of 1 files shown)


apps/store/src/components/BankIdV6Dialog/BankIdV6Dialog.tsx
```diff
@@ -218,20 +218,31 @@ export const BankIdV6Dialog = () => {
...
-  // For IOS though, 'redirect' must have a value. '#bankid-auth' is a 'hack'
-  // for returning to the same safari tab.
-  bankidUrl.searchParams.append('redirect', isIOS ? `${window.location.href}#bankid-auth` : 'null')
+  bankidUrl.searchParams.append('redirect', getReturnURL())
 
   return bankidUrl.toString()
 }
 
-const useTryAgainButtonProps = (bankIdOperation: BankIdOperation | null): Partial<ButtonProps> => {
+function getReturnURL() {
```
Youakeem, line 230: I am curious, how will this change prevent the redirect on iOS?

Youakeem (approved):

apps/store/src/components/BankIdV6Dialog/BankIdV6Dialog.tsx
guilhermespopolin: Yes, so before we had this:

```bankidUrl.searchParams.append('redirect', isIOS ? `${window.location.href}#bankid-auth` : 'null')```

_BankId_ recommends to 'null' as the `redirect` param. That will just work with IOS being the only exception. For it, redirect must be provided otherwise user will have to manually move from _BankId_ app to hedvig-com. Before we were providing `redirect` for IOS but the are cases when it still will not work. So we passing 'null' for IOS as well which will prevent the redirect.

guilhermespopolin:
### Merge activity

* **Mar 19, 12:18 PM EDT**: @guilhermespopolin started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4071).

apps/store/src/components/BankIdV6Dialog/BankIdV6Dialog.tsx
Youakeem: Thank you, sir!

PR #4074, opened 2024-03-18
fix: Validate debug session SSN server side

Youakeem:
## Describe your changes
Remove the fallback test SSN and validate input on the server side.

We agreed to not fallback to a test SSN for this feature and handle missing input on the server.

### DIFF
--- a/apps/store/src/app/debugger/actions.ts
+++ b/apps/store/src/app/debugger/actions.ts
@@ -21,7 +21,6 @@ import { PageLink } from '@/utils/PageLink'
 
 const DEFAULT_LOCALE: RoutingLocale = 'se-en'
 const DEFAULT_COUNTRY_CODE = CountryCode.Se
-const TEST_SSN = '199808302393'
 const productNames = ['SE_APARTMENT_RENT', 'SE_ACCIDENT'] as const
 
 const getRandomEmailAddress = () => {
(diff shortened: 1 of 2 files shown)


Youakeem:
* **#4074** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4074?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

apps/store/src/app/debugger/actions.ts
```diff
@@ -21,20 +21,34 @@ import { PageLink } from '@/utils/PageLink'
...
 const getRandomEmailAddress = () => {
   const randomId = Math.random().toString(36).substring(2, 5)
   return `sven.svensson.${randomId}@hedvig.com`
 }
 
-export const createCustomerSession = async (formData: FormData) => {
-  try {
-    const ssn = formData.get('ssn') || TEST_SSN
+type CreateCustomerSessionFormState = null | {
+  errors?: {
```
alebedev, line 32: Let's think how we can model this better
- Most errors would be non-field specific.  We can probably assume at most one such error at a time
- do we want some explicit `isSuccess: true` or something similar?  Relying on absence of errors is less obvious

Youakeem: We could provide an `isSuccess` flag in the response. However, with server actions, we don't usually expect a response.

A server action can either:
1. Succeed: Usually revalidate or redirect
2. Throws: Show the error component
3. Return specific errors: Show them to the user and, in this case, we'll intentionally access the errors object anyways.

We could have a `generic` errors as well

```
errors?: {
  fields: { ... }
  generic: string[]
}
```

alebedev (approved):

gustaveen (approved):
Looks good!

PR #4079, opened 2024-03-20
GEN-1950 - fix(ProductReviewsMetadataProvider): make sure we have one atom for product

guilhermespopolin:
## Describe your changes

- Fix an issue where the same average rating data were being used for multiple products

https://github.com/HedvigInsurance/racoon/assets/19200662/3b7fbf87-7c38-4199-b6e9-ab4689e6f5eb

That happen becuase product reviews atom were being hydrated only once. We need to make sure we have a product reviews
atom per product.

### DIFF
--- a/apps/store/src/features/memberReviews/ProductReviewsMetadataProvider.tsx
+++ b/apps/store/src/features/memberReviews/ProductReviewsMetadataProvider.tsx
@@ -1,18 +1,24 @@
 'use client'
 import { atom, useAtomValue } from 'jotai'
-import { useHydrateAtoms } from 'jotai/utils'
+import { useHydrateAtoms, atomFamily } from 'jotai/utils'
 import { type PropsWithChildren } from 'react'
+import { useProductData } from '@/components/ProductData/ProductDataProvider'
 import type { ReviewsMetadata } from './memberReviews.types'
 
-const productReviewsMetadataAtom = atom<ReviewsMetadata | null>(null)
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+const productReviewsMetadataAtomFamily = atomFamily((productName: string) =>
+  atom<ReviewsMetadata | null>(null),
+)
 
 type Props = PropsWithChildren<{ productReviewsMetadata: ReviewsMetadata | null }>
 
 export const ProductReviewsMetadataProvider = ({ children, productReviewsMetadata }: Props) => {
-  useHydrateAtoms([[productReviewsMetadataAtom, productReviewsMetadata]])
+  const productData = useProductData()
+  useHydrateAtoms([[productReviewsMetadataAtomFamily(productData.name), productReviewsMetadata]])
   return children
 }
 
 export const useProuctReviewsMetadata = () => {
-  return useAtomValue(productReviewsMetadataAtom)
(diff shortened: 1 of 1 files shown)


guilhermespopolin:
* **#4079** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4079?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

apps/store/src/features/memberReviews/ProductReviewsMetadataProvider.tsx
```diff
@@ -1,18 +1,24 @@
...
 import { atom, useAtomValue } from 'jotai'
-import { useHydrateAtoms } from 'jotai/utils'
+import { useHydrateAtoms, atomFamily } from 'jotai/utils'
 import { type PropsWithChildren } from 'react'
+import { useProductData } from '@/components/ProductData/ProductDataProvider'
 import type { ReviewsMetadata } from './memberReviews.types'
 
-const productReviewsMetadataAtom = atom<ReviewsMetadata | null>(null)
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+const productReviewsMetadataAtomFamily = atomFamily((productName: string) =>
```
Youakeem, line 9: I am not sure I understand why we need to pass `productName` if it's not being used in the atom

guilhermespopolin: TL;DR: That's a bummer but it's just how there API works. That's the way to have an atom for each product but we don't actually need to use the `productName` during atom creation.

Yeah, that's confusing. That one of the things it makes me think with jotai only adds more confusion when compared with plain react context. The thing here is that we need a `productReviewsMetadataAtom` for each product. If we keep only one atom there will be issues because they only hydrate (get initialize) once on the client. So if you start your journey by visiting _car insurance_ product page for example, `productReviewsMetadata` atom will be initialize with _car insurance_ related info and never change it's content while doing page transitions. So by visiting _home insurance_ you'll see _car insurance_ related data still.

Youakeem: I see, would changing how the atom is structured to something like a table which maps a product id/name to it's reviews solve this problem? We would have only one atom let's say we call it `productsReviewsMetaData` but we'll be able to select which product reviews to display.

guilhermespopolin: Yes, that would also works. But can we get this merged this and then I can add that in another PR? People are waiting for this for some time already :)

Youakeem (approved):

apps/store/src/features/memberReviews/ProductReviewsMetadataProvider.tsx
Youakeem: Sure thing! Just approved!

guilhermespopolin:
### Merge activity

* **Mar 20, 10:56 AM EDT**: @guilhermespopolin started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4079).

PR #4091, opened 2024-03-25
RND-524 - fix(cookie-banner): make sure it doesn't get in the way of our dialogs

guilhermespopolin:
# Describe your changes

* Make sure cookie banner consent doesn't get in into the way of our dialogs.

### The issue

Sometimes, cookie banner get's into the way of our dialogs which can be anoying:

https://github.com/HedvigInsurance/racoon/assets/19200662/9ad982bb-a2d2-420a-bb67-274a7f6ddd08

An easy way to reproduce this is load a product page with price calculator opened on mobile (`?openPriceCalculator=1`) while having disabled the cache.

### The reason

That happens for the cases our dialog gets added into the DOM before the cookie consent. Since both are positioned elements in the same stacking context, DOM order is used to determine should should be on top. In this case, cookie banner will be on top which can be annoying.

reason

### The solution

[Nextjs recommends](https://shorturl.at/gkAQ8) to have cookie consent scripts loaded with `beforeInteractive` strategy.  If we follow that recommendation we can make sure our dialogs always get added after one trust into the DOM which then will make they override it without having to fight with `z-index` .

https://github.com/HedvigInsurance/racoon/assets/19200662/aef1f352-76db-42bf-9028-81138f8b36a4

As you can see in the video, cookie banner get's show first and then when our app scripts take in, our dialog is shown on top of it.

### DIFF
--- a/apps/store/src/pages/_app.tsx
+++ b/apps/store/src/pages/_app.tsx
@@ -12,15 +12,14 @@ import { AppErrorDialog } from '@/components/AppErrorDialog'
 import { BankIdDialogDynamic } from '@/components/BankIdDialogDynamic'
 import { BankIdV6DialogDynamic } from '@/components/BankIdV6Dialog/BankIdV6DialogDynamic'
 import { ContactUs } from '@/components/ContactUs/ContactUs'
-import { CookieConsentLoader } from '@/components/CookieConsentLoader'
 import { GlobalBannerDynamic } from '@/components/GlobalBanner/GlobalBannerDynamic'
+import { GTMLoader } from '@/components/GTMLoader'
 import { GlobalLinkStyles } from '@/components/RichText/RichText.styles'
 import { usePublishWidgetInitEvent } from '@/features/widget/usePublishWidgetInitEvent'
 import { useApollo } from '@/services/apollo/client'
 import { AppErrorProvider } from '@/services/appErrors/AppErrorContext'
 import { BankIdContextProvider } from '@/services/bankId/BankIdContext'
 import { CustomerFirstScript, hasHiddenChat } from '@/services/CustomerFirst'
-import { GTMAppScript } from '@/services/gtm'
 import { useInitDatadogAfterInteractive } from '@/services/logger/client'
 import { PageTransitionProgressBar } from '@/services/nprogress/pageTransition'
 import { OneTrustStyles } from '@/services/OneTrust'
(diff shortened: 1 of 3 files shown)


apps/store/src/pages/_app.tsx
```diff
@@ -135,9 +134,4 @@ const ShopSessionTrackingProvider = (props: { children: ReactNode }) => {
   return <TrackingProvider shopSession={shopSession}>{props.children}</TrackingProvider>
 }
 
-// When cookie consent is enabled we need to wait until consent value is available before loading GTM
-// See logic inside CookieConsentLoader
-const GTMLoader = () =>
-  Features.enabled('COOKIE_BANNER') ? <CookieConsentLoader /> : <GTMAppScript />
-
```
gustaveen, line 142: Shouldn't we keep this conditional to keep the behaviour for the feature flag?
```js
const GTMLoader = () =>
  Features.enabled('COOKIE_BANNER') ? <GTMLoader /> : <GTMAppScript />
```

gustaveen (commented):
Looks good overall! 👍

Just wondering if we should keep the feature flag behaviour. Also posted a question on how we treat the the cookie banner in modals but no blocker for this PR though

apps/store/src/pages/_app.tsx
Youakeem: The way I understood it is that `GTMAppScript` is always loaded. If cookie banner is enabled, then its implicitly loaded inside `CookieConsentLoader`, otherwise it's loaded directly.

With this change, we are decoupling cookie banner from GTM. GTM is added directly on the page and the cookie banner script is added in `_document` behind a flag. I think this maintains the same behaviour.

Youakeem (approved):
Great work, man!

apps/store/src/pages/_app.tsx
guilhermespopolin: Yes, that's correct Rafael 👍🏻

gustaveen: Cool, maybe I just misunderstand the behaviour with the feature flag disabled. I though the fallback was loading OneTrust from GTM container and hence let GTM keep track of consent groups as it did before :)

guilhermespopolin:
### Merge activity

* **Mar 25, 10:44 AM EDT**: @guilhermespopolin started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4091).

gustaveen (approved):

PR #4092, opened 2024-03-25
Make sprinkles global

gustaveen:
## Describe your changes
Extract text sprinkles to global sprinkles since they are reusable 

Using sprinkles for commonly used properties reduces the output of CSS

### DIFF
--- a/packages/ui/src/theme/sprinkles.css.ts
+++ b/packages/ui/src/theme/sprinkles.css.ts
@@ -1,0 +1,44 @@
+import { defineProperties, createSprinkles } from '@vanilla-extract/sprinkles'
+import { mediaQueries } from './media'
+import { theme } from './theme'
+import { fontSizes } from './typography'
+
+const textColors = {
+  textPrimary: theme.colors.textPrimary,
+  textSecondary: theme.colors.textSecondary,
+  textSecondaryOnGray: theme.colors.textSecondaryOnGray,
+  textTertiary: theme.colors.textTertiary,
+  textTranslucentPrimary: theme.colors.textTranslucentPrimary,
+  textTranslucentSecondary: theme.colors.textTranslucentSecondary,
+  textTranslucentTertiary: theme.colors.textTranslucentTertiary,
+  textDisabled: theme.colors.textDisabled,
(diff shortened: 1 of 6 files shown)


gustaveen:
* **#4092** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4092?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

packages/ui/src/theme/media-query.ts
```diff
@@ -15,6 +15,13 @@ export const mq = Object.fromEntries(
   }),
 )
 
+// Vanilla extract specific breakpoints object
+export const mediaQueries = Object.fromEntries(
```
Youakeem, line 19: We have the `minWidth` breakpoints in `ui/src/theme/media.ts`. I believe they are the same thing.

packages/ui/src/theme/sprinkles.css.ts
```diff
@@ -0,0 +1,44 @@
+import { defineProperties, createSprinkles } from '@vanilla-extract/sprinkles'
+import { mediaQueries } from './media-query'
+import { theme } from './theme'
+import { fontSizes } from './typography'
+
+const textColors = {
```
Youakeem, line 6: I am wondering if we can just spread `...theme.colors`

guilhermespopolin: Good question

guilhermespopolin (approved):
Nice. I assume you gonna to reuse it in another PR or did I missed something? 🤔

packages/ui/src/theme/media-query.ts
gustaveen: Ah, I see that we have two files for handling media queries, let's remove that when we are fully migrated at least. The syntax is a bit different, but maybe we could map over them and transform in to the correct format. 👍

packages/ui/src/theme/sprinkles.css.ts
Youakeem: I see, makes sense!

Youakeem (approved):

PR #4095, opened 2024-03-26
Migrate Button component to vanilla-extract

gustaveen:
### Migrate Button component to vanilla-extract
The goal is to keep the same API for button component as before:
```js
<Button variant="secondary" size={{ base: 'small', lg: 'medium' }} />
```

Did a lot of digging on different approaches. The main challenge is the responsive variants. 
With `sprinkles` it's easy to have single css rules responsive but that would be to verbose in this case. For creating multi variants you could use `recipes`, but they don't provide a way for creating responsive variants. 

So for creating the the responsive styles I looked at the use cases we have, and we only used breakpoint `lg` for responsive styles. So instead of creating styles for all breakpoints we can start with just supporting `base` and `lg`.

I found `recipes` to be a bit overkill for variants so I used [styleVariants](https://vanilla-extract.style/documentation/api/style-variants/) since we can simply map component props to styles `<button className={buttonVariant[variant]}>` and it also don't require a second package. 

This is a start but we can of course improve on this. The main thing for me is that we can find a way to support responsive variants in some way, especially for supporting the `Heading` component.

Migrate from emotion

### DIFF
--- a/packages/ui/package.json
+++ b/packages/ui/package.json
@@ -55,6 +55,7 @@
     "@radix-ui/react-dialog": "1.0.5",
     "@vanilla-extract/css": "1.14.2",
     "@vanilla-extract/dynamic": "2.1.0",
+    "@vanilla-extract/recipes": "0.5.1",
     "@vanilla-extract/sprinkles": "1.6.1",
     "clsx": "2.1.0",
     "react": "18.2.0",
--- a/packages/ui/src/components/Button/Button.css.ts
+++ b/packages/ui/src/components/Button/Button.css.ts
@@ -1,0 +1,227 @@
+import { style, styleVariants } from '@vanilla-extract/css'
+import { minWidth, theme } from '../../theme'
+
+export const baseButton = style({
+  // opt out of double tap to zoom to immediately respond to taps
+  touchAction: 'manipulation',
+  borderRadius: theme.radius.sm,
+  whiteSpace: 'nowrap',
+  lineHeight: 1,
+  transition: `background-color ${theme.transitions.hover}, box-shadow 0.15s cubic-bezier(0.47, 0.03, 0.49, 1.38) 0s`,
+  display: 'inline-flex',
+  alignItems: 'center',
+  justifyContent: 'center',
+
+  cursor: 'pointer',
+  ':disabled': {
+    cursor: 'default',
+  },
+
+  ':focus-visible': {
+    boxShadow: theme.shadow.focus,
+  },
+})
+
+export const fullWidthStyles = style({
+  width: '100%',
+})
+
+const shadow = style({
+  boxShadow: theme.shadow.default,
+  backdropFilter: 'blur(30px)',
+})
+
+export const buttonVariant = styleVariants({
+  primary: [
+    baseButton,
+    {
+      backgroundColor: theme.colors.gray1000,
+      color: theme.colors.textNegative,
+
+      '@media (hover: hover)': {
+        '&:hover': {
+          backgroundColor: theme.colors.grayTranslucent900,
+        },
+      },
+
+      ':active': {
+        backgroundColor: theme.colors.grayTranslucent900,
+      },
+
+      ':focus-visible': {
+        boxShadow: theme.shadow.focusAlt,
+      },
+
+      selectors: {
+        '&:disabled:not([data-loading])': {
+          backgroundColor: theme.colors.gray200,
+          color: theme.colors.textDisabled,
+        },
+      },
+    },
+  ],
+
+  'primary-alt': [
+    baseButton,
+    shadow,
+    {
+      backgroundColor: theme.colors.green50,
+      color: theme.colors.textPrimary,
+
+      '@media (hover: hover)': {
+        '&:hover': {
+          backgroundColor: theme.colors.green100,
+        },
+      },
+
+      ':active': {
+        backgroundColor: theme.colors.green100,
+      },
+
+      selectors: {
+        '&:disabled:not([data-loading])': {
+          backgroundColor: theme.colors.gray200,
+          color: theme.colors.textDisabled,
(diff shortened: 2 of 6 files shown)


gustaveen:
* **#4095** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4095?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

packages/ui/src/components/Button/Button.css.ts
```diff
@@ -0,0 +1,227 @@
...
+
+      ':active': {
+        backgroundColor: theme.colors.grayTranslucent900,
+      },
+
+      ':not([data-loading])': {
+        ':disabled': {
+          backgroundColor: theme.colors.gray200,
+          color: theme.colors.textDisabled,
+        },
```
guilhermespopolin, line 55: I think those are not being applied.  You might want to change it to the following:

```suggestion
      selectors: {
        ':not([data-loading]):disabled': {
           backgroundColor: theme.colors.gray200,
           color: theme.colors.textDisabled,
        },
      }
```

guilhermespopolin (commented):
[Discussion]
[ Complex selectors](https://vanilla-extract.style/documentation/styling/#complex-selectors) should be written under `selectors` field, which also can take _simple selectors_ like `:hover` and `:hover`. I think it would be easier if we always add selectors under `:selectors` field leaving _first level_ of `style` object containing only CSS properties like `display` and `backgroundColor`. In that way there will be no overhead thinking which approach should your selector be using.

packages/ui/src/components/Button/Button.css.ts
gustaveen: Good catch @Guilherme :raised_hands:

packages/ui/src/components/Button/Button.helpers.ts
```diff
@@ -1,54 +1,19 @@
...
-
-const HEIGHT = {
-  large: '3.5rem',
-  medium: '2.5rem',
-  small: '2rem',
-}
+import { buttonSizeBase, buttonSizeLarge } from './Button.css'
 
 type ButtonSizeVariant = 'small' | 'medium' | 'large'
+type ButtonLevels = 'base' | 'lg'
```
alebedev, line 4: Great that we're reducing number of possible combination here!  Does it generalize to other responsive components - are we mostly/only using `lg` breakpoint?

packages/ui/src/components/Button/Button.css.ts
```diff
@@ -0,0 +1,227 @@
...
+      selectors: {
+        '&:disabled:not([data-loading])': {
+          backgroundColor: theme.colors.gray200,
+          color: theme.colors.textDisabled,
+        },
+      },
+    },
+  ],
+
+  'primary-alt': [
```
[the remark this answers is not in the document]
gustaveen: Good point! That's the nice thing with vanilla extract and how they handle style composition. It will create separate class for the base styles and a new class with the additional styles for the variant 👌 

https://vanilla-extract.style/documentation/api/style-variants/#style-composition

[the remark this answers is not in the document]
gustaveen: Screenshot 2024-04-02 at 18 25 06

packages/ui/src/components/Button/Button.helpers.ts
gustaveen: I think the button is the one component that's not changing in size that much. Heading component will need more breakpoints, but I'll check if there's unused ones. 👍

packages/ui/src/components/Button/Button.helpers.ts
```diff
@@ -1,54 +1,19 @@
...
+export const getButtonSizeStyles = (buttonSize: ButtonSize) => {
+  if (typeof buttonSize !== 'object') {
+    return buttonSizeBase[buttonSize]
   }
 
-  const styles = Object.entries(size).reduce((acc, [level, variant]) => {
-    if (level === 'base') return { ...acc, ...SIZE_STYLES[variant] }
+  const responsiveStyles = Object.entries(buttonSize).map(([level, size]) => {
+    if (level === 'base') return buttonSizeBase[size]
+    if (level === 'lg') return buttonSizeLarge[size]
```
Youakeem, line 15: Just an idea, would having one object of size styles with different properties for breakpoints make this more flexible?

```
return buttonSizes[level][size]?
```

Then we won't need the if conditions and can add support for more breakpoints a little easier.

alebedev (approved):
Great!

packages/ui/package.json
```diff
@@ -55,6 +55,7 @@
     "@radix-ui/react-dialog": "1.0.5",
     "@vanilla-extract/css": "1.14.2",
     "@vanilla-extract/dynamic": "2.1.0",
+    "@vanilla-extract/recipes": "0.5.1",
```
guilhermespopolin, line 58: Should we remove `@vanilla-extract/recipes` then, since we're not gonna use it?

guilhermespopolin (approved):
Noice

packages/ui/package.json
gustaveen: It's used in another component but I think we could easily just use `styleVariants` for that case. I can look at refactoring that 👍

packages/ui/src/components/Button/Button.helpers.ts
gustaveen: That could work! I'll try in a new PR 🙌

Youakeem (approved):

gustaveen:
### Merge activity

* **Apr 3, 5:22 AM EDT**: [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4095) rebased this pull request as part of a merge.

PR #4096, opened 2024-03-26
GEN-1978 - bug(BankIdV6Dialog): remove close button from sucess state

guilhermespopolin:
## Describe your changes

- Remove close button (X) for _BankIdDialog_ success state

That can avoid reaching an inconsistent state inside the app.

### DIFF
--- a/apps/store/src/components/BankIdV6Dialog/BankIdV6Dialog.tsx
+++ b/apps/store/src/components/BankIdV6Dialog/BankIdV6Dialog.tsx
@@ -65,6 +65,7 @@ export function BankIdV6Dialog() {
 
   let Content: ReactNode = null
   let Footer: ReactNode = null
+  let Header: ReactNode = undefined
 
   const { ssn } = currentOperation ?? {}
   if (currentOperation !== null && ssn) {
@@ -169,6 +170,8 @@ export function BankIdV6Dialog() {
       }
 
       case BankIdState.Success: {
+        // Remove close button from header
+        Header = null
         Content = (
           <Text className={iconWithText}>
             <CheckIcon size="1rem" color={theme.colors.signalGreenElement} />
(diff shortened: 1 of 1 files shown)


apps/store/src/components/BankIdV6Dialog/BankIdV6Dialog.tsx
```diff
@@ -166,6 +167,8 @@ export function BankIdV6Dialog() {
       }
 
       case BankIdState.Success: {
+        // Remove close button from header
+        Header = null
```
Youakeem, line 171: I am curious on how this work. Why set it as `undefined` then to `null`. Shouldn't both cases not show the component?

guilhermespopolin: Yeah, that's a little bit confusing as it's usually a good practice with those two _nullable_ values. However it turns out to be handy on this case. If you look at the [implementation of `FullScreenDialog`](https://github.com/HedvigInsurance/racoon/blob/main/apps/store/src/components/FullscreenDialog/FullscreenDialog.tsx#L16-L24) when `Header` is not given we display a custom dialog header that includes a close button. In that case "not given" means `Header === undefined` because if you don't want to provide a custom header but also don't want to have the default header, you can pass `null` which means your given a header that should render nothing.

Youakeem: Oh, now I see! tricky, but I get it! Thanks for explaining, man!

Youakeem (approved):

gustaveen (approved):
Makes sense!

guilhermespopolin:
### Merge activity

* **Mar 27, 5:24 AM EDT**: @guilhermespopolin started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4096).

PR #4115, opened 2024-04-03
GEN-1893 - refact(QuickAdd): removes 'QUICK_ADD_BUNDLE_VIEW' feature flag along with old QuickAdd implementation

guilhermespopolin:
## Describe your changes

* Removes old version of cross sell card along with its feature flag

**Current version**

Screenshot 2024-04-05 at 10 29 01
Screenshot 2024-04-04 at 16 41 12

---

**Old version**

Screenshot 2024-04-05 at 10 31 26

The new/current version was released ~4months ago. While checking with Sonny we don't plan to get it back so it would be nice to remove it before adding the new version of it:

image


guilhermespopolin:
* **#4115** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4115?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

apps/store/src/components/QuickAdd/QuickAddBundleView.stories.tsx
```diff
@@ -129,6 +129,26 @@ export const Default: Story = {
         <Text as="p" color="textTranslucentSecondary">
           Increase your coverage with our accident insurance. Get compensation for dental, scars and
           much more.
+          <QuickAddInfoDialog
```
alebedev, line 132: Can we add another layer on top to make it product-specific?  `QuickAddAccidentInsurance`. I'm assuming we want to store product name, highlights and required fields somewhere - why not in the code or such component?

guilhermespopolin: I see your point. But that would mean to write product specific code right? I mean, we're kind of already doing that since we assume accident and, therefore, uses copy related with accident but ideally all that content could be get from the API right?!

🤔 since we only handle accident today, one thing we can is to add that layer you're mentioning and, inside the container (QuickAddOfferContainer) only deal with accident. We can then have a generic solution for more products if/when that's required. Are we on the same page? In any case, I think that should done down below the stack.

alebedev: Agree, this is more for the future cross-sell changes

alebedev (approved):
Nice cleanup!

PR #4124, opened 2024-04-04
Make sure Button renders <a> tag when href is passed

gustaveen:
## Describe your changes
When a ui button component is wrapped in a link (as in `ButtonNextLink` component), we should output an a tag instead of a button tag as specified in [Next docs ](https://nextjs.org/docs/pages/api-reference/components/link#if-the-child-is-a-functional-component)

Before when we used emotion we could use the as prop to let emotion either render and `<a>` or a `<button>` and the typing was not actually not matching the behaviour, but it didn't complain. But now we need to check if a href is provided and either render an `<a>` or a `<button>`

### The problem
~~I can't find the right way to type this, since forwardRef could either be the an `HTMLAnchorElement` passed from `ButtonNextLink` or a `HTMLButtonElement` passed from a component where we need to pass a ref to a button (see [ManyPetsMigrationPage](https://github.com/HedvigInsurance/racoon/blob/main/apps/store/src/features/manyPets/ManyPetsMigrationPage/ManyPetsMigrationPage.tsx#L107))~~

Use type assertion for now

We need to output a link (<a> tag) when a href is passed

### DIFF
--- a/packages/ui/src/components/Button/Button.tsx
+++ b/packages/ui/src/components/Button/Button.tsx
@@ -1,7 +1,13 @@
 'use client'
 
 import clsx from 'clsx'
-import { ButtonHTMLAttributes, forwardRef, type ReactNode } from 'react'
+import {
+  ButtonHTMLAttributes,
+  forwardRef,
+  type ForwardedRef,
+  type LegacyRef,
+  type ReactNode,
+} from 'react'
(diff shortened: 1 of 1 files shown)


guilhermespopolin (approved):
🚀

alebedev (approved):

PR #4127, opened 2024-04-05
tech: Support using different elements for the Button component

Youakeem:
## Describe your changes
Transform the `Button` component to a polymorphic component with the `as` prop.

Support using different elements with the Button component such as `a` and `Link`.

### DIFF
--- a/apps/store/src/components/ButtonNextLink.tsx
+++ b/apps/store/src/components/ButtonNextLink.tsx
@@ -1,30 +1,13 @@
-import Link, { type LinkProps } from 'next/link'
+import Link from 'next/link'
 import { type ComponentProps } from 'react'
 import { Button } from 'ui'
 
-type ButtonProps = ComponentProps<typeof Button>
+type Props = ComponentProps<typeof Button>
 
-type Props = LinkProps &
-  Partial<Pick<HTMLAnchorElement, 'target' | 'title' | 'download' | 'rel'>> &
-  Pick<ButtonProps, 'variant' | 'size' | 'loading' | 'onClick' | 'children' | 'className'>
-
-export const ButtonNextLink = (props: Props) => {
-  const { onClick, variant, size, loading, children, className, title, target, rel, ...linkProps } =
-    props
+export const ButtonNextLink = ({ children, ...props }: Props) => {
   return (
-    <Link {...linkProps} passHref={true} legacyBehavior={true}>
-      <Button
-        onClick={onClick}
-        variant={variant}
-        size={size}
-        loading={loading}
-        title={title}
-        className={className}
-        rel={rel}
-        target={target}
-      >
-        {children}
-      </Button>
-    </Link>
+    <Button as={Link} passHref={true} legacyBehavior={true} {...props}>
+      {children}
+    </Button>
   )
 }
--- a/apps/store/src/components/ButtonNextLink.tsx
+++ b/apps/store/src/components/ButtonNextLink.tsx
@@ -1,7 +1,7 @@
 import Link from 'next/link'
 import { Button, ButtonProps } from 'ui'
 
-export const ButtonNextLink = ({ children, ...props }: ButtonProps) => {
+export const ButtonNextLink = ({ children, ...props }: ButtonProps<typeof Link>) => {
   return (
     <Button as={Link} passHref={true} legacyBehavior={true} {...props}>
       {children}
--- a/apps/store/src/components/ButtonNextLink.tsx
+++ b/apps/store/src/components/ButtonNextLink.tsx
@@ -3,7 +3,7 @@ import { Button, ButtonProps } from 'ui'
 
 export const ButtonNextLink = ({ children, ...props }: ButtonProps<typeof Link>) => {
   return (
-    <Button as={Link} passHref={true} legacyBehavior={true} {...props}>
+    <Button as={Link} {...props}>
       {children}
     </Button>
   )
--- a/apps/store/src/components/ButtonNextLink.tsx
+++ b/apps/store/src/components/ButtonNextLink.tsx
@@ -1,10 +1,7 @@
 import Link from 'next/link'
-import { type ComponentProps } from 'react'
-import { Button } from 'ui'
+import { Button, ButtonProps } from 'ui'
 
-type Props = ComponentProps<typeof Button>
-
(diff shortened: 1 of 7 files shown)


Youakeem:
* **#4127** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4127?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

apps/store/src/components/ButtonNextLink.tsx
```diff
@@ -1,30 +1,13 @@
-import Link, { type LinkProps } from 'next/link'
+import Link from 'next/link'
 import { type ComponentProps } from 'react'
 import { Button } from 'ui'
 
-type ButtonProps = ComponentProps<typeof Button>
+type Props = ComponentProps<typeof Button>
```
guilhermespopolin, line 5: I think we can import that from `Button` right?

Youakeem: Yup, I wrote this line before noticing that we should expose a `ButtonProps` type. Good catch!

guilhermespopolin (commented):
Nice work with this 👏

guilhermespopolin (approved):

Youakeem:
> Nice work with this 👏 
> 
> 

Thank you, sir!

packages/ui/src/components/Button/Button.tsx
```diff
@@ -1,93 +1,73 @@
...
-    return (
-      <a
-        className={classNames}
-        href={href}
-        ref={ref as LegacyRef<HTMLAnchorElement>}
-        {...buttonProps}
-      />
-    )
-  }
+  const Component = as ?? 'button'
```
gustaveen, line 49: nit: Maybe a minor thing since we should use the `ButtonNextLink`, but you could render a Button as a link without a href

packages/ui/src/components/TypeUtils.ts
```diff
@@ -0,0 +1,20 @@
...
+export type PolymorphicComponentsProp<
+  C extends React.ElementType,
+  Props = Record<string, never>,
+> = React.PropsWithChildren<Props & AsProp<C>> &
+  Omit<React.ComponentPropsWithoutRef<C>, PropsToOmit<C, Props>>
+
+export type PolymorphicComponentPropsWithRef<
+  C extends React.ElementType,
+  Props = Record<string, never>,
+> = PolymorphicComponentsProp<C, Props> & { ref?: PolymorphicRef<C> }
```
gustaveen, line 20: Very nice! 🙌

gustaveen (commented):
Wow, great job with this @Youakeem! 🥳

packages/ui/src/components/TypeUtils.ts
```diff
@@ -0,0 +1,20 @@
+// Polymorphic components utils
+
+type AsProp<C extends React.ElementType> = {
+  as?: C
```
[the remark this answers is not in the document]
Youakeem: I think we could but I am wondering about the value we'll get from this limitation vs the flexibility we have by allowing all elements which keeps the component uncoupled from it's usage with a certain framework or tool.

packages/ui/src/components/TypeUtils.ts
Youakeem: Thank you, sir!

packages/ui/src/components/Button/Button.tsx
Youakeem: I am not sure I am following what you mean

packages/ui/src/components/Button/Button.tsx
```diff
@@ -1,93 +1,73 @@
...
+  const isDisabled = disabled || loading
+  const componentRel = target === '_blank' ? 'noopener' : rel
 
   return (
-    <button className={classNames} ref={ref as ForwardedRef<HTMLButtonElement>} {...buttonProps} />
+    <Component
+      className={classNames}
+      data-loading={loading}
+      disabled={isDisabled}
+      rel={componentRel}
```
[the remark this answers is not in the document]
Youakeem: Great point, will update!

packages/ui/src/components/TypeUtils.ts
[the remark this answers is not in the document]
Youakeem: I also think that It would complicate typing quite a bit

packages/ui/src/components/Button/Button.tsx
gustaveen: If you would write this: 
`<Button as="a">Some button link</Button>`
It would render a link tag without a href and not complain. So that's one benefit of looking at `href` prop when deciding on what element to render. But it feels like an edge case so we could omit it 👍

packages/ui/src/components/TypeUtils.ts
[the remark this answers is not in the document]
gustaveen: Yes, I hear you. It's nice that we don't make it specific to Next.js 👍

packages/ui/src/components/Button/Button.tsx
Youakeem: Oh, I see now! That's because a native anchor doesn't require an href tag. So it's up to the author intention on how to use it.

apps/store/src/features/widget/ProductItem.tsx
```diff
@@ -144,9 +144,11 @@ export const ProductItem = (props: Props) => {
   )
 }
 
-export const ActionButton = forwardRef<HTMLButtonElement, ButtonProps>((props, ref) => {
-  return <Button ref={ref} size="medium" variant="secondary-alt" {...props} />
-})
+export const ActionButton = forwardRef<HTMLButtonElement, ButtonProps<React.ElementType>>(
```
guilhermespopolin, line 147: Should we say those `ActionButton`s are suppose to be used as `button`s? I mean, I understand the flexibility now. I'm just wondering if at this point we already decide the shape of it.

```suggestion
export const ActionButton = forwardRef<HTMLButtonElement, ButtonProps<'button'>>(
```

Youakeem: Sure! Sorry I merged before noticing this comment.

PR #4134, opened 2024-04-08
GEN-1893 - feat(quick-add): enable updating of quick accident offers

guilhermespopolin:
## Describe your changes

* Add _edit_ functionality to cross sell accident card.

https://github.com/HedvigInsurance/racoon/assets/19200662/8ff991a7-9d21-45e1-b11f-d5b3fcb89d1d


apps/store/src/components/QuickAdd/UpdateButton.tsx
```diff
@@ -0,0 +1,143 @@
...
+
+function useEditOffer(params: Params) {
+  const [createPriceIntent, createPriceIntentResult] = usePriceIntentCreateMutation()
+  const [updatePriceIntent, updatePriceIntentResult] = usePriceIntentDataUpdateMutation()
+  const [confirmPriceIntent, confirmPriceIntentResult] = usePriceIntentConfirmMutation()
+  const [updateOfferStartDate, updateOfferStartDateResult] = useStartDateUpdateMutation()
+
+  const editOffer = async () => {
+    try {
+      const { data: createPriceIntentData } = await createPriceIntent({
```
alebedev, line 73: Would this create unneeded priceIntents when we're updating existing offer?

For example
- x-sell card first shown (no offer)
- user edits date or number of co-insured and clicks update (offer created)
- user edits something again

apps/store/src/components/QuickAdd/UpdateButton.tsx
```diff
@@ -0,0 +1,143 @@
...
+          productName: params.productName,
+        },
+      })
+      const priceIntentId = createPriceIntentData?.priceIntentCreate.id
+      if (!priceIntentId) {
+        datadogLogs.logger.error('Cross sell | price intent could not be created')
+        return
+      }
+
+      const { errors: updatePriceIntentErrors } = await updatePriceIntent({
```
alebedev, line 85: Similar thing, what if user only wants to update start date and we don't need to change and confirm priceIntent?

apps/store/src/components/QuickAdd/UpdateButton.tsx
```diff
@@ -0,0 +1,143 @@
...
+        },
+      })
+      if (updatePriceIntentErrors) {
+        datadogLogs.logger.error('Cross sell | price intent could not be updated', {
+          priceIntentId,
+        })
+        return
+      }
+
+      const { data: confirmPriceIntentData } = await confirmPriceIntent({
```
alebedev, line 99: Since edit + confirm is always a single action, we can probably try to make GraphQL operation with both mutations in the same document.  Less network roundtrips, for one

alebedev (dismissed):
Left some comments on coordinating GraphQL operations, I think we need to solve them before proceeding

apps/store/src/components/QuickAdd/QuickAddEditableView.tsx
```diff
@@ -18,20 +22,48 @@ import {
...
+    getOfferNumberCoInsuredWithFallback(internalOffer),
+  )
+
+  const resetForm = (offer: OfferRecommendationFragment) => {
+    setInternalOffer(offer)
+    setStartDate(getOfferStartDateWithFallback(offer))
+    setNumberCoInsured(getOfferNumberCoInsuredWithFallback(offer))
+  }
+
+  const isFormPristine = useMemo(() => {
```
Youakeem, line 56: I like the naming :D

apps/store/src/components/QuickAdd/QuickAddEditableView.tsx
```diff
@@ -18,20 +22,48 @@ import {
...
+    const offerDate = getOfferStartDateWithFallback(internalOffer)
+    const isStartDatePristine = isSameDay(offerDate, startDate)
+    const offerNumberCoInsured = getOfferNumberCoInsuredWithFallback(internalOffer)
+    const isNumberCoInsuredPristine = offerNumberCoInsured === numberCoInsured
+
+    return isStartDatePristine && isNumberCoInsuredPristine
```
Youakeem, line 62: What do you think about consolidating the form state in one `useState` hook and provide a default then check here if the current state === the default state?

apps/store/src/components/QuickAdd/QuickAddEditableView.tsx
```diff
@@ -83,24 +116,43 @@ export function QuickAddEditableView(props: Props) {
...
+            {isFormPristine ? (
+              <AddToCartButton
+                shopSessionId={props.shopSessionId}
+                offer={internalOffer}
+                productName={props.productName}
+              >
+                {t('QUICK_ADD_BUTTON')}
+              </AddToCartButton>
+            ) : (
+              <UpdateButton
```
Youakeem, line 136: Do we need to have two buttons? I am thinking of we handle the action onSubmit we can check the "intent" and handle the logic in the submit handler instead.

```

File: apps/store/src/components/QuickAdd/UpdateButton.tsx (hunk 1 above)
Comment (guilhermespopolin) [reply apps/store/src/components/QuickAdd/UpdateButton.tsx]: Have update it. We now use `priceIntentSerivice` which persists the price intent.

File: apps/store/src/components/QuickAdd/QuickAddEditableView.tsx (hunk 6 above)
Comment (guilhermespopolin) [reply apps/store/src/components/QuickAdd/QuickAddEditableView.tsx]: Have update it. Kept "add" action untouched since it's being used by other view as well.

File: apps/store/src/components/QuickAdd/useEditCrossSellOfferForm.ts [hunk 7]
```diff
@@ -0,0 +1,222 @@
...
+  const apolloClient = useApolloClient()
+  const [state, disptach] = useFormState(initialOffer)
+
+  const handleSubmit: FormEventHandler<HTMLFormElement> = async (event) => {
+    event.preventDefault()
+
+    try {
+      disptach({ type: 'SUBMIT' })
+
+      const hasNumberCosInsuredChanged =
```
Youakeem, line 35: Typo?

guilhermespopolin: Yes. Thanks :)

apps/store/src/components/QuickAdd/useEditCrossSellOfferForm.ts
```diff
@@ -0,0 +1,222 @@
...
+  | { type: 'ERROR' }
+  | { type: 'REST_FORM'; payload: OfferRecommendationFragment }
+
+function reducer(state: State, action: Action): State {
+  switch (action.type) {
+    case 'CHANGE_START_DATE':
+      return {
+        ...state,
+        [Fields.START_DATE]: action.payload,
+        isPristine: !hasFormChanged({
```
Youakeem, line 159: Wouldn't this always be true because it's invoked when the start date changes?

I also think that changing the date doesn't affect the quote. So I assume we don't need the update step here.

apps/store/src/components/QuickAdd/useEditCrossSellOfferForm.ts
```diff
@@ -0,0 +1,222 @@
...
+}
+
+type Action =
+  | { type: 'CHANGE_START_DATE'; payload: Date }
+  | { type: 'CHANGE_NUMBER_CO_INSURED'; payload: number }
+  | { type: 'SUBMIT' }
+  | { type: 'ERROR' }
+  | { type: 'REST_FORM'; payload: OfferRecommendationFragment }
+
+function reducer(state: State, action: Action): State {
```
Youakeem, line 153: I am wondering if we need a reducer here. Would something like this work?

```
const handleChange = (key, value) => {
 setFormState((prevState) => ({
  ...prevState,
  [key]: value,
  isPristine: !hasFormChanged(...),
 }))
}

const handleSubmit = () => {
  // use formState here for the mutations
}
```

apps/store/src/components/QuickAdd/useEditCrossSellOfferForm.ts
guilhermespopolin: Start date don't affect the quote in terms of pricing but we do need to change it so the start date is correct in case that quote is added to the cart. Or maybe I miss understood your point ?! 🤔

Youakeem: You are right, I forgot that this state is used for the add action as well.

apps/store/src/components/QuickAdd/useEditCrossSellOfferForm.ts
Youakeem: For the reset action, I am thinking of having a `getDefaultFormValues` function.

```
const handleChange = (key, value) => {
 setFormState((prevState) => ({
  ...prevState,
  [key]: value,
  isPristine: !hasFormChanged(...),
 }))
}

const resetForm = () => {
  setFormState(getDefaultFormValues())
}

const handleSubmit = () => {
  // use formState here for the mutations
}
```

I find this easier to read and reason about but a reducer works as well!

Youakeem (approved):

apps/store/src/components/QuickAdd/useEditCrossSellOfferForm.ts
```diff
@@ -0,0 +1,220 @@
...
+
+      const hasStartDateChanged = !isSameDay(
+        state[Fields.START_DATE],
+        getOfferStartDateWithFallback(updatedOffer),
+      )
+      if (hasStartDateChanged) {
+        updatedOffer = await updateOfferStartDate(updatedOffer.id, state[Fields.START_DATE])
+      }
+
+      disptach({ type: 'REST_FORM', payload: updatedOffer })
```
Youakeem, line 93: It seems like `REST_FORM` is not resetting it to the default value but setting the state to match the updated offer, did I get that right?

guilhermespopolin: Yes, you got that reset. The idea is to "reset" it based on a offer. Meaning being `prstine`, in the `idle` status and having form fields initialized based on the offer.

guilhermespopolin:
### Merge activity

* **Apr 16, 5:51 AM EDT**: @guilhermespopolin started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4134).

PR #4137, opened 2024-04-09
Convert Banner styles to vanilla-extract

alebedev:
## Describe your changes

Subj

## Justify why they are needed

Refactor before changing banner-related logic in next PR

## Checklist before requesting a review

- [ ] I have performed a self-review of my code

PR #4135, opened 2024-04-09
Move breadcrumbs structured data from head to body (prevents duplication)

alebedev:
## Describe your changes

Fix duplication of `structured-data-breadcrumb-list` element during navigation by moving it from `head` to `body`

## Checklist before requesting a review

- [ ] I have performed a self-review of my code

PR #4138, opened 2024-04-09
Move CMS announcement block inside content component

alebedev:
## Describe your changes

Move announcement banner block inside page content

Align it visually above header with `order: -1` CSS rule

Support for persistent layout in app router -> layout cannot know about page details, therefore announcement is moved out of it

### DIFF
--- a/apps/store/src/blocks/PageBlock.css.ts
+++ b/apps/store/src/blocks/PageBlock.css.ts
@@ -3,3 +3,7 @@ import { style } from '@vanilla-extract/css'
 export const main = style({
   width: '100%',
 })
+
+export const announcementBanner = style({
+  order: -1,
+})
--- a/apps/store/src/components/Banner/Banner.tsx
+++ b/apps/store/src/components/Banner/Banner.tsx
@@ -12,13 +12,14 @@ import type { BannerVariant } from './Banner.types'
 
 type Props = {
   children: ReactNode
+  className?: string
   handleClose: () => void
   variant?: BannerVariant
 }
 
-export const Banner = ({ children, handleClose, variant = 'info' }: Props) => {
+export const Banner = ({ children, handleClose, variant = 'info', className }: Props) => {
   return (
-    <div className={clsx(bannerRoot.base, bannerRoot[variant])}>
+    <div className={clsx(bannerRoot.base, bannerRoot[variant], className)}>
       <div className={bannerContent}>
         <Icon variant={variant} className={bannerIcon} />
         {children}
--- a/apps/store/src/components/Banner/Banner.tsx
+++ b/apps/store/src/components/Banner/Banner.tsx
@@ -1,13 +1,14 @@
-import styled from '@emotion/styled'
-import type { ComponentPropsWithoutRef} from 'react';
+import { clsx } from 'clsx'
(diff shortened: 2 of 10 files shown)


alebedev:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4138?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#4138** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4138?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#4137** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4137?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4136** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4136?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4135** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4135?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

apps/store/src/blocks/PageBlock.css.ts
```diff
@@ -3,3 +3,7 @@ import { style } from '@vanilla-extract/css'
 export const main = style({
   width: '100%',
 })
+
+export const announcementBanner = style({
+  order: -1,
+})
```
guilhermespopolin, line 9: 🤔 my only concern with this is accessibility and DOM order. There's a close button inside the announcement that will not be the first focusable thing, despite its rendering position. Not sure if that's problematic on this case tho.

Youakeem: Great point! It's tricky with the page layout. The announcement block is in the page but it appears before the header which comes from the layout.

I am thinking maybe we can treat it like a toaster. Have `AnnouncementBlock` in the layout, but trigger it in the page with a `useEffect` and `addAnnouncement`.

Would something like this work?

apps/store/src/components/Banner/Banner.tsx
```diff
@@ -12,13 +12,14 @@ import type { BannerVariant } from './Banner.types'
 
 type Props = {
```
Youakeem, line 13: Should we `extend ComponentPropsWithoutRef<"div">` instead of manually typing children and classname?

apps/store/src/blocks/PageBlock.css.ts
alebedev: I like toaster solution!  We're running on effects anyway, so first appearance delay is unavoidable.  On the other hand, not having to use `order: -1` makes everything more obvious

gustaveen: I don't think it's the end of the world if it's not the first focusable element, I doubt that announcements are dismissed that often. But if it's possible in another way, sure!

gustaveen (approved):
Works for me but feel free to update if you fins a better solution 👍

guilhermespopolin (approved):

PR #4142, opened 2024-04-10
Refactor GlobalBanner to model different banners identified by id

alebedev:
## Describe your changes

Refactor GlobalBanner to support new logic we'll need for CMS-driven page announcements

- Expose atoms directly instead of providing hook wrappers, wasn't valuable enough abstraction layer
- Change dismiss logic so that specific banner is dismissed instead of current visible banner
- Document why we support only single banner for now

Step towards moving AnnouncementBlock outside of global layout


alebedev:
* **#4142** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4142?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

apps/store/src/components/ForeverPage/ForeverPage.tsx
```diff
@@ -26,13 +25,11 @@ export const ForeverPage = ({ code: initialCode }: Props) => {
 
   const { shopSession } = useShopSession()
   const locale = useRoutingLocale()
-  const { addBanner } = useGlobalBanner()
   const redirectUrl = PageLink.store({ locale })
   const [addCampaign, { errorMessage, loading, called }] = useRedeemCampaign({
     shopSessionId: shopSession?.id ?? '',
-    async onCompleted() {
-      await router.push(redirectUrl)
-      addBanner(t('GLOBAL_BANNER_CAMPAIGN'))
```
alebedev, line 35: wasn't needed, banner appears automatically when there's a discount code applied

apps/store/src/components/GlobalBanner/GlobalBanner.tsx
```diff
@@ -1,28 +1,31 @@
...
+import {
+  dismissedBannedIdAtom,
+  globalBannerAtom,
+} from '@/components/GlobalBanner/globalBannerState'
+import { useDebugShopSessionId } from '@/utils/useDebugShopSessionId'
 
 const GlobalBanner = () => {
-  const { banner, dismissBanner } = useGlobalBanner()
+  const globalBanner = useAtomValue(globalBannerAtom)
+  const setDismissedBannerId = useSetAtom(dismissedBannedIdAtom)
```
Youakeem, line 14: Typo?  banned instead of banner

apps/store/src/components/GlobalBanner/globalBannerState.ts
```diff
@@ -0,0 +1,42 @@
...
+
+type SetBannerOptions = {
+  force: boolean // If true, replace existing banner when present and reset dismissed flag
+}
+
+// Represents 0 or 1 global banner announcements and tracks visible/dismissed state within the same session
+// Banner is identified by string id, so dismissing a banner does not prevent banner with different id from being shown
+//
+// We intentionally don't support banner stack or any other forms of multiple banners being tracked or displayed
+export const globalBannerAtom = atom(
```
Youakeem, line 15: I feel like I don't fully understand the expected behaviour here. Since we only want to support one banner at a time, why do we need the `dismissedGlobalBannerId` state?

I am thinking we could either have a banner or not, and the dismiss banner function should be "internal", used by the close button and when a new banner is triggered.

```
const [showBanner, dismissBanner] = useGlobalBanner()

// Implementation psuedo-code for useGlobalBanner
const globalBannerAtom = atom(...)

const [globalBanner, setGlobalBanner] = useAtom(globalBannerAtom)

const showBanner = (bannerData) => {
  setGlobalBanner(bannerData)
}

const dismissBanner = () => {
  setGlobalBanner(null)
}
```

Am I missing something?

apps/store/src/components/GlobalBanner/globalBannerState.ts
```diff
@@ -0,0 +1,42 @@
...
+    }
+    if (options?.force && dismissedId === newBanner.id) {
+      set(dismissedBannerIdAtom, null)
+    }
+  },
+)
+
+const currentBannerAtom = atom<BannerWithId | null>(null)
+
+export const dismissedBannerIdAtom = atomWithStorage<string | null>(
```
Youakeem, line 38: I understand that we can only have one banner at a time, but can't we have multiple dismissed banners?

gustaveen (approved):
Nice refactoring! 🙌

apps/store/src/components/GlobalBanner/globalBannerState.ts
alebedev: Thought about that and there're probably no cases where we'll need it with current banners we have.  But if we need to support it, we can change this atom to a list and store multiple dismissed IDs

Youakeem (approved):

PR #4144, opened 2024-04-10
styles(InputDay): migrate InputDate to vanilla css

guilhermespopolin:
## Describe your changes

* Migrate `InputDay` styles to use `vanilla-extract/css`

Gonna need it for adjusting some stuff for Accident Cross Sell.

### DIFF
--- a/apps/store/src/components/InputDay/InputDay.css.ts
+++ b/apps/store/src/components/InputDay/InputDay.css.ts
@@ -1,0 +1,136 @@
+import { style, keyframes, styleVariants, globalStyle } from '@vanilla-extract/css'
+import { theme } from 'ui/src/theme'
+
+const slideUpAndFadeAnimation = keyframes({
+  '0%': { opacity: 0, transform: 'translateY(10px)' },
+  '100%': { opacity: 1, transform: 'translateY(0)' },
+})
+
+const slideDownAndFadeAnimation = keyframes({
+  '0%': { opacity: 0, transform: 'translateY(-10px)' },
+  '100%': { opacity: 1, transform: 'translateY(0)' },
+})
+
+export const trigger = style({
+  position: 'relative',
+  width: '100%',
+  height: '4.5rem',
+  borderRadius: theme.radius.sm,
+  backgroundColor: theme.colors.translucent1,
+  cursor: 'pointer',
+  ':focus-visible': {
+    boxShadow: theme.shadow.focus,
+    borderRadius: theme.radius.sm,
+  },
+})
+
+export const label = style({
+  position: 'absolute',
+  left: theme.space.md,
+  top: theme.space.sm,
+  fontSize: theme.fontSizes.xs,
+  overflow: 'visible',
+  whiteSpace: 'nowrap',
+  textOverflow: 'ellipsis',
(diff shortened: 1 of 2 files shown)


apps/store/src/components/InputDay/InputDay.css.ts
```diff
@@ -0,0 +1,136 @@
...
+  position: 'absolute',
+  left: theme.space.md,
+  top: theme.space.sm,
+  fontSize: theme.fontSizes.xs,
+  overflow: 'visible',
+  whiteSpace: 'nowrap',
+  textOverflow: 'ellipsis',
+  color: theme.colors.textSecondary,
+  selectors: {
+    '&&[data-disabled=true]': {
```
alebedev, line 37: What does `&&` do vs just `&`?

apps/store/src/components/InputDay/InputDay.css.ts
```diff
@@ -0,0 +1,136 @@
...
+  paddingRight: '1.25rem',
+  paddingBottom: theme.space.sm,
+  paddingLeft: theme.space.md,
+})
+
+const chevronIconBase = style({
+  pointerEvents: 'none',
+})
+export const chevronIcon = styleVariants({
+  animated: [
```
alebedev, line 59: If you do class composition with clsx in React, you can remove `chevronIconBase` from every variant - this'll be less code overall

apps/store/src/components/InputDay/InputDay.tsx
```diff
@@ -104,146 +117,40 @@ export const InputDay = (props: Props) => {
...
+          alignOffset={-4}
+        >
           {/* Fix issue on iOS where month dropdown is focused and can't be tapped */}
           <div tabIndex={1} aria-hidden={true} />
 
-          <StyledDayPicker
+          <DayPicker
             {...dayPickerProps}
+            className={dayPicker}
+            style={
```
alebedev, line 136: I think you can assign vars in vanilla-extract styles.  It's probably better when values are static

alebedev (approved):
Great work!

apps/store/src/components/InputDay/InputDay.tsx
guilhermespopolin: I couldn't make it work 😢 
I mean, couldn't find a way to redefine those third party CSS variables inside `vanilla-extract`

apps/store/src/components/InputDay/InputDay.css.ts
guilhermespopolin: That's true. Gonna update it.

apps/store/src/components/InputDay/InputDay.css.ts
guilhermespopolin: To be honest I've just migrated that one. It seems '&&' is used to increase specificity 🤔 
We don't actually need it here so I'm going to replaced with single '&'

guilhermespopolin:
### Merge activity

* **Apr 11, 4:05 AM EDT**: @guilhermespopolin started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4144).

PR #4158, opened 2024-04-18
Add widget loaded event

gustaveen:
## Describe your changes
Send an event when the widget has loaded.
Currently it will send it 2 times if the partner loads the initial widget page and then proceeds to the flow page. But it doesn't hurt I suppose. Or do you think we should make sure to only send it once? I'm thinking using an atom to check if it's been sent on a previous screen.

Screenshot 2024-04-18 at 10 52 05

Request from partners to have an init event

### DIFF
--- a/apps/store/src/features/widget/publishWidgetEvent.ts
+++ b/apps/store/src/features/widget/publishWidgetEvent.ts
@@ -1,4 +1,5 @@
 enum Status {
+  WidgetLoaded = 'widgetLoaded',
   Success = 'success',
   Signed = 'signed',
   Close = 'close',
@@ -9,6 +10,7 @@ enum Status {
  * WidgetEvent
  * Event sent from the widget to the parent window for partner integrations to listen to.
  *
+ * - `widgetLoaded` - Sent oncne the widget has first been render on widget landing page or first step of the flow page
  * - `signed` - The user has successfully signed the insurance
  * - `success` - The user has successfully signed the insurance and connected payment
  * - `close` - The user pressed one of our close buttons.
@@ -16,6 +18,7 @@ enum Status {
  */
 
 export type WidgetEvent =
+  | { status: Status.WidgetLoaded }
   | { status: Status.Success }
   | { status: Status.Signed }
   | { status: Status.Close }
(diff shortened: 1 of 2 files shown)


gustaveen:
* **#4158** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4158?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

apps/store/src/features/widget/usePublishWidgetInitEvent.ts
```diff
@@ -1,14 +1,22 @@
...
-import { useEffect } from 'react'
+import { useEffect, useRef } from 'react'
 import { publishWidgetEvent } from './publishWidgetEvent'
 import { EXTERNAL_REQUEST_ID_QUERY_PARAM } from './widget.constants'
 
 export const usePublishWidgetInitEvent = () => {
   const router = useRouter()
 
   const requestId = router.query[EXTERNAL_REQUEST_ID_QUERY_PARAM]
+  const widgetLoadedSent = useRef(false)
```
gustaveen, line 10: Avoid sending it twice on flows initial page, but maybe this will just happen in dev mode?

alebedev: This only happens in dev mode, but probably good idea to deduplicate as you did

apps/store/src/features/widget/publishWidgetEvent.ts
```diff
@@ -1,4 +1,5 @@
 enum Status {
+  WidgetLoaded = 'widgetLoaded',
```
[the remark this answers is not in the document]
alebedev: No better ideas yet

apps/store/src/features/widget/usePublishWidgetInitEvent.ts
alebedev: Maybe you can set some global flag on `window` instead to avoid double event across pages?

alebedev (approved):
Would be nice to fix event duplication, see comment inside.  Other option is just documenting it - you know best which way to go

PR #4160, opened 2024-04-18
Set correct no lang

gustaveen:
## Describe your changes
Set correct no lang - currently a missmatch, should be no-NO
image (5)

Request from Peter


alebedev (approved):
Curious how long it took to notice this.  The fix looks good

gustaveen:
> Curious how long it took to notice this. The fix looks good

Yes interesting that it was picked up right now, it's been like this for a long time

PR #4167, opened 2024-04-22
GEN-1893 - feat(QuickAddEditableView): updating start date without extra step

guilhermespopolin:
## Describe your changes

* Remove `AddToCartButton` component and move its functionality to a hook: `useAddRecommendationOfferToCart`. In that way I can share the same "add to car logic" between bundle and editable view while I can change new cross sell card by unifying how all actions are handled: by submitting the form.
* Add `react-hook-form` and refactor `QuickAddEditableView` (new cross sell card) to make usage of it. It's less "manage form state" code to maintain while I can have access to things file field "dirtyness". More on that below;
* Update how `QuickAddEditableView` works:
  * Show "Update price" button only when _numberCoInsured_ field is _dirty_.
  * When only _start date_ has change we keep the "Add" button and, when clicked, update the start date (when necessary) and then add the updated offer to the cart.

Better UX as we don't require two click now for cases where only start date has changed.

https://github.com/HedvigInsurance/racoon/assets/19200662/d57b8718-91e7-4d62-add1-3dc4c31a5f15


apps/store/src/components/QuickAdd/useEditCrossSellOfferForm.ts
```diff
@@ -69,152 +87,66 @@ export function useEditCrossSellOfferForm({ shopSessionId, productName, initialO
...
-      )
-      if (hasStartDateChanged) {
-        updatedOffer = await updateOfferStartDate(updatedOffer.id, state[Fields.START_DATE])
+      const hasNumberCoInsuredChanged = getFieldState(Fields.NUMBER_CO_INSURED).isDirty
+      const hasStartDateChanged = getFieldState(Fields.START_DATE).isDirty
+      const itent = hasNumberCoInsuredChanged ? 'get-price' : 'add-to-cart'
+
+      let updatedOffer = offer
+      if (itent === 'get-price') {
+        if (hasNumberCoInsuredChanged) {
```
gustaveen, line 98: Do we need to check `hasNumberCoInsuredChanged` since we already do it on line 94?

apps/store/src/components/QuickAdd/useEditCrossSellOfferForm.ts
```diff
@@ -69,152 +87,66 @@ export function useEditCrossSellOfferForm({ shopSessionId, productName, initialO
...
-
-      const hasStartDateChanged = !isSameDay(
-        state[Fields.START_DATE],
-        getOfferStartDateWithFallback(updatedOffer),
-      )
-      if (hasStartDateChanged) {
-        updatedOffer = await updateOfferStartDate(updatedOffer.id, state[Fields.START_DATE])
+      const hasNumberCoInsuredChanged = getFieldState(Fields.NUMBER_CO_INSURED).isDirty
+      const hasStartDateChanged = getFieldState(Fields.START_DATE).isDirty
+      const itent = hasNumberCoInsuredChanged ? 'get-price' : 'add-to-cart'
```
gustaveen, line 94: nit: 
```suggestion
      const intent = hasNumberCoInsuredChanged ? 'get-price' : 'add-to-cart'
```

gustaveen (approved):
Great work! Much nicer UX now

Have we talked about adding `react-hook-form` before? I think I remember some discussion when we started with coverage store :) IMO I think it makes sense to use it 👍 

Just hade some small comments

apps/store/src/components/QuickAdd/useEditCrossSellOfferForm.ts
guilhermespopolin: Yes, that's true. Removing it.

guilhermespopolin:
> Great work! Much nicer UX now
> 
> Have we talked about adding `react-hook-form` before? I think I remember some discussion when we started with coverage store :) IMO I think it makes sense to use it 👍
> 
> Just hade some small comments

Yes, so that was mentioned by Rafael earlier. At that point, we didn't need field state like `dirty`. I'm gonna use it now for switching between _Add_ and _Update Price_ button. Firstly I've coded myself but the end result is more code to maintain and a little of _Reinventing the wheel_. So I though we could leverage `react-hook-form` for this case.

guilhermespopolin:
### Merge activity

* **Apr 25, 5:49 AM EDT**: @guilhermespopolin started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4167).

PR #4186, opened 2024-04-25
Add support for Heading in TickerBlock

gustaveen:
## Describe your changes
Add support for adding Heading block and Text Block to allow different fonts and sizes for the ticker.
In order to for the animation to work, we need a height to calculate the animation from. We can't derive it from the nested blocks so the best way I found was to add a setting for it Storyblok. 

After deploy we need to migrate the content in Storyblok before we remove the deprecated fields in code

#### Ticker with heading component
https://github.com/HedvigInsurance/racoon/assets/6661511/83fc23ed-40b1-495c-80e7-5432c36359d3

#### Ticker with text component
https://github.com/HedvigInsurance/racoon/assets/6661511/9be0eb3d-96b5-43c9-8778-8c2bdda0e871

For the new Home campaign we what to have a ticker animation of headings

### DIFF
--- a/apps/store/src/components/Ticker/Ticker.tsx
+++ b/apps/store/src/components/Ticker/Ticker.tsx
@@ -1,13 +1,23 @@
-import styled from '@emotion/styled'
-import type { Variants} from 'framer-motion';
+import clsx from 'clsx'
+import type { Variants } from 'framer-motion'
 import { AnimatePresence, motion } from 'framer-motion'
 import { type ReactNode, useEffect, useState, Children } from 'react'
-import { CheckIcon, Text, theme } from 'ui'
+import { sprinkles } from 'ui/src/theme/sprinkles.css'
+import type { FontSizeProps, FontSizes } from 'ui'
+import { CheckIcon } from 'ui'
+import { list, listItem, tickerItemWrapper } from './Ticker.css'
 
 const DURATION = 1
 
+export type TickerHeight = {
+  tickerHeight: FontSizes
+  tickerHeightDesktop: FontSizes
+}
+
 type Props = {
   children: Iterable<ReactNode>
+  showCheckIcon?: boolean
+  size: FontSizeProps
 }
 
 export const Ticker = (props: Props) => {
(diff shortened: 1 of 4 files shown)


gustaveen:
* **#4186** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4186?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

apps/store/src/components/Ticker/Ticker.tsx
```diff
@@ -23,67 +33,48 @@ export const Ticker = (props: Props) => {
...
           )
         })}
       </AnimatePresence>
-    </List>
+    </ul>
   )
 }
 
-const DRIFT_HEIGHT = '1.5rem'
+const DRIFT_HEIGHT = '0.8em'
```
gustaveen, line 60: Adapts to the current font size instead of being static

Youakeem: Nice!

apps/store/src/components/Ticker/Ticker.tsx
```diff
@@ -23,67 +33,48 @@ export const Ticker = (props: Props) => {
   }, [childrenCount])
 
   return (
-    <List>
+    <ul className={clsx(list, sprinkles({ fontSize: props.size }))}>
```
Youakeem, line 36: Just to make sure I got this right, we set the `fontSize` on the list and then use the `1.4em` to set the actual height, right?

If so, why not name the Storyblok field `fontSize` instead?

gustaveen: Good question @Youakeem ! So, we want the height to be the same as the font size of the text in HeadingBlock or TextBlock. The actual font size of the ticker is set by the HeadingBlock or TextBlock. 

I was a bit hesitant on wether to call if fontSize or height, since fontSize might give the impression to editors that it would change the actual size of the text.

But I will check with Peter as well what makes most sense to him 👍

Youakeem: I see! I did think the height will be used for the text. `height` might be a better name then. I would just add a little comment here explaining the logic.

Youakeem (approved):

PR #4188, opened 2024-04-25
Adjust past price intent offer date to today

Youakeem:


Youakeem:
* **#4188** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4188?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

guilhermespopolin (commented):
I wonder if we could solve this on the BE somehow. Otherwise I can think on other places we'd need to add just adjustment on the FE as well. Accident cross sell card for example.

Youakeem:
> I wonder if we could solve this on the BE somehow. Otherwise I can think on other places we'd need to add just adjustment on the FE as well. Accident cross sell card for example.
> 
> 

Sorry for the late reply. We've eventually decided to move the logic to the backend.

PR #4190, opened 2024-04-25
Explicitly pass locale to ApolloClient in app router

alebedev:
## Describe your changes

App router
- pass locale to `ApolloClient` creation function explicitly
- move Apollo provider so that it's in locale-aware part of layout
- show available product titles on test app router page

We broke language link in app router during some previous refactorings - client always used default language, which resulted in a mix of languages when showing content

### DIFF
--- a/apps/store/src/app/[locale]/test/page.tsx
+++ b/apps/store/src/app/[locale]/test/page.tsx
@@ -13,13 +13,19 @@ type Props = {
 const Page = async (props: Props) => {
   const { t } = await initTranslationsServerSide(props.params.locale)
   // Same graphql request as in layout, only one network request gets executed thanks to Apollo SSR cache
-  const apolloClient = await getApolloClient()
+  const apolloClient = getApolloClient(props.params.locale)
   const productMetadata = await fetchGlobalProductMetadata({ apolloClient })
 
-  console.log('productMetadata@page, items:', productMetadata.length)
   return (
     <div className={wrapper}>
       <h1>Server-side translation: {t('404_PAGE_MESSAGE')}</h1>
+      <br />
+      <h3>Available products</h3>
+      <ul>
+        {productMetadata.map((item) => (
+          <li key={item.id}>- {item.displayNameFull}</li>
+        ))}
+      </ul>
       <br />
       <Suspense>
         <ClientComponent />
--- a/apps/store/src/services/apollo/app-router/rscClient.ts
+++ b/apps/store/src/services/apollo/app-router/rscClient.ts
@@ -1,11 +1,39 @@
 import { ApolloClient, ApolloLink, InMemoryCache } from '@apollo/client'
 import { registerApolloClient } from '@apollo/experimental-nextjs-app-support/rsc'
-import { getLocaleOrFallback } from '@/utils/l10n/localeUtils'
+import { FALLBACK_LOCALE } from '@/utils/l10n/locales'
+import { toRoutingLocale } from '@/utils/l10n/localeUtils'
 import type { RoutingLocale } from '@/utils/l10n/types'
 import { errorLink } from '../errorLink'
 import { httpLink } from '../httpLink'
 import { userErrorLink } from '../userErrorLink'
 import { serverStaticHeadersLink } from './serverStaticHeadersLink'
+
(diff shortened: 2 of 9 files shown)


alebedev:
* **#4190** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4190?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

apps/store/src/services/apollo/app-router/rscClient.ts
```diff
@@ -1,12 +1,40 @@
...
+const { getClient } = registerApolloClient(() => {
+  return new ApolloClient({
+    name: 'Web:Racoon:Store',
+
+    link: ApolloLink.from([
+      // Has to be the first to process output last
+      // We re-raise userError results as errors, we don't want errorLink to see those
+      userErrorLink,
+      errorLink,
+      serverStaticHeadersLink({ locale: currentLocale }),
```
Youakeem, line 29: I don't understand why the previous implementation didn't work. We used `getLocaleOrFallback(params?.locale).routingLocale` which used `params` that were provided by Apollo.

I am not sure how Apollo get the locale, tho!

On a related note, I noticed we also have the `getUrlLocale`. Will this get us the correct locale in this case?

I guess I am trying to understand the dynamics between Apollo and locales.

apps/store/src/app/[locale]/test/page.tsx
```diff
@@ -13,14 +13,20 @@ type Props = {
 const Page = async (props: Props) => {
   const { t } = await initTranslationsServerSide(props.params.locale)
   // Same graphql request as in layout, only one network request gets executed thanks to Apollo SSR cache
-  const apolloClient = await getApolloClient()
+  const apolloClient = getApolloClient(props.params.locale)
```
Comment (Youakeem) [inline apps/store/src/app/[locale]/test/page.tsx:16]: When do we use the client provided by `ApolloProvider` vs the one from `getApolloClient`?

Comment (alebedev) [reply apps/store/src/app/[locale]/test/page.tsx:16]: Provider is for client components, `getApolloClient` is for servers-side code (components, actions, etc)

apps/store/src/services/apollo/app-router/rscClient.ts
alebedev: Old implementation did not work
1. It was created for earlier unstable version of apollo client
2. then we stopped passing locale (probably a bug)
3. we updated client library, now passing parameters to `getClient` started to throw

Hence new implementation. Can't say I'm happy with the way we pass locale to Apollo either here or in pages, but library API does not make it easy for us

apps/store/src/services/apollo/app-router/rscClient.ts
```diff
@@ -1,12 +1,40 @@
...
+// Passing locale through module-level static var is ugly, but it works
+// When `registerApolloClient` stops complaining about passed params, we should switch to it
+let currentLocale = toRoutingLocale(FALLBACK_LOCALE)
+export const getApolloClient = (locale: RoutingLocale): ApolloClient<any> => {
+  currentLocale = locale
+  return getClient()
+}
+
+// TODO: Support RSC client for dynamic pages too (auth headers, hedvig-shop-session-id header)
+const { getClient } = registerApolloClient(() => {
```
guilhermespopolin, line 20: Do we know or is it reasonable to assume that they will change this function so we can pass in the locale? 🤔

apps/store/src/app/[locale]/test/page.tsx
Comment (Youakeem) [reply apps/store/src/app/[locale]/test/page.tsx:16]: Yes, true. Thanks!

apps/store/src/services/apollo/app-router/rscClient.ts
alebedev: Let's watch the changelog, I don't have better suggestions right now

guilhermespopolin (approved):

Youakeem (approved):

alebedev:
### Merge activity

* **Apr 30, 2:04 AM EDT**: @alebedev started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4190).

PR #4197, opened 2024-04-29
Statically render only most popular pages

alebedev:
## Describe your changes

Manually list most popular pages and only pre-build those instead of all CMS pages

Makes build faster. Would be more important for app router since builds are slower there right now

Also would allow us to detect errors that break static rendering of product pages before merging into `main`


Youakeem (approved):
LGTM! 👍

PR #4204, opened 2024-04-30
GEN-2050 - feat(widget): enable tier selector for car

guilhermespopolin:
## Describe your changes

* Extract tiers/deductibles definition into a hook: `useTiersAndDeductibles`; `OfferPresenter` and widget's `ProductItemContainer` now uses it to get tiers and deductibles.
* `ProductItemContainer`/`ProductItem` refactoring: move presentational logic from the container to the item
* `ProductItem`/`ProductItemContainer`: updated so they support tiers selection 

https://github.com/HedvigInsurance/racoon/assets/19200662/3cdb8a72-6afe-49ec-9d25-44ecf2d717ee

Discussion points

1. We probably need some sort of loading state for when changing tiers/deductibles. Checking with WIlliam.
2. Observe that we're using current implementation/design for _Tier Selector_. The idea is to modify the UI to match latest designs later.
3. Observe that other parts of the page - like perils - doesn't get update when tiers are selected. That's gonna be added later.
4. Our current model is simple: once in sign/checkout/cart pages, the source of truth is whatever offer you have into your shop session cart. Adding "editing" at the this point means needing to have access to the price intent(s) that generate those offers so we can replace what you have in your cart. _Hedvig Embeded_ is a subset of Hedvig-dot-com by consisting of a linear flow flow purchasing a single product, so we need to handle with only one price intent. But things could be complicated if more than one product is in the cart...

First step towards enabling car+pet for Hedvig Embeded.

### DIFF
--- a/apps/store/src/components/ProductPage/PurchaseForm/useTiersAndDeductibles.ts
+++ b/apps/store/src/components/ProductPage/PurchaseForm/useTiersAndDeductibles.ts
@@ -1,0 +1,30 @@
+import { useMemo } from 'react'
+import { type ProductOfferFragment } from '@/services/graphql/generated'
+import { getOffersByPrice } from '@/utils/getOffersByPrice'
+
+type Params = {
+  offers: Array<ProductOfferFragment>
+  selectedOffer: ProductOfferFragment
+}
+
+export function useTiersAndDeductibles({ offers, selectedOffer }: Params) {
+  return useMemo(() => {
+    const sortedOffers = getOffersByPrice(offers)
+
+    const tiers: Array<ProductOfferFragment> = []
+    const usedTiers = new Set<string>()
+    for (const offer of sortedOffers) {
+      const typeOfContract = offer.variant.typeOfContract
+      if (usedTiers.has(typeOfContract)) continue
+
+      usedTiers.add(typeOfContract)
+      tiers.push(typeOfContract === selectedOffer.variant.typeOfContract ? selectedOffer : offer)
+    }
+
+    const deductibles = sortedOffers.filter(
+      (item) => item.variant.typeOfContract === selectedOffer.variant.typeOfContract,
+    )
+
+    return { tiers, deductibles }
+  }, [offers, selectedOffer])
+}
--- a/apps/store/src/features/widget/ConfirmationPage.tsx
+++ b/apps/store/src/features/widget/ConfirmationPage.tsx
@@ -10,7 +10,7 @@ import { useDiscountProps } from '@/components/ShopBreakdown/useDiscountExplanat
 import type { ShopSession } from '@/services/shopSession/ShopSession.types'
 import type { ConfirmationStory } from '@/services/storyblok/storyblok'
 import { Header } from './Header'
-import { ProductItemContainer } from './ProductItemContainer'
+import { ProductItem } from './ProductItem'
 import { publishWidgetEvent } from './publishWidgetEvent'
 
 type Props = {
@@ -39,7 +39,12 @@ export const ConfirmationPage = (props: Props) => {
             </Heading>
             <Space y={1}>
               {props.shopSession.cart.entries.map((item) => (
-                <ProductItemContainer key={item.id} offer={item} disableStartDate={true} />
+                <ProductItem
+                  key={item.id}
+                  shopSessionId={props.shopSession.id}
+                  selectedOffer={item}
+                  disableStartDate={true}
+                />
               ))}
 
               {discount && (
--- a/apps/store/src/features/widget/ProductItem.tsx
+++ b/apps/store/src/features/widget/ProductItem.tsx
@@ -1,10 +1,8 @@
-// This is a copy of @/components/ProductItem/ProductItem
-// We will to merge this file with the original one
-
+import { datadogRum } from '@datadog/browser-rum'
 import styled from '@emotion/styled'
 import { useTranslation } from 'next-i18next'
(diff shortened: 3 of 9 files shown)


guilhermespopolin:
* **#4204** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4204?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

apps/store/src/components/ProductPage/PurchaseForm/useTiersAndDeductibles.ts
```diff
@@ -0,0 +1,31 @@
...
+import { type ProductOfferFragment } from '@/services/graphql/generated'
+import { getOffersByPrice } from '@/utils/getOffersByPrice'
+
+type Params = {
+  offers: Array<ProductOfferFragment>
+  selectedOffer: ProductOfferFragment
+}
+
+export function useTiersAndDeductibles({ offers, selectedOffer }: Params) {
+  const sortedOffers = useMemo(() => getOffersByPrice(offers), [offers])
```
alebedev, line 11: `sortedOffers` will update every time input `offers` change. You can simplify the structure by doing everything inside single `useMemo` instead 3 different ones.  The only thing you're giving away is price of recalculation when `selectedOffer` changes, don't know how significant that is

apps/store/src/features/widget/ConfirmationPage.tsx
```diff
@@ -39,7 +39,14 @@ export const ConfirmationPage = (props: Props) => {
             </Heading>
             <Space y={1}>
               {props.shopSession.cart.entries.map((item) => (
-                <ProductItemContainer key={item.id} offer={item} disableStartDate={true} />
+                <ProductItem
+                  key={item.id}
+                  shopSessionId={props.shopSession.id}
+                  selectedOffer={item}
+                  deductibles={[]}
```
alebedev, line 46: Is this mergeable with hardcoded empty arrays here?

apps/store/src/features/widget/ProductItem.tsx
```diff
@@ -145,6 +195,13 @@ export const ProductItem = (props: Props) => {
   )
 }
 
+const getTierLevelDisplayName = (item: Pick<ProductOfferFragment, 'variant' | 'product'>) => {
+  // TODO: small hack, move logic to API
```
alebedev, line 199: Good point! Can you create a ticket if there isn't one alread?

alebedev (approved):
Good job!  Some minor comments inside

apps/store/src/components/ProductPage/PurchaseForm/useTiersAndDeductibles.ts
```diff
@@ -0,0 +1,31 @@
...
+  const tiers = useMemo(() => {
+    const tierList: Array<ProductOfferFragment> = []
+    const usedTiers = new Set<string>()
+    for (const offer of sortedOffers) {
+      const typeOfContract = offer.variant.typeOfContract
+      if (usedTiers.has(typeOfContract)) continue
```
Youakeem, line 17: Can we have multiple offers with the same `typeOfContract`? and If we do, do we really want to just take the first one?

apps/store/src/features/widget/ConfirmationPage.tsx
guilhermespopolin: Yes, but now thinking about it it's better to make those props optional. Gonna update it.

apps/store/src/features/widget/ProductItem.tsx
guilhermespopolin: Yes. I created this [one](https://www.notion.so/hedviginsurance/Storefront-add-tier-name-display-name-c27d9d8a49194e57ab0e3fac50961d5b?pvs=4)
It's currently part of _Enable car on Widget_ project. Let me know if I should move it somewhere else.

apps/store/src/components/ProductPage/PurchaseForm/useTiersAndDeductibles.ts
guilhermespopolin: Yes, for both questions. For this to work is required sorting the offers by price (ascending), which is the first thing we do. Now about your questions, take the following example:

image

As you can see, we have two deductibles options for Basic coverage. That means will have two offers with SE_BASIC as their type of contract, but at the end we're dealing with one tier with two deductibles options,

apps/store/src/components/ProductPage/PurchaseForm/useTiersAndDeductibles.ts
guilhermespopolin: Have updated it

guilhermespopolin:
### Merge activity

* **May 2, 5:17 AM EDT**: @guilhermespopolin started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4204).

apps/store/src/components/ProductPage/PurchaseForm/useTiersAndDeductibles.ts
Youakeem: I see, but we are sorting by price and only keeping the first one, how do we make sure we keep the right one?

guilhermespopolin: Aha, I see. Yes, since we're sorting by price (ascending) the first one for each type of contract (tier) is the one with the lowest deductible option, or at the list that's how I understood it.

PR #4208, opened 2024-05-02
Convert ButtonBlock styles to vanilla-extract and clean code a little

alebedev:
## Describe your changes

- Convert ButtonBlock wrapper style to vanilla-extract
- Removed ConditionalWrapper, code is easier to read without it

### DIFF
--- a/apps/store/src/blocks/HeroBlock.tsx
+++ b/apps/store/src/blocks/HeroBlock.tsx
@@ -1,12 +1,16 @@
 'use client'
 
 import styled from '@emotion/styled'
-import type { SbBlokData} from '@storyblok/react';
+import type { SbBlokData } from '@storyblok/react'
 import { StoryblokComponent, storyblokEditable } from '@storyblok/react'
-import type { ButtonBlockProps } from '@/blocks/ButtonBlock';
+import type { ButtonBlockProps } from '@/blocks/ButtonBlock'
 import { ButtonBlock } from '@/blocks/ButtonBlock'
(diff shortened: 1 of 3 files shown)


apps/store/src/blocks/HeroBlock.tsx
```diff
@@ -23,7 +27,7 @@ type HeroBlockProps = SbBaseBlockProps<{
 }>
 
 export const HeroBlock = ({ blok }: HeroBlockProps) => {
-  const buttonBlocks = filterByBlockType(blok.buttons, ButtonBlock.blockName)
+  const buttonBlocks = filterByBlockType(blok.buttons, 'button')
```
guilhermespopolin, line 30: It's sad that app router doesn't work with `.blockName`. I image this is going to be easy to forget in case we change button block's name 😢

guilhermespopolin (approved):
Nice one :)

apps/store/src/blocks/HeroBlock.tsx
alebedev: Maybe we can keep it as is for 80% of cases and only worry about exceptions where `filterByBlockType` and `blockName` are across client - server boundary.  I'll revert this change here, since both are client components

alebedev:
### Merge activity

* **May 2, 4:24 AM EDT**: [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4208) rebased this pull request after merging its parent, because this pull request is set to merge when ready.

PR #4213, opened 2024-05-03
feature: [RND-704] Move Swedish homepage to hedvig.com root

Youakeem:
## Describe your changes
Serve the Swedish homepage under hedvig.com instead of hedvig.com/se.

All internal pages still use the `/se` URL segment.

Requested by Peter for branding and SEO reasons.

### DIFF
--- a/apps/store/src/middleware.ts
+++ b/apps/store/src/middleware.ts
@@ -4,6 +4,7 @@ import { NextResponse } from 'next/server'
 import { experimentMiddleware } from '@/services/Tracking/experimentMiddleware'
 import { FALLBACK_LOCALE, LOCALE_COOKIE_KEY, locales } from '@/utils/l10n/locales'
 import { isRoutingLocale } from '@/utils/l10n/localeUtils'
+import type { RoutingLocale } from '@/utils/l10n/types'
 
 export const config = {
   matcher: [
@@ -35,6 +36,20 @@ const localeMiddleware = (req: NextRequest): NextResponse | undefined => {
   const targetUrl = req.nextUrl.clone()
   const firstSegment = targetUrl.pathname.split('/')[1]
 
+  const handleLocaleRouting = (locale: RoutingLocale) => {
+    targetUrl.pathname = `/${locale}${targetUrl.pathname}`
+
+    // Rewrite homepage response to default locale
+    if (locale === locales[FALLBACK_LOCALE].routingLocale) {
+      // Rewrite is like a "behind the scenes" redirect
+      // It displays the given URL without redirecting visitors
+      // https://nextjs.org/docs/app/building-your-application/routing/middleware#nextresponse
+      return NextResponse.rewrite(targetUrl)
+    }
+
+    return NextResponse.redirect(targetUrl, 308)
+  }
+
   // Localized route
   if (isRoutingLocale(firstSegment)) {
     const isSwedishHomepage = targetUrl.pathname === '/se'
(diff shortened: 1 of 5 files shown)


Youakeem:
* **#4213** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4213?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

apps/store/src/middleware.ts
```diff
@@ -34,11 +34,46 @@ export async function middleware(req: NextRequest) {
...
   const url = new URL(req.url)
   const firstSegment = url.pathname.split('/')[1]
-  if (isRoutingLocale(firstSegment)) return
+
+  // Localized route
+  if (isRoutingLocale(firstSegment)) {
+    const isSwedish = firstSegment === locales['sv-SE'].routingLocale
+
+    // URL only have a locale
+    const isHomePage = url.pathname.split('/').length === 2
```
alebedev, line 43: Can we express it more directly? Something like `url.pathname === '/se'`?

apps/store/src/middleware.ts
```diff
@@ -34,11 +34,46 @@ export async function middleware(req: NextRequest) {
...
   const redirectToLocale = (locale: string) => {
     const targetUrl = req.nextUrl.clone()
     targetUrl.pathname = `/${locale}${targetUrl.pathname}`
+
+    // This function is invoked when requesting a non-localized route
+    // To redirect it to the relevant localized version
+    // When called with a the Swedish or default locales
+    // We rewrite the URL instead of redirecting
+    // To display the translated pages without changing the URL
+    if (
```
alebedev, line 70: I'm confused - why do we want this rewrite? Why are we doing a rewrite inside function called `redirect`, isn't this misleading?

apps/store/src/middleware.ts
Youakeem: Good idea! That could replace the two checks.

apps/store/src/middleware.ts
Youakeem: We rewrite to keep the URL the same, without appending `/se`. Good point about using the naming. A rewrite still "redirects" traffic behind the scenes to a specific next route.

What do you think of renaming this function to something like `handleLocaleRouting` ?

Youakeem: I've restructured this part as discussed 👍

apps/store/src/middleware.ts
Youakeem: Done!

apps/store/src/middleware.ts
Youakeem: So updating it caused an infinite loop in case the user has a Swedish preferred language cookie. The cookie redirected to `/se` and then `/se` was redirected back to `/` for the cookie logic to kick in again.

By having the routing logic in one place, it doesn't matter what sets the language to `/se`.

gustaveen:
Good work! Just wondering about if we should still do a redirect instead of a rewrite when locale is missing but if Peter thinks it's fine I'm all good since it makes the code simpler 👍

apps/store/src/utils/PageLink.ts
```diff
@@ -349,3 +353,19 @@ const REVIEWS_URL: Partial<Record<RoutingLocale, URL>> = {
...
 }
+
+export const removeTrailingSlash = (url: string) => {
+  return url.endsWith('/') ? url.slice(0, -1) : url
+}
+
+export const removeSEHomepageLangSegment = (urlString: string): string => {
+  const url = new URL(urlString)
+
+  const isSwedishHomepage = removeTrailingSlash(url.pathname) === '/se'
```
guilhermespopolin, line 364: Suggestion: Would the following work as well?
I found `removeTrailingSlash(url.pathname) === '/se'` line a little bit confusing because in theory `removeTrailingSlash` expects a full URL. But maybe it's just me :)

```suggestion
  const url = new URL(removeTrailingSlash(urlString))

  const isSwedishHomepage = url.pathname === '/se'
```

guilhermespopolin (approved):
Nice work. Just left a small comment.

Youakeem:
> Good work! Just wondering about if we should still do a redirect instead of a rewrite when locale is missing but if Peter thinks it's fine I'm all good since it makes the code simpler 👍 

Not sure I am following. Where should we redirect? When the locale is missing we just want to serve the page and we are serving the Swedish page through rewriting the response to `/se`.

apps/store/src/middleware.ts
```diff
@@ -32,24 +34,60 @@ export async function middleware(req: NextRequest) {
...
-    const targetUrl = req.nextUrl.clone()
-    targetUrl.pathname = `/${locale}${targetUrl.pathname}`
-    return NextResponse.redirect(targetUrl, 308)
+  const handleLocaleRouting = (locale: RoutingLocale) => {
+    const targetURL = req.nextUrl.clone()
+
+    // Append locale to the pathname
+    targetURL.pathname = removeTrailingSlash(`/${locale}${targetURL.pathname}`)
+
+    // If are targeting the default language homepage
```
gustaveen, line 46:
```suggestion
    // If we are targeting the default language homepage
```

gustaveen (approved):
Let's go! 🙌

PR #4211, opened 2024-05-02
GEN-2063 - refact: removed old widget switch feature flag

guilhermespopolin:
## Describe your changes

* Remove "switching page" feature flag. Additionally, make sure only home insurance makes usage of it.

## Justify why they are needed

This is a foundation change. Starting with car, switching will be done in the sign page making "switching page" obsolete.

image

PR #4219, opened 2024-05-03
GEN-2063 - refact(ProductItem): migrate to vanilla extract css

guilhermespopolin:
## Describe your changes

* Refact: migrate widget `ProducItem` styles into `vanilla-extract`.

### DIFF
--- a/apps/store/src/features/widget/ProductItem.css.ts
+++ b/apps/store/src/features/widget/ProductItem.css.ts
@@ -1,0 +1,102 @@
+import { style } from '@vanilla-extract/css'
+import { minWidth, theme } from 'ui/src/theme'
+
+export const hoverable = style({
+  ':hover': {
+    cursor: 'pointer',
+  },
+})
+
+export const card = style({
+  borderRadius: theme.radius.md,
+  padding: theme.space.md,
+  paddingBottom: 0,
+  backgroundColor: theme.colors.opaque1,
+  '@media': {
+    '(hover: hover)': {
+      selectors: {
+        [`&:has(${hoverable}:hover)`]: {
+          backgroundColor: theme.colors.grayTranslucent200,
+        },
+      },
+    },
+    [minWidth.lg]: {
+      padding: theme.space.lg,
+      paddingBottom: `calc(${theme.space.lg} - ${theme.space.md})`,
+    },
+  },
+})
+
+export const cardGreenVariant = style({
+  backgroundColor: theme.colors.signalGreenFill,
+  '@media': {
+    '(hover: hover)': {
+      selectors: {
+        [`&:has(${hoverable}:hover)`]: {
+          backgroundColor: theme.colors.green200,
+        },
+      },
+    },
+  },
+})
+
+export const cardHeader = style({
+  display: 'grid',
+  columnGap: theme.space.md,
+  alignItems: 'center',
+  paddingBottom: theme.space.md,
+})
+
+export const cardHeaderRow = style({
+  display: 'flex',
+  justifyContent: 'space-between',
+  alignItems: 'center',
+  width: '100%',
+})
+
+export const cardFooter = style({
+  display: 'grid',
+  gridAutoFlow: 'column',
+  columnGap: theme.space.xs,
+  paddingBottom: theme.space.md,
+})
+
+export const deleteButton = style({
+  display: 'flex',
+  justifyContent: 'center',
+  alignItems: 'center',
+  width: '1.5rem',
+  aspectRatio: '1 / 1',
+  borderRadius: '50%',
+  backgroundColor: theme.colors.grayTranslucent200,
+  cursor: 'pointer',
+  '@media': {
+    '(hover: hover)': {
+      ':hover': {
+        backgroundColor: theme.colors.grayTranslucent300,
+      },
+    },
+  },
+})
+
+export const fakeInput = style({
+  width: '100%',
+  height: '4.5rem',
+  padding: theme.space.sm,
+  borderRadius: theme.radius.sm,
+  backgroundColor: theme.colors.translucent1,
+})
+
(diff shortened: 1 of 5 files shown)


guilhermespopolin:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4219?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#4222** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4222?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4221** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4221?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4220** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4220?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4219** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4219?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#4211** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4211?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

apps/store/src/features/widget/ProductItem.css.ts
```diff
@@ -0,0 +1,102 @@
...
+})
+
+export const card = style({
+  position: 'relative',
+  borderRadius: theme.radius.md,
+  padding: theme.space.md,
+  paddingBottom: 0,
+  backgroundColor: theme.colors.opaque1,
+  selectors: {
+    '&[data-variant="green"]': {
```
Youakeem, line 19: Interesting use of data attributes! Are we not able to use `styleVariants` here?

apps/store/src/features/widget/ProductItem.css.ts
```diff
@@ -0,0 +1,102 @@
...
+  display: 'flex',
+  justifyContent: 'center',
+  alignItems: 'center',
+  width: '1.5rem',
+  aspectRatio: '1 / 1',
+  borderRadius: '50%',
+  backgroundColor: theme.colors.grayTranslucent200,
+  cursor: 'pointer',
+  '@media': {
+    '(hover: hover)': {
```
Youakeem, line 73: I am not sure I understand why we need `@media` and `(hover: hover)`. If we are after hover styles then shouldn't`:hover` be enough?

guilhermespopolin: Yeah, this is a nice detail but to be honest it doesn't add too much gain on this particular case. Here's some context:

> **Interaction Media Queries**
Apparently there's those media queries out there that one can use to apply styles based on the input mechanism the user is using. They're called [Interaction Media Features](https://drafts.csswg.org/mediaqueries-4/#mf-interaction). They can be used to improve the effectives of hover styles. A general issue with hover  styles is that they usually doesn't make sense when your app is being accessed by a mobile device because, you know, there's no "hover" on a touchscreen. As a result, on mobile, your hover styles get applied when you tap an interactive element which might not be ideal. With that said, you can use those new media features to apply hover styles only when they make sense:
>
> `@media (hover: hover) and (pointer: fine) {
   /* Your hover styles here */
}`
>
>The hover  media feature refers to the ability of a device to move the cursor without also triggering a click/tap on the element. hover means that's possible and none , the opposite. Additionally, the pointer media feature can assume the values fine | coarse | none and it refers on how precise the users can be with their clicks. E.g: mouse --> fine; touchscreen --> coarse; keyboard navigation --> none.
>
> While relatively new they're broadly [supported by browsers](https://caniuse.com/css-media-interaction). (edited)

Take this [example](https://codepen.io/guilhermespopolin/full/LYvwdyp). Both buttons have the same hover styles. However, when opening that page on a phone (browser responsive move is fine), you'll see hover styles are applied while taping the second button, but not the first one. That's because the first one add its hover styles under a `@media (hover: hover)` media query.

apps/store/src/features/widget/ProductItem.css.ts
guilhermespopolin: Yes, that would also work. I can modify it.

apps/store/src/features/widget/ProductItem.css.ts
Youakeem: Got it! Today I learned something new, thanks, man!

apps/store/src/features/widget/ProductItem.css.ts
guilhermespopolin: It's updated.

Youakeem (approved):

apps/store/src/features/widget/ProductItem.css.ts
```diff
@@ -0,0 +1,105 @@
...
+      ':hover': {
+        backgroundColor: theme.colors.grayTranslucent300,
+      },
+    },
+  },
+})
+
+export const fakeInput = style({
+  position: 'relative',
```
alebedev, line 85: Noticed a few `  position: 'relative',` here, but not `position: absolute` on child elements.  Do we still need `relative` part for something inside?

guilhermespopolin: That's true. Removing it.

guilhermespopolin:
### Merge activity

* **May 6, 8:28 AM EDT**: @guilhermespopolin started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4219).

PR #4226, opened 2024-05-06
Convert ImageWithPlaceholder styles to vanilla-extract

alebedev:
## Describe your changes

Subj

This was part of debugging NextJs issue that looks useful on its own

### DIFF
--- a/apps/store/src/components/ImageWithPlaceholder/ImageWithPlaceholder.tsx
+++ b/apps/store/src/components/ImageWithPlaceholder/ImageWithPlaceholder.tsx
@@ -1,26 +1,30 @@
-import styled from '@emotion/styled'
-import type { ImageProps } from 'next/image';
+import { clsx } from 'clsx'
+import type { ImageProps } from 'next/image'
 import { default as NextImage } from 'next/image'
+import type { SyntheticEvent } from 'react'
 import { useState } from 'react'
-import { theme, DEFAULT_IMAGE_QUALITY } from 'ui'
+import { DEFAULT_IMAGE_QUALITY } from 'ui'
+import { imageStyles } from './ImageWithPlaceholder.css'
 
-export const ImageWithPlaceholder = ({ ...props }: ImageProps) => {
+export const ImageWithPlaceholder = ({
+  quality,
+  className,
+  onLoad,
+  ...forwardedProps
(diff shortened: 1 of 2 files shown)


apps/store/src/components/ImageWithPlaceholder/ImageWithPlaceholder.tsx
```diff
@@ -1,26 +1,30 @@
...
-    ...props,
-    ...(hasLoaded && { 'data-loaded': true }),
-  } as const
-
-  return <StyledImage {...imageProps} onLoad={handleLoad} />
+  return (
+    <NextImage
+      {...forwardedProps}
+      quality={quality ?? DEFAULT_IMAGE_QUALITY}
+      className={clsx(imageStyles.base, hasLoaded && imageStyles.loaded, className)}
```
gustaveen, line 26: Just to follow the logic of variants it would be nice to only apply the desired variant class I think :) 


```suggestion
  const classNames = clsx(hasLoaded ? imageStyles.loaded : imageStyles.base, className)

  return (
    <NextImage
         className={classNames}
```

gustaveen (approved):
Great! 🙌 Just a comment

apps/store/src/components/ImageWithPlaceholder/ImageWithPlaceholder.tsx
```diff
@@ -1,26 +1,30 @@
...
-    ...props,
-    ...(hasLoaded && { 'data-loaded': true }),
-  } as const
-
-  return <StyledImage {...imageProps} onLoad={handleLoad} />
+  return (
+    <NextImage
+      {...forwardedProps}
+      quality={quality ?? DEFAULT_IMAGE_QUALITY}
+      className={clsx(hasLoaded ? imageStyles.loaded : imageStyles.base, className)}
```
Youakeem, line 26: Since `loaded` seems like a boolean variant. and we are using `clsx` to add it. How about we separate the styles into separate `style` functions?

https://www.notion.so/hedviginsurance/Design-systems-with-vanilla-extract-1cf3214204d6496bad3d4326666f4bca

Youakeem (approved):

guilhermespopolin (approved):

PR #4228, opened 2024-05-06
Report deviceInfo event to internal analytics

alebedev:
## Describe your changes

Report `deviceInfo` even once when session is initialized

First part of behavior-based pricing


apps/store/src/services/Tracking/Tracking.ts
```diff
@@ -28,11 +30,15 @@ type TrackingOffer = {
...
 
+// Naming rules:
+// - snake_case for Analytics / ecommerce events (backward compatibility)
+// - camelCase for internal anayltics events
 export enum TrackingEvent {
   AddToCart = 'add_to_cart',
   Adtraction = 'adtraction',
   BeginCheckout = 'begin_checkout',
   DeleteFromCart = 'delete_from_cart',
+  DeviceInfo = 'deviceInfo',
```
gustaveen, line 41: So `deviceInfo` is sent to GTM as well?

apps/store/src/services/Tracking/Tracking.ts
```diff
@@ -68,6 +74,21 @@ export class Tracking {
     initializeGtm(countryCode)
   }
 
+  #reportedDeviceInfo = false
```
gustaveen, line 77: Why do we use `#`?

apps/store/src/services/Tracking/internalEvents.ts
```diff
@@ -0,0 +1,39 @@
...
+import { useShopSession } from '@/services/shopSession/ShopSessionContext'
+import type { TrackingEvent } from '@/services/Tracking/Tracking'
+import { isBrowser } from '@/utils/env'
+
+type EventData = {
+  type: TrackingEvent
+  data: unknown
+}
+
+const eventBuffer: Array<Omit<EventInput, 'sessionId'>> = []
```
gustaveen, line 13: Could we add a comment on what `internalEvents` are?

apps/store/src/services/Tracking/internalEvents.ts
```diff
@@ -0,0 +1,39 @@
...
+export const reportInternalEvent = (event: EventData) => {
+  if (!isBrowser()) throw new Error('Must be used client-side')
+  eventBuffer.push({
+    id: crypto.randomUUID(),
+    clientTimestamp: new Date().toISOString(),
+    ...event,
+  })
+}
+
+export const useFlushInternalEvents = () => {
```
gustaveen, line 23: I'm slightly confused with `flush` in the name :)
```suggestion
export const useSendInternalEvents = () => {
```

gustaveen (approved):
Looks good! Just had some minor things I didn't catch

apps/store/src/services/Tracking/internalEvents.ts
Youakeem, line 23: The term `flush` makes me think of removing or resetting the events instead of submitting or reporting them.

apps/store/src/services/Tracking/useReportAndFlushDeviceInfo.ts
```diff
@@ -0,0 +1,16 @@
...
+import { useFlushInternalEvents } from '@/services/Tracking/internalEvents'
+import { useTracking } from '@/services/Tracking/useTracking'
+
+export const useReportAndFlushDeviceInfo = () => {
+  const { onReady } = useShopSession()
+  const tracking = useTracking()
+  const flushInternalEvents = useFlushInternalEvents()
+  useEffect(() => {
+    return onReady(() => {
+      tracking.reportDeviceInfo()
```
Youakeem, line 12: Why do we need the events array if we are submitting directly after tracking?

guilhermespopolin (approved):
Just curious: why to keep a buffer of internal events?

apps/store/src/services/Tracking/Tracking.ts
alebedev: No, only internally. GTM has browser info automatically, so no need to post it there

apps/store/src/services/Tracking/internalEvents.ts
```diff
@@ -0,0 +1,39 @@
...
+  const shopSessionId = useShopSession().shopSession?.id
+  const [sendEventBatch] = useSendEventBatchMutation({
+    onError() {
+      console.log('Failed to send internal events, ignoring the error')
+    },
+  })
+  return useCallback(() => {
+    if (shopSessionId == null) return
+    const inputList = eventBuffer.splice(0).map((item) => ({
```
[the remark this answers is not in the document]
alebedev: Equivalent of "save copy of original items, clear array content, then process original items"

apps/store/src/services/Tracking/useReportAndFlushDeviceInfo.ts
alebedev: This callback will fire
- later, when shopSession is ready (i.e. created or fetched)
- potentially multiple times

apps/store/src/services/Tracking/Tracking.ts
alebedev: native JS/TS private properties.  Could be written as `private reportedDeviceInfo = false` with the same effect

apps/store/src/services/Tracking/useReportAndFlushDeviceInfo.ts
Youakeem: But both of them run inside of `onReady` so invoking it will register/track and then immediately send, no?

alebedev: Good catch, I was referring to a limitation of original version that's no longer there

apps/store/src/services/Tracking/internalEvents.ts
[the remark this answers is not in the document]
alebedev: Not needed in new version of code, so let's shelve this for now. I don't see a better way to solve this particular task in JS. Not our fault that `Array.prototype` methods don't have clear enough names

guilhermespopolin (approved):
Well done. Very straightforward now :)

apps/store/src/services/shopSession/ShopSessionContext.tsx
```diff
@@ -68,8 +59,18 @@ const useShopSessionContextValue = (initialShopSessionId?: string) => {
   const callbacksRef = useRef(new Set<OnReadyCallback>())
   queryResult.onReady = useCallback(
     (callback) => {
-      if (queryResult.shopSession) {
-        callback(queryResult.shopSession)
+      const shopSession = queryResult.shopSession
+      if (shopSession) {
+        // Prevent duplicate invocations in dev mode
+        setTimeout(() => {
+          if (callbacksRef.current.has(callback)) {
```
Youakeem, line 66: Not sure if I understand correctly but since `callback` is a function, are we should that it'll maintain it's referential equality for the `has` to work as expected?

PR #4238, opened 2024-05-07
Migrate Video component to vanilla-extract

gustaveen:
## Describe your changes
Migrate Video component to vanilla-extract

https://github.com/HedvigInsurance/racoon/assets/6661511/774141b6-ca33-45e9-86d2-9bfb2397a724

Migrate to vanilla-extract and get rid of visual button bug

### DIFF
--- a/apps/store/src/components/Video/Video.css.ts
+++ b/apps/store/src/components/Video/Video.css.ts
@@ -1,0 +1,139 @@
+import { createVar, keyframes, style } from '@vanilla-extract/css'
+import { hoverStyles, minWidth, theme } from 'ui'
+
+export const videoWrapper = style({
+  position: 'relative',
+})
+
+export const heightPortraitVar = createVar()
+export const maxHeightPortraitVar = createVar()
+export const aspectRatioPortraitVar = createVar()
+export const heightLandscapeVar = createVar()
+export const maxHeightLandscapeVar = createVar()
+export const aspectRatioLandscapeVar = createVar()
+
+export const videoBase = style({
+  width: '100%',
+  objectFit: 'cover',
+  '@media': {
+    '(orientation: portrait)': {
+      height: heightPortraitVar,
+      maxHeight: maxHeightPortraitVar,
+      aspectRatio: aspectRatioPortraitVar,
+    },
+    '(orientation: landscape)': {
+      aspectRatio: aspectRatioLandscapeVar,
+      height: heightLandscapeVar,
+      maxHeight: maxHeightLandscapeVar,
+    },
+  },
+})
+
+export const videoRoundedCorners = style({
+  borderRadius: theme.radius.md,
+
+  '@media': {
+    [minWidth.lg]: {
+      borderRadius: theme.radius.xl,
+    },
+  },
+})
+
+export const videoControls = style({
+  position: 'absolute',
+  inset: 0,
+  display: 'flex',
+  flexDirection: 'column',
+  padding: theme.space.md,
+  justifyContent: 'flex-end',
+  alignItems: 'flex-start',
+  cursor: 'pointer',
+})
+
+export const videoControlsVisibility = style({
+  display: 'flex',
+  gap: theme.space.xs,
+  width: '100%',
+
+  '@media': {
+    '(hover: hover)': {
+      opacity: 0,
+      visibility: 'hidden',
+      transition: 'all 200ms cubic-bezier(0, 0, 0.2, 1) 2s',
+      selectors: {
+        [`${videoControls}[data-state="paused"] > &, ${videoControls}:hover > &`]: {
+          opacity: 1,
+          visibility: 'visible',
+          transitionDelay: '0s',
+        },
+      },
+    },
+  },
+})
+
+export const controlButton = style({
+  display: 'inline-flex',
+  alignItems: 'center',
+  gap: theme.space.xs,
+  height: '2rem',
+  paddingInline: theme.space.xs,
+  backgroundColor: theme.colors.grayTranslucentDark700,
+
+  ...hoverStyles({
+    backgroundColor: theme.colors.grayTranslucentDark600,
+  }),
+
+  ':active': {
+    backgroundColor: theme.colors.grayTranslucentDark600,
+  },
+})
+
+export const soundBarsAnimation = keyframes({
+  '50%': { opacity: 0.2, transform: 'scaleY(.2)' },
+  '100%': { opacity: 1, transform: 'scaleY(0.9)' },
+})
+
+export const soundBars = style({
+  position: 'relative',
+  display: 'flex',
+  alignItems: 'center',
+  justifyContent: 'center',
+  height: '16px',
+  width: '16px',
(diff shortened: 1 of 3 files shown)


gustaveen:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4238?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#4238** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4238?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#4237** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4237?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

apps/store/src/components/Video/Video.css.ts
```diff
@@ -0,0 +1,138 @@
...
+  animation: `${soundBarsAnimation} 1s 2 alternate`,
+
+  selectors: {
+    '&:first-of-type': { transform: 'scaleY(.85)', animationDelay: '.4s' },
+    '&:nth-of-type(2)': { transform: 'scaleY(.43)', animationDelay: '.2s' },
+    '&:nth-of-type(3)': { transform: 'scaleY(.72)', animationDelay: '.6s' },
+
+    [`${videoControls}[data-muted=false]${videoControls}[data-state="playing"] &`]: {
+      animation: `${soundBarsAnimation} 1s infinite alternate`,
+    },
```
gustaveen, line 123: I'm a bit sad vanilla-extract don't support this type of nesting
```suggestion
    [`${videoControls}[data-muted=false]${videoControls}[data-state="playing"] &`]: {
      animation: `${soundBarsAnimation} 1s infinite alternate`,
      '&:nth-of-type(1)': { transform: 'scaleY(.85)', animationDelay: '.4s' },
      '&:nth-of-type(2)': { transform: 'scaleY(.43)', animationDelay: '.2s' },
      '&:nth-of-type(3)': { transform: 'scaleY(.72)', animationDelay: '.6s' },
    },
```

apps/store/src/components/Video/Video.tsx
```diff
@@ -177,23 +195,34 @@ export const Video = ({
...
         onEnded={handleVideoEnded}
         muted={muted}
         {...autoplayAttributes}
         {...delegated}
+        style={{
+          ...(poster && {
+            backgroundImage: `url(${poster})`,
+            backgroundRepeat: 'no-repeat',
+            backgroundSize: 'cover',
+          }),
```
gustaveen, line 216: What do you feel like in these situations where we apply multiple rules based on a prop. We could use CSS vars but this feels easier to read IMO

apps/store/src/components/Video/Video.tsx
```diff
@@ -177,23 +195,34 @@ export const Video = ({
...
+          ...assignInlineVars({
+            [heightPortraitVar]: aspectRatioPortrait === '100' ? '100vh' : null,
```
gustaveen, line 218: Make sure we sent no value if prop is not specified

apps/store/src/components/Video/Video.tsx
alebedev: Looks good to me in this case.  We're forced to use inline styles for `backgroundImage` anyway

apps/store/src/components/Video/Video.tsx
alebedev, line 218: Can this be simplified by making `fullSizePortrait` and `fullSizeLandscape` into special cases with their own classes?

apps/store/src/components/Video/Video.tsx
```diff
@@ -167,8 +183,10 @@ export const Video = ({
     }
   }
 
+  const classNames = clsx(videoBase, roundedCorners && videoRoundedCorners, poster)
```
alebedev, line 186: Any reason not to inline this? I find inline className values are almost always easier to read

apps/store/src/components/Video/Video.tsx
```diff
@@ -203,15 +232,17 @@ export const Video = ({
...
-        <VideoControls
+        <div
+          className={videoControls}
           data-state={state}
           data-muted={muted}
           onClick={() => playPauseButtonRef.current?.click()}
         >
-          <Controls>
-            <ControlButton
+          <div className={controls}>
```
alebedev, line 243: Do we need extra div wrapper here? Maybe margin on outer element can serve the same goal

apps/store/src/components/Video/Video.css.ts
alebedev: This could a valuable comment in the code

apps/store/src/components/Video/Video.css.ts
```diff
@@ -0,0 +1,138 @@
...
+  alignItems: 'flex-start',
+  cursor: 'pointer',
+})
+
+export const controls = style({
+  display: 'flex',
+  gap: theme.space.xs,
+  width: '100%',
+
+  '@media': {
```
alebedev, line 58: Can we use `hoverStyles` here?

alebedev (approved):
Great work!  Some style / markup comments

apps/store/src/components/Video/Video.tsx
guilhermespopolin: Yeah, that's fine :)

guilhermespopolin (approved):
Nice work with this

apps/store/src/components/Video/Video.tsx
gustaveen: Initially I thought I would put a longer list with conditionals but for shorter cases I agree it's better to inline 👍

apps/store/src/components/Video/Video.tsx
gustaveen: Could be, let's try it! 👍

gustaveen: Hm, for this case it actually felt cleaner to keep it as vars. Otherwise I would create 6 single classes that would all specify the same breakpoints.

apps/store/src/components/Video/Video.css.ts
gustaveen: In this case we're not applying the styles for the actual hover event, but base styles only if the device supports `hover`. So this is probably a rare case, I didn't spot the difference at first as well :)

apps/store/src/components/Video/Video.tsx
gustaveen: We need it to handle the fade in/fade out animation but the positioning can be removed from here 👍

gustaveen:
### Merge activity

* **May 8, 8:28 AM EDT**: [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4238) rebased this pull request as part of a merge.

PR #4246, opened 2024-05-10
Navigation progress indicator for app router

alebedev:
## Describe your changes

Implementation of nav progress indicator for app router

We no longer have router events to rely on.

Search for existing libraries
- Found almost-good-enough solution in https://github.com/tomcru/holy-loader lib, but unfortunately it does not support changing progress bar color after initial load.
- There's also https://github.com/TheSGJ/nextjs-toploader, but is has unsolved issues in tracker that make it probably not very good option

This PR combines navigation detection logic from holy-loader with our original solution for styling nprogress with random color

### DIFF
--- a/apps/store/src/appComponents/RootLayout/NavigationProgressIndicator.tsx
+++ b/apps/store/src/appComponents/RootLayout/NavigationProgressIndicator.tsx
@@ -1,0 +1,146 @@
+'use client'
+
+import { setElementVars } from '@vanilla-extract/dynamic'
+import NProgress from 'nprogress'
+import 'nprogress/nprogress.css'
+import { useEffect } from 'react'
+import { theme } from 'ui'
+import { navigationProgressColor } from '@/appComponents/RootLayout/NavigationProgressIndicator.css'
+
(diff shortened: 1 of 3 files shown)


apps/store/src/appComponents/RootLayout/NavigationProgressIndicator.tsx
```diff
@@ -0,0 +1,146 @@
...
+  useEffect(() => {
+    const startProgress = (): void => {
+      setElementVars(document.body, { [navigationProgressColor]: getRandomColor() })
+      NProgress.start()
+    }
+
+    const stopProgress = (): void => {
+      NProgress.done()
+    }
+
```
alebedev, line 21: startProgress and stopProgress are only modifications to holy-loader effect

gustaveen (approved):
Looks like a good replacement 👍, let's see if we can bring back randomization on every route change

alebedev:
@gustaveen randomization works, it was not working when I tried external libraries instead of copying and modifying their underlying logic

PR #4253, opened 2024-05-13
GEN-2066 - styles(widget-product): redesign

guilhermespopolin:
## Describe your changes

* Redesign Widget product item

https://github.com/HedvigInsurance/racoon/assets/19200662/57800736-fcad-4d21-9095-a71755768807

Part of _Enabling Car for Widget flows_ project


guilhermespopolin:
* **#4253** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4253?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

apps/store/src/features/widget/ProductDetails.css.ts
```diff
@@ -0,0 +1,32 @@
...
+  rowGap: theme.space.md,
+})
+
+export const listItem = style({
+  display: 'flex',
+  justifyContent: 'space-between',
+  alignItems: 'center',
+})
+
+export const documentLink = style({
```
Youakeem, line 16: Can't we use the `Text` component instead?

apps/store/src/features/widget/ProductItem.css.ts
```diff
@@ -2,28 +2,37 @@ import { style, styleVariants } from '@vanilla-extract/css'
...
 const cardBase = style({
   borderRadius: theme.radius.md,
   padding: theme.space.md,
-  paddingBottom: 0,
   backgroundColor: theme.colors.opaque1,
   '@media': {
     '(hover: hover)': {
       selectors: {
-        [`&:has(${hoverable}:hover)`]: {
+        [`&:has(${cardHeader}:hover)`]: {
```
Youakeem, line 29: Out of curiosity, can we achieve the same behaviour by using a css variable for the background color and re-assigning it on hover?

apps/store/src/features/widget/ProductItem.tsx
```diff
@@ -70,63 +69,84 @@ export const ProductItem = (props: Props) => {
     url: item.url,
   }))
 
-  const handleClickHoverable = () => {
+  const toggleExpandCard = () => {
```
Youakeem, line 72: Nice!

apps/store/src/features/widget/ProductItem.tsx
```diff
@@ -70,63 +69,84 @@ export const ProductItem = (props: Props) => {
...
-                  <CrossIconSmall color={theme.colors.textSecondary} />
-                </button>
-              )}
-            </div>
-            <Text as="p" color="textTranslucentSecondary">
-              {props.selectedOffer.exposure.displayNameShort}
+      <div
+        className={cardHeader}
+        style={{ gridTemplateColumns: pillow.src ? 'auto 1fr' : '1fr' }}
+        onClick={toggleExpandCard}
```
Youakeem, line 83: From an a11y point of view this might not be the best behaviour. Having nested interactive elements is tricky.

I found this codepend that does something similar with a button inside of a clickable card, however, they aren't really nested.
https://codepen.io/kkhenriquez/pen/yLeXZXE

The same concept should work with a button as well as far as I know.

apps/store/src/features/widget/ProductItemContainer.tsx
```diff
@@ -8,27 +8,21 @@ type Props = {
   selectedOffer: Offer
 } & Pick<
   ComponentProps<typeof ProductItem>,
-  'shopSessionId' | 'greenVariant' | 'onDelete' | 'defaultExpanded' | 'mode' | 'children'
+  'shopSessionId' | 'greenVariant' | 'onDelete' | 'onEdit' | 'defaultExpanded' | 'mode'
 >
 
-export const ProductItemContainer = (props: Props) => {
+export const ProductItemContainer = ({ offers, selectedOffer, ...delegated }: Props) => {
```
Youakeem, line 14: Do we need this component for server components or does it serve another purpose?

apps/store/src/features/widget/ProductDetails.css.ts
guilhermespopolin: Yes, good suggestion. We still gonna need to tweak the hover and focus-visible styles but the remaining styles can be added via `Text` component.

apps/store/src/features/widget/ProductItem.tsx
guilhermespopolin: Yes, that's a common solution for "link" cards, or card that should be clickable as a whole. Cases like these are trickier because we want only a small portion of the card to be clickable. Maybe I can get away with it by making the div work as it was a button with `role=button` and making it focusable with `tabindex=0`.

apps/store/src/features/widget/ProductItem.tsx
```diff
@@ -70,63 +69,84 @@ export const ProductItem = (props: Props) => {
...
-
-      <Collapsible.Root open={expanded} onOpenChange={setExpanded}>
-        <Collapsible.Trigger asChild={true}>
-          <ProductDetailsHeader
-            className={detailsHeader}
-            price={getOfferPrice(props.selectedOffer.cost)}
-            expanded={expanded}
+      <Space y={1}>
+        {mode === 'edit' && (
+          <EditUI
```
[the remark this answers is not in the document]
guilhermespopolin: Kind of. For starter I think it's a good idea add those changes for widget's ProductItem only because for widget we have a simpler flow. Adding those functionality for hedvig-com could increase the scope of the project as we have some different use cases for it on regular flows like the possibility to remove every cart entry, which is not possible for the widget.

apps/store/src/features/widget/ProductItemContainer.tsx
guilhermespopolin: Yeah, that's a bummer one. In summary I'm only using it for getting `tiers` and `deductibles` and them passing them to `ProductItem`. The issue is that, on [SignPage](https://github.com/HedvigInsurance/racoon/blob/main/apps/store/src/features/widget/SignPage.tsx#L153) we only render the ProductItemContainer if we found the "main offer", meaning and offer for the product the user has selected during the first step of that flow. While it shouldn't be case, that mainOffer can be null so that's why the check for it's presence before rendering ProductItemContainer. We could move useTiersAndDeductibles call from ProductItemContainer to SignPage and them remove the container, the issue with it is that useTiersAndDeductibles expects a selected offer (it can't be null) and since mainOffer (which is nullable) is the selectedOffer we have type issues. One solution would be to throw an error if mainOffer is not found just don't know if that would be ok.

apps/store/src/features/widget/ProductDetails.tsx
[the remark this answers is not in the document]
guilhermespopolin: Yeah, so they're slightly different:

1. Regular components/ProductItem
ProductDetails corresponds to the marked area. Observe "details" is the trigger of the collapsable component and therefore doesn't belong to ProductDetails
Screenshot 2024-05-14 at 10 02 30

2. Widget ProductItem
Screenshot 2024-05-14 at 10 03 09
"Details" is not a heading of ProductDetails as "Documents". The trigger for the Collapsible component is the Show/Hide details button.

That's another case were I wanted to make them work the same way for both cases but the design for hedvig-com ProductItem is a "blocker" due the controls it offers (Edit and Remove buttons)

apps/store/src/features/widget/ProductItem.tsx
Youakeem: Why not have the card header it's own "card"?

```
<Card>
  // CardHeader is clickable
  <CardHeader>
    ...
    <DeleteButton/>
  </CardHeader>

 ...
</Card>

```

apps/store/src/features/widget/ProductItemContainer.tsx
Youakeem: Can we not use `useTiersAndDeductibles` in `ProductItem` and then conditionally render `ProductItem` in `SignPage`?

apps/store/src/features/widget/ProductDetails.css.ts
guilhermespopolin: Turned out to be more difficult to understand and making hover/focus styles to work is harder with this. Gonna move away with the current solution.

apps/store/src/features/widget/ProductItemContainer.tsx
guilhermespopolin: Yes, we can but then `ProductItem` would need to take all the offers as prop which would be counter intuitive.

Youakeem: Aha, Ok, I see now! I personally don't mind passing `offers` to `ProductItem` since that's what is needed to get `tiers` and `deductibles` but I see your point.

apps/store/src/features/widget/ProductDetails.css.ts
Youakeem: Makes sense, I think it would be easier if we could do `<Text as="a" ...>`

apps/store/src/features/widget/ProductItem.css.ts
guilhermespopolin: It doesn't work 😔

Youakeem: I see, thanks for looking into it, man! 🙏

apps/store/src/features/widget/ProductItem.tsx
guilhermespopolin: Could implement it with regular button and aria attributes.

https://github.com/HedvigInsurance/racoon/assets/19200662/d44bcdb1-3541-4983-b925-95c7a7089924

Youakeem: How does it work with the delete button showing? If I understand correctly that's the only nested button in the card header, no?

apps/store/src/features/widget/ProductDetails.tsx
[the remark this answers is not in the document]
gustaveen: Got it, thanks for explaining!

apps/store/src/features/widget/ProductItem.tsx
guilhermespopolin: Yes. It works fine, pretty much as it's working so far. We prevent bubbling inside the delete handler so the card doesn't get expanded when deleting an item.

gustaveen (approved):
Good job with this! 🙌

guilhermespopolin:
### Merge activity

* **May 14, 8:33 AM EDT**: @guilhermespopolin started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4253).

apps/store/src/features/widget/ProductItem.tsx
Youakeem: I see! I think it's not valid HTML to nest button tho!

PR #4295, opened 2024-05-20
UX clarifications for bundle discounts

alebedev:
## Describe your changes

More case-specific texts for bundle discounts
- Only show total saved amount if all insurances start at the same day
- Add earliest start date to discount summary if cart entries have different start dates
- Make cart toast subtitle for bundle product links into separate key so that it can be made shorter than similar subtitle on cart page where we need to mention Accident not being counted for the bundle limit


alebedev:
* **#4295** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4295?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

apps/store/src/components/CartNotification/CartToast.tsx
```diff
@@ -68,7 +68,19 @@ export const CartNotificationContent = ({ onClose, ...productItemProps }: Props)
         })}
       </BundleDiscountSummary>
     )
-  } // Use default header from the cart page otherwise
+  } else {
+    // This is awkward, but we're duplicating default header just to have different subtitle
+    bundleDiscountHeader = (
+      <Text>
```
Youakeem, line 74: Do we need to wrap it in a `Text` instead of a `div` maybe?

apps/store/src/features/bundleDiscount/BundleDiscountCartSummary.tsx
```diff
@@ -1,27 +1,45 @@
...
+    const totalSaved = formatter.money({
+      amount: cost.discount.amount * redeemedCampaign.discount.months,
+      currencyCode: cost.net.currencyCode,
+    })
+    content = t('BUNDLE_DISCOUNT_SUMMARY', {
+      percentage,
+      totalSaved,
+    })
+  } else {
+    const earliestStartDate = [...startDates.values()].toSorted()[0]
```
Youakeem, line 37: This assumes dates are in ISO format, right? It might not behave the same if the date format change for some reason.

alebedev: We use ISO 8601 strings in our GraphAPI, which by definition are sorted correctly with string sort

alebedev:
@gustaveen accident does not count towards bundle size, but if you get bundle otherwise, accident will get same discount.  Home + Accident - no bundle, Home + Car + Accident - 20% off everything

PR #4297, opened 2024-05-21
Use App router to render most CMS pages

alebedev:


alebedev:
* **#4297** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4297?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

apps/store/src/app/[locale]/test/ClientComponent.tsx
```diff
@@ -18,13 +21,23 @@ export const ClientComponent = () => {
   // Read jotai value from provider
   const productMetadata = useAtomValue(productsMetadataAtom(locale))
   console.log('productMetadata@ClientComponent, items:', productMetadata?.length)
+  const formatter = useFormatter()
+  const priceReduction = formatter.monthlyPrice({ amount: 20.0, currencyCode: CurrencyCode.Sek })
```
Comment (Youakeem) [inline apps/store/src/app/[locale]/test/ClientComponent.tsx:25]: Should we hardcode the amount?

Comment (alebedev) [reply apps/store/src/app/[locale]/test/ClientComponent.tsx:25]: It's a test page similar to Storybook story.  I'm planning to remove it after merging this PR

apps/store/src/features/blog/BlogArticleListBlock/BlogArticleListBlock.tsx
```diff
@@ -36,12 +34,8 @@ export const BlogArticleListBlock = (props: Props) => {
   }, [teaserList, props.blok.categories])
 
   const pathname = usePathname()
-  const searchParams = useSearchParams()
-
   const initiallyVisibleCount = props.blok.initiallyVisibleCount ?? 12
-  const showAll =
-    filteredTeaserList.length <= initiallyVisibleCount ||
-    searchParams?.has(SHOW_ALL_QUERY_PARAM, '1')
+  const [showAll, setShowAll] = useState(filteredTeaserList.length <= initiallyVisibleCount)
```
Youakeem, line 38: It seems that `ShowAllQueryTrigger` checks the query param and then set the state to true. If so, why not just read the query param directly?

```
const showAll = searchParams?.has(SHOW_ALL_QUERY_PARAM, '1')
```

apps/store/src/features/blog/BlogArticleListBlock/BlogArticleListBlock.tsx
```diff
@@ -84,7 +78,23 @@ export const BlogArticleListBlock = (props: Props) => {
           )}
         </Space>
       </GridLayout.Content>
+      <Suspense>
```
Youakeem, line 81: I am not sure why `Suspense` is needed here since `ShowAllQueryTrigger` doesn't seem to suspend. Am I missing something?

apps/store/src/services/storyblok/storyblok.rsc.ts
```diff
@@ -20,7 +20,7 @@ import type { RoutingLocale } from '@/utils/l10n/types'
 // Default cache policy assumes that anything we fetch with `cv` is immutable
 // We're passing it explicitly to avoid being affected by changing defaults in Next 15
 const STORYBLOK_CACHE_SETTINGS = {
-  cache: 'force-cache',
+  // Sets cache: 'force-cache' automatically
```
Youakeem, line 23: This comment and the one above are contradicting. Do we want to set it explicitly or is it set automatically?

apps/store/src/features/blog/BlogArticleListBlock/BlogArticleListBlock.tsx
alebedev: It's impossible to access searchParams from render function unless you're doing client-side rendering AND opt out from SSR entirely.  Hence trick with moving `ShowAllQueryTrigger` inside `Suspense`

alebedev: Alternative way is to have client-side effect that reads window.location, but it won't be reactive. Which is fine if you never change the param while on the page, but not here

apps/store/src/features/blog/BlogArticleListBlock/BlogArticleListBlock.tsx
alebedev: Answered above, this is to delay rendering of component that has `useSearchParams()` hook

apps/store/src/services/storyblok/storyblok.rsc.ts
alebedev: If there's both `cache` and `next.revalidate`, NextJs gives a warning. We should probably get back here when Next 15 upgrade arrives

Youakeem: I see, let's remove this bit then `We're passing it explicitly to avoid being affected by changing defaults in Next 15` since we aren't passing it explicitly yet.

apps/store/src/features/blog/BlogArticleListBlock/BlogArticleListBlock.tsx
alebedev: Makes sense, I'll comment it

Youakeem (approved):

PR #4303, opened 2024-05-24
fix/load GTM at initial page load to avoid dropped affiliate tracking and fix bug where consent isn't updated correctly after consent is given

gustaveen:
## Describe your changes
We have a bug with the consent/tracking feature released last week.

- For some reason consent is not update correctly in GTM when we defer loading the script. Might be a timing issue, we try to send update before script is fully loaded
- Affiliate tracking is dropped when consent aren't updated properly

Immediate fix is to load GTM on page load as we did before. Let's see if we can find a way around it. But if it's too flaky I see the risk of loosing tracking and affiliate tracking is more significant than the savings in first page load. 

We will still omit loading OneTrust script other than strictly necessary which is the biggest win 

We have a bug with the consent/tracking feature released last week.

### DIFF
--- a/apps/store/src/components/CookieConsent/CookieConsent.tsx
+++ b/apps/store/src/components/CookieConsent/CookieConsent.tsx
@@ -39,10 +39,11 @@ export function CookieConsent() {
   const { t } = useTranslation(['cookieConsent', 'common'])
   const locale = useRoutingLocale()
 
-  // Make sure we don't show the cookie banner or load GTM before we have read the OneTrust cookie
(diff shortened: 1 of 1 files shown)


gustaveen:
* **#4303** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4303?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

alebedev (approved):
OK as a temporary solution. We 100% should follow up and find a way to handle delayed GTM init correctly

PR #4320, opened 2024-05-30
tech: Create Alert component

Youakeem:
## Describe your changes
Create a composable Alert component which supports different variants and icons.

To be used to create the InfoCard component with more flexibility with composition.

### DIFF
--- a/packages/ui/src/components/Alert/Alert.css.ts
+++ b/packages/ui/src/components/Alert/Alert.css.ts
@@ -1,4 +1,4 @@
-import { createVar, globalStyle, style } from '@vanilla-extract/css'
+import { createVar, style } from '@vanilla-extract/css'
 import { recipe } from '@vanilla-extract/recipes'
 import { tokens } from '../../theme'
 
@@ -54,10 +54,7 @@ export const rootStyles = recipe({
 
 export const iconStyles = style({
   marginTop: tokens.space.xxs,
-})
-
-globalStyle(`${iconStyles} path`, {
-  fill: AlertElement,
+  color: AlertElement,
 })
 
 export const bodyStyles = style({
--- a/packages/ui/src/components/Alert/Alert.css.ts
+++ b/packages/ui/src/components/Alert/Alert.css.ts
@@ -1,0 +1,79 @@
+import { createVar, globalStyle, style } from '@vanilla-extract/css'
+import { recipe } from '@vanilla-extract/recipes'
+import { tokens } from '../../theme'
+
+export const AlertBackground = createVar()
+export const AlertForeground = createVar()
+export const AlertElement = createVar()
+
+export const rootStyles = recipe({
+  base: {
+    paddingBlock: tokens.space.sm,
+    paddingInline: tokens.space.md,
+    borderRadius: tokens.radius.md,
+    display: 'flex',
+    gap: tokens.space.sm,
+    backgroundColor: AlertBackground,
+  },
+  variants: {
+    variant: {
+      info: {
+        vars: {
+          [AlertBackground]: tokens.colors.signalBlueFill,
+          [AlertForeground]: tokens.colors.signalBlueText,
+          [AlertElement]: tokens.colors.signalBlueElement,
+        },
+      },
+      warning: {
+        vars: {
+          [AlertBackground]: tokens.colors.signalAmberFill,
+          [AlertForeground]: tokens.colors.signalAmberText,
+          [AlertElement]: tokens.colors.signalAmberElement,
+        },
+      },
+      error: {
+        vars: {
+          [AlertBackground]: tokens.colors.signalRedFill,
+          [AlertForeground]: tokens.colors.signalRedText,
+          [AlertElement]: tokens.colors.signalRedElement,
+        },
+      },
+      success: {
+        vars: {
+          [AlertBackground]: tokens.colors.signalGreenFill,
+          [AlertForeground]: tokens.colors.signalGreenText,
+          [AlertElement]: tokens.colors.signalGreenElement,
+        },
+      },
+    },
+  },
+  defaultVariants: {
+    variant: 'info',
+  },
+})
+
+export const iconStyles = style({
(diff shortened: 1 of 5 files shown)


Youakeem:
* **#4320** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4320?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

packages/ui/src/components/Alert/Alert.css.ts
```diff
@@ -0,0 +1,79 @@
...
+  defaultVariants: {
+    variant: 'info',
+  },
+})
+
+export const iconStyles = style({
+  marginTop: tokens.space.xxs,
+})
+
+globalStyle(`${iconStyles} path`, {
```
gustaveen, line 59: Does this style need to be set globally?

gustaveen (commented):
Looks great! Just wondering about if we need to abstract it to a new Alert component? Will we ever use it other than in the InfoCard? Then we could probably just move this into the InfoCard component directly, it's nice if it's a 1 to 1 mapping with the UI kit in Figma IMO

packages/ui/src/components/Alert/Alert.css.ts
Youakeem: Yup, not possible to select children unfortunately. I tried using `selectors` but it broke and the terminal said this is how it should be done instead.

Youakeem:
> Looks great! Just wondering about if we need to abstract it to a new Alert component? Will we ever use it other than in the InfoCard? Then we could probably just move this into the InfoCard component directly, it's nice if it's a 1 to 1 mapping with the UI kit in Figma IMO
> 
> Does this style need to be set globally?

Either would work with me. I think of `Alert` as the primitive that supports different compositions (Icon, text, buttons) and `InfoCard`, `AttentionCard` ..etc as abstractions for common use cases (Specific icon, no buttons).

packages/ui/src/components/Alert/Alert.css.ts
gustaveen: Ah right, it's a bit of a mind shift to not target childrens :)

alebedev: Can we use color=currentColor in SVG to inherit color from DOM parent instead?

packages/ui/src/components/Alert/Alert.css.ts
```diff
@@ -0,0 +1,79 @@
...
+  fill: AlertElement,
+})
+
+export const bodyStyles = style({
```
alebedev, line 63: It is time to app sprinkle for `flexGrow: 1`?

packages/ui/src/components/Alert/Alert.stories.tsx
```diff
@@ -0,0 +1,123 @@
...
+    variant: {
+      options: ['info', 'warning', 'error', 'success'],
+      control: { type: 'select' },
+    },
+  },
+  args: {
+    variant: 'info',
+  },
+  parameters: {
+    design: {
```
alebedev, line 37: Nice!

packages/ui/src/components/Alert/Alert.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+function Actions({ children, className, ...props }: ActionsProps) {
+  return (
+    <div className={clsx(actionsContainerStyles, className)} {...props}>
+      {children}
+    </div>
+  )
+}
+
+type ActionProps = ButtonProps<'button'>
```
alebedev, line 66: `ComponentProps` above is a nice touch.  Can you use something like `ComponentProps<typeof Button<'button'>>` and does it look better or worse than just `ButtonProps`?

gustaveen:
> > Looks great! Just wondering about if we need to abstract it to a new Alert component? Will we ever use it other than in the InfoCard? Then we could probably just move this into the InfoCard component directly, it's nice if it's a 1 to 1 mapping with the UI kit in Figma IMO
> > Does this style need to be set globally?
> 
> Either would be fine by me. However, I think of `Alert` as the primitive that supports different compositions (Icon, text, buttons) and `InfoCard`, `AttentionCard` ..etc as abstractions for common use cases (Specific icon, no buttons).

I like that it's composable! 👌 Let's put it in InfoCard then perhaps and we could move InfoCard to ui kit

packages/ui/src/components/Alert/Alert.css.ts
Youakeem: Good point, the icon components take a `color` prop which sets `fill` on the `path` and it defaults to `currentColor`. I'll try setting color on the SVG itself.

alebedev (approved):

packages/ui/src/components/Alert/Alert.css.ts
Youakeem: Nice! works beautifully!

packages/ui/src/components/Alert/Alert.css.ts
Youakeem: I am thinking about it! I think we should have base sprinkles which is what we have now, and then extend those to create `patterns` like `flexSprinkles` for example.

I'll play with it a little and open a PR.

packages/ui/src/components/Alert/Alert.stories.tsx
Youakeem: Yes, I think it's a pretty nice addition as well! I got it from another story tho!

packages/ui/src/components/Alert/Alert.tsx
Youakeem: They are the same. I've updated to use `ComponentProps`.

Youakeem:
### Merge activity

* **Jun 3, 2:54 AM EDT**: @Youakeem started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4320).

PR #4332, opened 2024-06-03
Optimize ProductPageBlock rendering by making lower content area render lazily

alebedev:
## Describe your changes

Improve rendering speed of product pages by rendering general content (below "Coverage") lazily when it comes into view

Before

After

"See your price" is one of known slow elements and this optimization makes it faster thanks to rendering much smaller tree when opening just top part of product page

### DIFF
--- a/apps/store/src/blocks/ProductPageBlock.tsx
+++ b/apps/store/src/blocks/ProductPageBlock.tsx
@@ -2,9 +2,10 @@
 import styled from '@emotion/styled'
 import type { SbBlokData } from '@storyblok/react'
 import { storyblokEditable, StoryblokComponent } from '@storyblok/react'
-import { motion, useInView, useScroll } from 'framer-motion'
+import { motion, useScroll } from 'framer-motion'
 import type { ReactNode } from 'react'
-import { useState, useEffect, useRef } from 'react'
+import { startTransition } from 'react'
+import { useCallback, useState, useEffect, useRef } from 'react'
 import { theme, mq } from 'ui'
 import * as GridLayout from '@/components/GridLayout/GridLayout'
 import { MAX_WIDTH } from '@/components/GridLayout/GridLayout.constants'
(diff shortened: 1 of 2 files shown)


guilhermespopolin (approved):
Nice improvement 🚀

apps/store/src/blocks/ProductPageBlock.tsx
```diff
@@ -261,14 +261,35 @@ const useActiveSectionChangeListener = (
...
+  )
+}
+
+const useRenderLazily = () => {
+  const [shouldRender, setShouldRender] = useState(false)
+  // Callback need to be stable to avoid rerender loop
+  const makeVisible = useCallback(() => startTransition(() => setShouldRender(true)), [])
+  useIdleCallback(makeVisible)
+  return shouldRender
+}
```
gustaveen, line 295: Could you explain how this works, what triggers it?

apps/store/src/blocks/ProductPageBlock.tsx
```diff
@@ -261,14 +261,35 @@ const useActiveSectionChangeListener = (
 // Optimization:
 // Perils list could be hundreds of DOM elements (400+ for Apartment rent) and if we lazy-render it, page load time of product pages improves
 const CoverageSection = (props: { blocks: Array<SbBlokData> }) => {
-  const ref = useRef(null)
-  const cameIntoView = useInView(ref, { once: true })
+  const shouldRenderContent = useRenderLazily()
```
gustaveen, line 264: Couldn't we just use the `useInView` hook from framer? I've used it on a few components and it's very straight forward :) 
https://www.framer.com/motion/use-in-view/

gustaveen: Ah sorry, I thought we wanted to load it when inView but now I see we just want it to load lazily. Nice!

alebedev: I changed it after discussion with Peter

gustaveen (approved):

apps/store/src/blocks/ProductPageBlock.tsx
alebedev: At the first render of using component, it schedules low-priority update that sets `shouldRender` to true.  So, short version will render during SSR and first client pass, then long version will be rendered after as low-piority task

PR #4333, opened 2024-06-03
feature: [GEN-2105] Show thank you message for handled cases

Youakeem:
## Describe your changes
1. Query for case status on page load
2. Show thank you message for already handled cases
3. Create common HandledMessage component

1. Improve UX


alebedev (approved):
Looks good!

apps/store/src/app/switcher/[id]/InsuranceSwitchExpirationDateForm.tsx
```diff
@@ -24,7 +24,7 @@ export default function InsuranceSwitchExpirationDateForm({ id }: Props) {
   const { t } = useTranslation('contractSwitchConfirmationForm')
   const [state, formAction] = useFormState(submitSwitchConfirmation, null)
 
-  console.log({ id })
+  const isSuccessful = state?.messages?.find((message) => message.type === 'success')
```
Comment (guilhermespopolin) [inline apps/store/src/app/switcher/[id]/InsuranceSwitchExpirationDateForm.tsx:27]: nit:
Since we don't actually need to hold the successful message but only a boolean representing if there's a successful message or not, `some` could be a better fit:

```suggestion
  const isSuccessful = state?.messages?.some((message) => message.type === 'success')
```

apps/store/src/app/switcher/[id]/SwitchConfirmation.graphql
```diff
@@ -0,0 +1,16 @@
...
+      message
+    }
+  }
+}
+
+query SwitcherCase($id: UUID!) {
+  switcherCase(id: $id) {
+    isCompleted
+  }
+}
```
Comment (guilhermespopolin) [inline apps/store/src/app/switcher/[id]/SwitchConfirmation.graphql:16]: Should we request `id` as well? I mean, for apollo caching purposes.

```suggestion
mutation SwitcherCaseComplete($id: UUID!, $currentExpiryDate: Date!) {
  switcherCaseComplete(id: $id, currentExpiryDate: $currentExpiryDate) {
    switcherCase {
      id
      isCompleted
    }
    userError {
      message
    }
  }
}

query SwitcherCase($id: UUID!) {
  switcherCase(id: $id) {
    id
    isCompleted
  }
}
```

guilhermespopolin (commented):
Nice one. Just left some small comments :)

apps/store/src/app/switcher/[id]/InsuranceSwitchExpirationDateForm.tsx
Comment (Youakeem) [reply apps/store/src/app/switcher/[id]/InsuranceSwitchExpirationDateForm.tsx:27]: Great point!

apps/store/src/app/switcher/[id]/SwitchConfirmation.graphql
Comment (Youakeem) [reply apps/store/src/app/switcher/[id]/SwitchConfirmation.graphql:16]: Good point! These are queried server side tho so there is no cache but it doesn't hurt to include `id` 👍

guilhermespopolin (approved):

PR #4336, opened 2024-06-04
Hide quick add section header when section is hidden

alebedev:
## Describe your changes

Fixes a bug when hiding cross-sale offer did not hide a heading "Popular add-on" that was related to it

Before fix:

After fix: no extract header

### DIFF
--- a/apps/store/src/components/QuickAdd/QuickAddOfferContainer.css.ts
+++ b/apps/store/src/components/QuickAdd/QuickAddOfferContainer.css.ts
@@ -1,0 +1,10 @@
+import { style } from '@vanilla-extract/css'
+import { theme } from 'ui/src/theme'
+import { sprinkles } from 'ui/src/theme/sprinkles.css'
+
+export const quickAddSection = style([
+  sprinkles({ display: 'flex', flexDirection: 'column' }),
+  {
+    rowGap: theme.space.md,
+  },
+])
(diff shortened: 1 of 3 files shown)


alebedev:
* **#4336** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4336?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

gustaveen (approved):
👍

apps/store/src/components/QuickAdd/QuickAddOfferContainer.css.ts
```diff
@@ -0,0 +1,10 @@
+import { style } from '@vanilla-extract/css'
+import { theme } from 'ui/src/theme'
+import { sprinkles } from 'ui/src/theme/sprinkles.css'
+
+export const quickAddSection = style([
+  sprinkles({ display: 'flex', flexDirection: 'column' }),
```
guilhermespopolin, line 6: Dunno if this were already discussed but wouldn't this just add more complexity on this case when compared with just adding the declarations without `sprinkles`? 🤔

guilhermespopolin (approved):
Nice catch

apps/store/src/components/QuickAdd/QuickAddOfferContainer.css.ts
alebedev: I was assuming it'll generate less CSS as a result, but we should probably test that

PR #4351, opened 2024-06-10
chore(Heading): migrate to vanilla extract

guilhermespopolin:
## Describe your changes

Migration of `Heading` into `vanilla-extract-css` consists on 3 steps:

1. Have simplified `Heading` API by removing support for adding margin via prop and adding it as an inline `style`. We had 4 places were this was being used. Personally I think things like that is simpler to implement as making the consumer of `Heading` to define CSS to add it instead of enable the change via props, which makes the `Heading` API too big (in terms of number of props one could provide). With that said I'm ok getting it back, so let me know what you think.
2. Support changing `Heading`'s color and alignment via props by using `sprinkles`. A side effect of this is making sure we always provide `color` and `textAlign` to `sprinkles` by either getting them from the consumer or, when none is provided, get a default value. There's a gotcha when using it with _Storyblok_ via `HeadingBlock` where, even though we define a default value for those in Storyblok, it still send '' (Empty string) to `Heading` component, which makes sprinkles to fail. That's why there are some code in `HeadingBlock` preventing that
3. Add a `responsiveVariants` abstraction on top of `vanilla-extract-css` in order to implement `Heading` responsive variant styles functionality.

Item 3 is the not straight forward change so more on it below:

* We have several places where that functionality is being used by consumers of `Heading`; directly in code and indirectly via `HeadingBlock`. So just simplifying the API was not a viable option.
* Yes, it's complex but at least the complexity is contained in a single module and I believe a form of implementing the same with `vanilla-extract-css` api/packages will be added in the future. When that happens we can stripe out the custom implementation. Optional reading on how it works can be found below.

**Optional: how 'responsiveVariants' abstraction works**

It's implementation is based on `vanilla-extract`'s [`addFunctionSerializer`](https://vanilla-extract.style/documentation/api/add-function-serializer/#addfunctionserializer) API more details can be found on their docs but basically it goes as follows:

We need to create something that get's executed during build time and during runtime:

* During build time we know all the possible variants and the styles associated with them as that is defined inside .css.ts files. Then by using `vanilla-extract` `style` function we create classes that applies each of variant styles for each breakpoint level. The name of those classes are hold in a map object so we can determine each class should be applied during runtime.
* By using `addFunctionSerializer`, a runtime function that has access to that map object (called config in the code) is bundled into client code, so whenever called during a component render process we can have access responsive styles configuration one has provided to `Heading` component - E.g: `variant={{ _: 'sm', md: 'lg' }}` - and with that information we can define the classes that should be applied for that component.  The runtime part is basically the same thing we had before with `Heading.helpers.ts` file.


guilhermespopolin:
* **#4351** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4351?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

packages/ui/src/components/Heading/Heading.tsx
```diff
@@ -21,50 +21,38 @@ type HeadingColors = Pick<
...
+  return (
+    <Component
+      className={clsx(
+        responsiveVariantStyles(variant),
+        sprinkles({ color, textAlign: align }),
+        className,
+      )}
+      {...rest}
+    >
+      {!balance ? children : <Balancer>{children}</Balancer>}
```
alebedev, line 55: What do you think about variant without negation? `balance ? <Balancer>{children}</Balancer> : children`

packages/ui/src/utils/responsiveVariants/responsiveVariants.ts
```diff
@@ -0,0 +1,74 @@
+import { style } from '@vanilla-extract/css'
+import { addFunctionSerializer } from '@vanilla-extract/css/functionSerializer'
+import { minWidth, type Level } from 'ui/src/theme'
+import { breakpoints } from 'ui/src/theme/media-query'
+import type { ResponsiveVariantsConfig, VariantStyles, Styles } from './repsonsiveVariants.types'
+import { createResponsiveVariantsRuntimeFn } from './responsiveVariantsRuntime'
+
+export function responsiveVariants<PossibleVariants extends string>(
```
alebedev, line 8: Could you add some overview comments describing how it works?  Link to vanilla extract docs for custom serializer would be nice

alebedev (approved):
Great work!  I especially like simplifying the problem by changing some use cases to inline styles

apps/store/src/app/debugger/car-trial/page.tsx
```diff
@@ -13,7 +13,7 @@ function CarTrialDebuggerPage() {
               align="center"
               balance={true}
               variant={{ _: 'serif.24', lg: 'serif.32' }}
-              mt={theme.space.xl}
+              style={{ marginTop: theme.space.xl }}
```
gustaveen, line 16: Personally, I like the margin as a utility prop for Heading and Text component but let's omit it to start with and we can see how we expose the sprinkles in a reusable way

apps/store/src/blocks/HeadingBlock.tsx
```diff
@@ -21,13 +21,19 @@ export type HeadingBlockProps = SbBaseBlockProps<{
 }>
 
 export const HeadingBlock = ({ blok, nested }: HeadingBlockProps) => {
+  // Gotcha: we may get emtpy string from Storyblok for 'color' and 'textAlignment', this should handled safely
+  // eslint-disable-next-line @typescript-eslint/prefer-nullish-coalescing
+  const color = blok.color || 'textPrimary'
+  // eslint-disable-next-line @typescript-eslint/prefer-nullish-coalescing
+  const align = blok.textAlignment || 'center'
```
gustaveen, line 28: 👍

packages/ui/src/utils/responsiveVariants/responsiveVariants.ts
gustaveen: Just curious, did you find any good example of someone who had implemented something similar before?

gustaveen (approved):
Nice work with this @guilhermespopolin 🙌

Very happy we can keep the component API! 🥳

apps/store/src/blocks/HeadingBlock.tsx
```diff
@@ -21,13 +21,19 @@ export type HeadingBlockProps = SbBaseBlockProps<{
 }>
 
 export const HeadingBlock = ({ blok, nested }: HeadingBlockProps) => {
+  // Gotcha: we may get emtpy string from Storyblok for 'color' and 'textAlignment', this should handled safely
```
Youakeem, line 24: What about defining default values for the `color` and `align` properties in `Heading`?

apps/store/src/components/QuickAdd/QuickAddBundleView.stories.tsx
```diff
@@ -141,7 +141,7 @@ export const Default: Story = {
                   size="xlarge"
                   src="https://assets.hedvig.com/f/165473/832x832/1bb4813dd1/hedvig-pillows-accident.png"
                 />
-                <Heading as="h1" variant="standard.18" mt={theme.space.md}>
+                <Heading as="h1" variant="standard.18" style={{ marginTop: theme.space.md }}>
```
Youakeem, line 144: Any reason not to use `sprinkles` instead of inline styles?

packages/ui/src/components/Heading/Heading.css.ts
```diff
@@ -0,0 +1,97 @@
+import { theme } from '../../theme'
+import { responsiveVariants } from '../../utils/responsiveVariants/responsiveVariants'
+
+export const responsiveVariantStyles = responsiveVariants({
```
Youakeem, line 4: Brilliant! 🤩

packages/ui/src/components/Heading/Heading.tsx
```diff
@@ -21,50 +21,38 @@ type HeadingColors = Pick<
...
-  return {
-    color: color ? getColor(color) : 'currentColor',
-    fontFamily: theme.fonts.heading,
-    fontWeight: 400,
-    lineHeight: 1.2,
-    textAlign: align ?? 'left',
-    whiteSpace: 'pre-wrap',
-    ...getMargins(props),
-    ...getHeadingVariantStyles(variant),
+export type HeadingProps = PolymorphicComponentsProps<
```
Youakeem, line 24: Nice!

packages/ui/src/components/Heading/Heading.tsx
```diff
@@ -21,50 +21,38 @@ type HeadingColors = Pick<
...
+  'h1' | 'h2' | 'h3' | 'h4' | 'p',
+  {
+    color?: keyof HeadingColors
+    variant?: PossibleHeadingVariant | Partial<Record<Level | '_', PossibleHeadingVariant>>
```
Youakeem, line 28: Note for the future, would be great to have a type utility that takes the typeof responsiveVariantStyles and outputs this.

Youakeem (approved):
Great work man! 🔥

packages/ui/src/utils/responsiveVariants/responsiveVariants.ts
guilhermespopolin: No, I actually just followed their docs for `addFunctionSerializer` function.

apps/store/src/components/QuickAdd/QuickAddBundleView.stories.tsx
guilhermespopolin: No, I think it was just muscle memory. I think sprinkles here would be quite nice. Thanks for the recommendation.

packages/ui/src/components/Heading/Heading.tsx
guilhermespopolin: Yes. As soon as I finished this and was testing with `store` I've realized that as well. Gonna give it a look.

guilhermespopolin:
### Merge activity

* **Jun 11, 5:21 AM EDT**: @guilhermespopolin started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4351).

PR #4353, opened 2024-06-10
fix(blog): make storyblok preview work again

guilhermespopolin:
## Describe your changes

- Make sure we load blog blocks for Blog Article and Blog Category content types, fixing previewing then on Storyblok

Issue caught by Peter and mentioned [here](https://hedviginsurance.slack.com/archives/C02JPS466NS/p1718020297897609)

### DIFF
--- a/apps/store/src/app/[locale]/[[...slug]]/BlogStoryContainer.tsx
+++ b/apps/store/src/app/[locale]/[[...slug]]/BlogStoryContainer.tsx
@@ -4,7 +4,9 @@ import type { SbBlokData } from '@storyblok/react'
 import type { ReactNode } from 'react'
 import {
   BLOG_ARTICLE_CATEGORIES_PAGE_PROP,
+  BLOG_ARTICLE_CATEGORY_CONTENT_TYPE,
   BLOG_ARTICLE_CATEGORY_LIST_BLOCK,
+  BLOG_ARTICLE_CONTENT_TYPE,
   BLOG_ARTICLE_LIST_BLOCK,
   BLOG_ARTICLE_TEASERS_PAGE_PROP,
 } from '@/features/blog/blog.constants'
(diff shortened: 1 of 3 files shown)


apps/store/src/app/[locale]/[[...slug]]/BlogStoryContainer.tsx
```diff
@@ -40,7 +43,17 @@ export async function BlogStoryContainer({ locale, story, children }: Props) {
     )
   }
   await Promise.all(requests)
-  return <BlogContextProvider blogPageProps={blogPageProps}>{children}</BlogContextProvider>
+
+  const isBlogArticlePage = story.content.component === BLOG_ARTICLE_CONTENT_TYPE
+  const isBlogCategoryPage = story.content.component === BLOG_ARTICLE_CATEGORY_CONTENT_TYPE
+  const rendersBlogBlocks = Object.values(blogPageProps).some(Boolean)
```
Comment (alebedev) [inline apps/store/src/app/[locale]/[[...slug]]/BlogStoryContainer.tsx:49]: that was some debug code committed by mistake. We used to check for blog-related blocks before, do you want to restore it or just rely on story type?

alebedev (approved):
One question inside, looks good otherwise

apps/store/src/app/[locale]/[[...slug]]/BlogStoryContainer.tsx
Comment (guilhermespopolin) [reply apps/store/src/app/[locale]/[[...slug]]/BlogStoryContainer.tsx:49]: 🤔 couldn't find check for blog-related blocks, at least on this file. I think it will be fine if we rely on story type here. I can remove the `renderBlogBlocks` thing.

guilhermespopolin:
### Merge activity

* **Jun 10, 10:17 AM EDT**: @guilhermespopolin started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4353).

PR #4359, opened 2024-06-11
refact(navigation): TopMenuDesktop --> TopMenu renaming

guilhermespopolin:


apps/store/src/components/Header/TopMenu/TopMenu.tsx
```diff
@@ -10,7 +10,7 @@ export type TopMenuDesktopProps = {
   defaultValue?: string
 }
 
-export const TopMenuDesktop = ({ children, defaultValue }: TopMenuDesktopProps) => {
+export const TopMenu = ({ children, defaultValue }: TopMenuDesktopProps) => {
```
alebedev, line 13: Could you document SEO reasons for doing it this way in the code? If we had it there, maybe I wouldn't optimize it away last time :)

guilhermespopolin: Yes, ofc 👍🏻

alebedev (approved):

PR #4364, opened 2024-06-12
pef(sprinkles): reduce the number of sprinkles generated classes by restricting the number of supported space tokens

guilhermespopolin:
# Describe your changes

See title 👆

Discussed [here](https://hedviginsurance.slack.com/archives/C041SD67G82/p1718184721148229)

Reduce amount of CSS sent to the client.


guilhermespopolin:
* **#4364** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4364?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

packages/ui/src/theme/sprinkles.css.ts
```diff
@@ -42,17 +47,17 @@ const responsiveProperties = defineProperties({
...
-    marginBottom: tokens.space,
-    marginLeft: tokens.space,
-    marginRight: tokens.space,
-    padding: tokens.space,
-    paddingTop: tokens.space,
-    paddingBottom: tokens.space,
-    paddingLeft: tokens.space,
-    paddingRight: tokens.space,
-    gap: tokens.space,
+    margin: spaceTokensSubset,
```
gustaveen, line 50: I'm thinking we could remove `margin` and `padding` to save some more. It's very rarely you set the same margin or padding on all sides :)

gustaveen (approved):
Great!

packages/ui/src/theme/sprinkles.css.ts
guilhermespopolin: Good idea

packages/ui/src/theme/sprinkles.css.ts
```diff
@@ -22,6 +23,10 @@ const textColors = {
   signalAmberText: tokens.colors.signalAmberText,
 } as const
 
+// Only a fraction of the space tokens are used (the most used ones) so we can reduce
+// the number of classes generated by sprinkles.
+const spaceTokensSubset = createSubset(tokens.space, ['xs', 'sm', 'md', 'lg', 'xl'])
+
```
alebedev, line 29: How small is resulting CSS file compared to what we had originally?

packages/ui/src/theme/sprinkles.css.ts
```diff
@@ -42,17 +47,15 @@ const responsiveProperties = defineProperties({
   defaultCondition: '_',
   properties: {
```
alebedev, line 48: How often do we use responsive paddings and font size?  Maybe we can optimize further by making all sprinkles non-responsive and assuming responsive styles would be written in css.ts files?

packages/ui/src/theme/sprinkles.css.ts
```diff
@@ -42,17 +47,15 @@ const responsiveProperties = defineProperties({
...
+    marginBottom: spaceTokensSubset,
+    marginLeft: spaceTokensSubset,
+    marginRight: spaceTokensSubset,
+    paddingTop: spaceTokensSubset,
+    paddingBottom: spaceTokensSubset,
+    paddingLeft: spaceTokensSubset,
+    paddingRight: spaceTokensSubset,
+    gap: spaceTokensSubset,
   },
   shorthands: {
```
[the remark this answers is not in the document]
gustaveen: If using as style props I think it's nice with shorthands :)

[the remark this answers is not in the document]
guilhermespopolin: I don't think they generate additional classes. They're just used during run time to match the correct classes.

packages/ui/src/theme/sprinkles.css.ts
guilhermespopolin: After latest changes it has 481 lines now

packages/ui/src/theme/sprinkles.css.ts
guilhermespopolin: Good point. I've kept only font size as responsive properties and we're now only supporting the actually used breakpoints: _, sm, md and lg.

packages/ui/src/theme/sprinkles.css.ts
```diff
@@ -61,14 +54,23 @@ const responsiveProperties = defineProperties({
     mr: ['marginRight'],
     mt: ['marginTop'],
     mb: ['marginBottom'],
-    m: ['margin'],
```
guilhermespopolin, line 64: As @gustaveen suggested, have removed 'margin' and 'padding' properties as well. We rarely use them (never used them, actually).

PR #4367, opened 2024-06-12
Fix storyblok cache revalidation

alebedev:
## Describe your changes

Fix CMS page revalidation:
- Introduce delays into revalidation webhook. This is a workaround for known cache invalidation issues on Vercel. I'm not 100% sure what happens, but likely hypothesis is that cache invalidation's network request looses a race with returning from our API handler and get aborted
- Fix problem where we went to check for new `cv` with old `cv` in request.  This always failed to show any changes. Problem is not new, but it got much worse recently, when Vercel reduced restart rate for serverless functions - old `cv` was kept in memory and if page handler was restarted right before invalidation, it all worked
- Reduce storyblok.cv TTL to 1 minute. Request is very cheap is we don't save anything by making it less frequently

Overall, these fixes should provide
- Production and staging: few seconds of delay between publishing Storyblok changes and them appearing on the site
- Preview apps: ~1 minute delay (since we don't get webhooks there)

### DIFF
--- a/apps/store/src/app/[locale]/[[...slug]]/page.tsx
+++ b/apps/store/src/app/[locale]/[[...slug]]/page.tsx
@@ -1,7 +1,7 @@
 import { StoryblokStory } from '@storyblok/react/rsc'
 import type { Metadata } from 'next'
 import { notFound } from 'next/navigation'
-import type { ReactNode } from 'react';
+import type { ReactNode } from 'react'
 import { cache } from 'react'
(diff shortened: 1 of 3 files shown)


alebedev:
* **#4367** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4367?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

Youakeem (approved):

guilhermespopolin (approved):

PR #4370, opened 2024-06-13
Refactor experiment config

gustaveen:
## Describe your changes
- Refactor experiments into an object with `id` as key
- Better logging of assigned experiment

### DIFF
--- a/apps/store/src/services/Tracking/experiment.helpers.ts
+++ b/apps/store/src/services/Tracking/experiment.helpers.ts
@@ -1,31 +1,30 @@
 import { getCookie } from 'cookies-next'
-import type {
-  ExperimentConfig} from './experiment.constants';
 import {
   CURRENT_EXPERIMENTS,
+  type CurrentExperiments,
   EXPERIMENT_COOKIE_NAME,
-  Experiments,
   type Experiment,
+  type ExperimentConfig,
   type ExperimentVariant,
+  Experiments,
 } from './experiment.constants'
 
-export const getCurrentExperiments = (): Array<Experiment> | undefined => {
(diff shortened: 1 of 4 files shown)


apps/store/src/services/Tracking/experiment.helpers.ts
```diff
@@ -1,31 +1,28 @@
...
-  EXPERIMENT_COOKIE_NAME,
-  Experiments,
-  type Experiment,
-  type ExperimentVariant,
+  CurrentExperiments,
+  Experiment,
+  ExperimentConfig,
+  ExperimentVariant,
 } from './experiment.constants'
+import { CURRENT_EXPERIMENTS, EXPERIMENT_COOKIE_NAME, Experiments } from './experiment.constants'
```
guilhermespopolin, line 8: Should we use a single import statement for these?

guilhermespopolin (approved):
Nice additions :)

apps/store/src/services/Tracking/experiment.helpers.ts
gustaveen: Ah, that's my eslint not working properly 😔

gustaveen: Might be my eslint not working properly. But shouldn't the new rule split up types and imports?

guilhermespopolin: I don't know for sure now 😬 I mean, I remember the new rule being keep everything on a single import statement

gustaveen: Got it working now, so lint should fix it 👍

PR #4380, opened 2024-06-18
Update button component and theme colors to gollow new UI kit

gustaveen:
## Describe your changes
Update button component with:
- Update colors (green button now have better contrast)
- Use semantic tokens for backgrounds
- Remove shadow on all buttons

The green swatches have a slight update but checked that all usage looked good. 

Screenshot 2024-06-18 at 14 15 17

## Justify why they are needed
Updated design system 
## Checklist before requesting a review

- [ ] I have performed a self-review of my code

PR #4383, opened 2024-06-18
Remove default 100% width on large Button variant

gustaveen:
## Describe your changes
Large `Button` has width set to 100% by default, always been a bit odd behaviour. Let's remove it an use a prop for it when needed and set a min width according to UI kit.

Screenshot 2024-06-18 at 17 23 44

### DIFF
--- a/packages/ui/src/components/Button/Button.css.ts
+++ b/packages/ui/src/components/Button/Button.css.ts
@@ -191,7 +191,7 @@ const SIZE_STYLES = {
   },
   large: {
     height: HEIGHT.large,
-    width: '100%',
+    minWidth: '18.75rem',
     paddingInline: tokens.space.xl,
     fontSize: tokens.fontSizes.md,
     textAlign: 'center',
--- a/apps/store/src/blocks/ButtonBlock.tsx
+++ b/apps/store/src/blocks/ButtonBlock.tsx
@@ -13,6 +13,7 @@ export type ButtonBlockProps = SbBaseBlockProps<{
   link: LinkField
   variant: ComponentProps<typeof Button>['variant']
   size: ComponentProps<typeof Button>['size']
+  fullWidth?: boolean
 }>
 
 export const ButtonBlock = ({ blok, nested }: ButtonBlockProps) => {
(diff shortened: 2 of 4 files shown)


gustaveen:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4383?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#4383** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4383?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#4380** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4380?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

packages/ui/src/components/Button/Button.css.ts
```diff
@@ -191,7 +191,7 @@ const SIZE_STYLES = {
   },
   large: {
     height: HEIGHT.large,
-    width: '100%',
+    minWidth: '18.75rem',
```
Youakeem, line 194: I am afraid this would introduce problems with responsiveness

alebedev (approved):

packages/ui/src/components/Button/Button.css.ts
gustaveen: I wonder if we would ever have a use case where we have a screen smaller than 300px and hence the button would overflow? Since all button layouts are one column in mobile. Or am I missing some use case? @Youakeem

Youakeem: I don't think there are specific use cases, I was thinking more in general since there could be paddings, margins, nested cards ...etc

If you think it would be fine then I am ok, we can give it another look if it causes an issue. 👍

gustaveen:
### Merge activity

* **Jun 19, 5:30 AM EDT**: @gustaveen started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4383).

PR #4396, opened 2024-06-24
Refactor TextField to vanilla-extract

gustaveen:
## Describe your changes
Refactor TextField to vanilla-extract

I also renamed `variant` to the more logical name `size` since variants was handling small or large sizing. This also makes it map better to Figma and we might add variants that controls the design (wether it's on a dark or light background)

TextField in Storybook: https://62ea2a095003502af2e247b8-jkzzykluig.chromatic.com/?path=/story/inputs-textfield--large&globals=viewport:iphonese2

Screenshot 2024-06-24 at 14 23 38


gustaveen:
* **#4396** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4396?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

apps/store/src/components/TextField/TextField.css.ts
```diff
@@ -0,0 +1,140 @@
...
+  },
+})
+
+export const largeInputWrapper = style([
+  {
+    position: 'absolute',
+    bottom: '0.625rem',
+    width: '100%',
+  },
+  sprinkles({ display: 'flex', alignItems: 'center', gap: 'xs', paddingRight: 'md' }),
```
Youakeem, line 92: I think we can use the `stack` pattern directly

apps/store/src/components/TextField/TextField.css.ts
```diff
@@ -0,0 +1,140 @@
...
+      display: 'none',
+    },
+
+    '&[data-warning=true]': {
+      animation: `${warningAnimation} 1.5s cubic-bezier(0.2, -2, 0.8, 2) 1`,
+    },
+  },
+})
+
+export const largeWrapper = style([baseWrapper, { height: '4.5rem' }])
```
Youakeem, line 45: Can we use `styleVariants` for the size?

gustaveen: For sure, was just doing a 1 to 1 refactoring 👍

apps/store/src/components/TextField/TextField.css.ts
```diff
@@ -0,0 +1,140 @@
...
+
+  fontSize: tokens.fontSizes.xl,
+  selectors: {
+    '&[data-variant=small]': { fontSize: tokens.fontSizes.lg },
+
+    '&&[data-disabled=true]': {
+      color: tokens.colors.textTertiary,
+    },
+
+    [`${largeWrapper}:focus-within > &, ${largeWrapper}[data-active=true] > &`]: {
```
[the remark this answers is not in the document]
gustaveen: Hm not sure I follow. We need to react to the `focus` state on the input somehow like in this case with `focus-within`.  

Or do you mean using css variables for the transform and assigning it inline?

apps/store/src/components/TextField/TextField.css.ts
gustaveen: Mm I was thinking about that, but we still would need sprinkle for padding right? Then I thought it would be cleaner to just use sprinkles in this case, and not both sprinkles and yStack :)

Youakeem: The patterns work the same as sprinkles so no need to use both, it should accept anything sprinkles does.

apps/store/src/components/TextField/TextField.css.ts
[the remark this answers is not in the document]
Youakeem: If I understand correctly we change some label styles based on the wrapper size, if that's the case, then I think we can use variables and set those in the wrapper variants.

On reacting to focus, I think it might be a little more straight forward to handle that in the parent instead of the label `.wrapper > .label` instead of `.label .wrapper &`.

Then the label wouldn't really care and it would just behave according to it's wrapper.

apps/store/src/components/TextField/TextField.css.ts
gustaveen: Aha, didn't know!  👍 Just wondering, is that intentional for patterns to accept all sprinkles, thought it was just a sub set related to the pattern. Padding feels fine though.

Youakeem: I think of them as sprinkles with some predefined values for ease of use.

guilhermespopolin (approved):
Nice work with this :)

apps/store/src/components/TextField/TextField.css.ts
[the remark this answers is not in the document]
gustaveen: It's updated so the label has size variants as well so it's not dependant on the wrapper size, only the the focus or active state. To me that feels easier to comprehend, that the styles for the current element is set with all the other styles for label. 

I created a draft based on variables set in the parent https://github.com/HedvigInsurance/racoon/pull/4402/files#diff-d24acf416a5465ff7dfa730db83c241eaa623624016361120c80d827dfd8f4cdR116-R117

Youakeem (approved):

gustaveen:
### Merge activity

* **Jun 26, 4:49 AM EDT**: @gustaveen started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4396).

PR #4451, opened 2024-07-05
Stop sending another full ProductStory to the client when we need just 3 fields from it

alebedev:
## Describe your changes

Send less data to `ProductPageDataProvider`

Needed for future refactoring where `story` would go deeper into page tree


gustaveen (approved):
Nice!

PR #4489, opened 2024-07-15
Fix video control buttons and handle visually hidden label properly

gustaveen:
## Describe your changes
- Pass icons properly to Video control buttons
- Add `visuallyHidden` class to global css
- Add prop to pass hidden text to buttons for accesibility

Icons for the video controls were not passed as Icons so it got a slight offset. 

### DIFF
--- a/apps/store/src/components/Video/Video.tsx
+++ b/apps/store/src/components/Video/Video.tsx
@@ -3,7 +3,7 @@ import { assignInlineVars } from '@vanilla-extract/dynamic'
 import clsx from 'clsx'
 import { useInView } from 'framer-motion'
 import React, { useCallback, useEffect, useId, useRef, useState } from 'react'
-import { PlayIcon, PauseIcon, Button } from 'ui'
+import { PlayIcon, PauseIcon, Button, visuallyHidden } from 'ui'
 import { useDialogEvent } from '@/utils/dialogEvent'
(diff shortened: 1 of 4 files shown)


packages/ui/src/global.css.ts
```diff
@@ -73,4 +73,17 @@ globalStyle('#__next', {
   isolation: 'isolate',
 })
 
+// Visually hide content but make it available to screen readers
+// https://www.a11yproject.com/posts/how-to-hide-content/
+globalStyle('.visuallyHidden', {
```
alebedev, line 78: Nice!

alebedev: Is it important that class is called `visuallyHidden`?  Otherwise we can export it as non-global thing that can be imported

alebedev (approved):
Looks good

packages/ui/src/global.css.ts
gustaveen: Not at all :) I couldn't decide, but maybe that's better 👍

gustaveen:
### Merge activity

* **Jul 16, 4:10 AM EDT**: @gustaveen started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4489).

PR #4453, opened 2024-07-05
GEN-2421 - feat(app router confirmation page): support swiching and member partner data

guilhermespopolin:
## Describe your changes

Seems like a lot of changes but a bunch of them are just small updates due API changes. Here is what changed:

* Update `getApolloClient` function. Now server components can instantiate apollo client with `cookies()`. That will be used by adding links that handle shop session id and authorization headers (serverDynamicHeadersLink). Observe that I had to update `app-router/persit.ts` - before nextjs `cookies()` were being imported and used directly there. The issue with it is that we don't want those dynamic functions for every instance. Adding them for CMS pages, for example, makes the app break as we have a we don't want CMS pages being rendered dynamically and by using dynamic functions during their rendering process (headers() or cookies()) we'd end up making the whole route be rendered dynamically.
* Had update `fetchJson` function. Now it can receive a locale instead of trying to figure it out by itself. Old implementation breaks when using RSC because those don't have access to i18n.
* Update app router migration page so it also fetches switching and member partner data as it's page router counter part do.
  
## Justify why they are needed

The idea is to get my hands dirty with Nextjs app router migration by migrating _ConfirmationPage_: `pages/[locale]/confirmation/[shopSessionId]` --> `app/[locale]/new/confirmation/[shopSessionId]` (`page.tsx` file)

PR #4492, opened 2024-07-16
GEN-2432 - Checkout page v2 - display cart items

guilhermespopolin:
## Describe your changes

* Refactor `rscClient.ts`: Changed _getApolloClient_ function so it also exposes `PreloadQuery` as [library authors](https://www.npmjs.com/package/@apollo/experimental-nextjs-app-support)  mentioned in the docs. As a side effect I had to change several places that uses `rscClient` but they're straightforward changes.
* Refactor `authApi/app-router/persist.ts`:  Revert some of the changes added [here](https://github.com/HedvigInsurance/racoon/pull/4453). To avoid breaking CMS builds, I've made some changes to make sure they didn't rely on dynamic functions, even indirectly. A better option however would be use `export const dynamic = 'force-static'` (instead of `error`) and keep `authApi` simple by importing NexJs `cookies()` directly.
* Feat: Added basic `Checkout` v2 page which displays cart entries.

https://github.com/user-attachments/assets/a7bb7ec8-6ec3-464b-b48f-74448189c856

First step into having a checkout+cart page.

[Figma file](https://www.figma.com/design/p1TEzLLVqJB9gKDNTDVGU4/Hedvig.com?node-id=472-12230&t=yFgWz55AmO4Ug3ql-4)


guilhermespopolin:
* **#4492** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4492?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

apps/store/src/app/[locale]/[[...slug]]/page.tsx
```diff
@@ -154,7 +154,7 @@ export async function generateStaticParams({
 export const dynamicParams = true
 // Make sure we don't accidentally use dynamic rendering
 // https://nextjs.org/docs/app/api-reference/file-conventions/route-segment-config#dynamic
-export const dynamic = 'error'
+export const dynamic = 'force-static'
```
Comment (guilhermespopolin) [inline apps/store/src/app/[locale]/[[...slug]]/page.tsx:157]: Changing to `force-static` here so we're making sure CMS pages are built as static pages but they can still reference code that make usage of nextjs dynamic functions like `cookies()` and `header()` without breaking.

apps/store/src/appComponents/providers/ApolloProvider.tsx
```diff
@@ -3,9 +3,9 @@
 import { ApolloLink } from '@apollo/client'
 import {
   ApolloNextAppProvider,
-  NextSSRApolloClient,
-  NextSSRInMemoryCache,
-} from '@apollo/experimental-nextjs-app-support/ssr'
+  ApolloClient,
+  InMemoryCache,
+} from '@apollo/experimental-nextjs-app-support'
```
guilhermespopolin, line 8: Here I'm just changing the imports as they're deprecating some of them.

apps/store/src/services/apollo/app-router/rscClient.ts
```diff
@@ -44,7 +35,7 @@ function makeGetApolloClient(locale: RoutingLocale, cookies?: NextCookiesStore)
     })
   })
 
-  return getClient
+  return { getApolloClient, PreloadQuery }
```
guilhermespopolin, line 38: Changing a little bit the function for getting apollo client in the context of RSC. This now also exposes `PreloadQuery` component which can be used to pre-populate apollo cache client for client components. That's basically the same thing we achieve with our `addApolloState` function

apps/store/src/services/authApi/app-router/persist.ts
```diff
@@ -1,5 +1,5 @@
 import { type OptionsType } from 'cookies-next/lib/types'
```
guilhermespopolin, line 1: I had changed this file before by making its functions to receive NextJs cookies store instead of import it directly here. The reasoning behind that was because we had `export const dynamic = 'error'` for CMS pages before which would break the build if CMS pages rely, even indirectly, on dynamic functions. I didn't realize at the time we could just change CMS pages to use `export const dynamic = 'force-static'` instead which is a better solution. So for simplicity sake, I'm reverting that change and import `cookies` directly here.

apps/store/src/app/[locale]/new/checkout/CheckoutPage.tsx
```diff
@@ -0,0 +1,72 @@
...
+  const tracking = useTracking()
+  const { showError } = useAppErrorHandleContext()
+
+  // TODO: add a custom merge function that makes sure to update the cache accordingly
+  const [removeCartItem] = useCartEntryRemoveMutation({
+    awaitRefetchQueries: true,
+    onError(error) {
+      datadogLogs.logger.error('Checkout Page | Failed to remove offer from cart', {
+        shopSessionId: shopSession.id,
+        // TODO: find a way to add offer.id here as well. Maybe from error.extraInfo
```
Comment (alebedev) [inline apps/store/src/app/[locale]/new/checkout/CheckoutPage.tsx:29]: Not sure this TODO is still relevant

apps/store/src/app/[locale]/new/checkout/CheckoutPage.tsx
```diff
@@ -0,0 +1,72 @@
...
+import { layout, shopBreakdowSection } from './CheckoutPage.css'
+
+type Props = { shopSessionId: string }
+
+export function CheckoutPage({ shopSessionId }: Props) {
+  const shopSession = useShopSession({ shopSessionId })
+  const tracking = useTracking()
+  const { showError } = useAppErrorHandleContext()
+
+  // TODO: add a custom merge function that makes sure to update the cache accordingly
```
Comment (alebedev) [inline apps/store/src/app/[locale]/new/checkout/CheckoutPage.tsx:23]: What if we just ask for `id` on cart items?  This should take care of merge warning from Apollo

apps/store/src/app/[locale]/new/checkout/CheckoutPage.tsx
```diff
@@ -0,0 +1,72 @@
+'use client'
```
Comment (alebedev) [inline apps/store/src/app/[locale]/new/checkout/CheckoutPage.tsx:1]: Can we make top component server side?  Interactive and animated components inside should stage client components, but is there a reason for full page to be client side?

apps/store/src/app/[locale]/new/confirmation/[shopSessionId]/page.tsx
```diff
@@ -17,7 +16,8 @@ type Props = { params: Params }
 export default async function Page({ params }: Props) {
   const confirmationStory = await fetchConfirmationStory(params.locale)
 
-  const apolloClient = getApolloClient({ locale: params.locale, cookies: cookies() })
+  const { getApolloClient } = setupApolloClient({ locale: params.locale })
+  const apolloClient = getApolloClient()
   const shopSessionService = setupShopSession(apolloClient)
   const [shopSession, switchingData, memberPartnerData] = await Promise.all([
     shopSessionService.fetchById(params.shopSessionId),
```
Comment (alebedev) [inline apps/store/src/app/[locale]/new/confirmation/[shopSessionId]/page.tsx:23]: <i class='graphite__hidden'>[Re: line 29]<br/><br/></i>

Can we move SuccessAnimation inside? It feels like presentation detail, it should probably wrap outermost presentation component, not full page<span class='graphite__hidden'><br/><br/>See this comment inline on <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4492?utm_source=unchanged-line-comment">Graphite</a>.</span>

apps/store/src/app/[locale]/new/checkout/CheckoutPage.tsx
Comment (guilhermespopolin) [reply apps/store/src/app/[locale]/new/checkout/CheckoutPage.tsx:23]: 🤔 I'd need to test it but I think we'd have the same problem. This is a "regular" warning for fields that holds a list. I think that even if we had a list of ids, whenever we receive a smaller list as the result of a mutation apollo would still raises that warning. I mean, it's not an issue for us actually. They raise this because it might be a problem but for this particular case we do actually want the default behavior which is the replacement of the whole list. I think it exists for cases like pagination where you probably want to concatenate the two lists instead of replacing them. You know what, I think we can ignore it. Those are silenced in prod and there's no actual problem here. It's a little bit noisy yes, but not the end of the world.

Screenshot 2024-07-16 at 11 52 46

apps/store/src/app/[locale]/new/confirmation/[shopSessionId]/page.tsx
Comment (guilhermespopolin) [reply apps/store/src/app/[locale]/new/confirmation/[shopSessionId]/page.tsx:23]: Yes, nice catch. Doing it.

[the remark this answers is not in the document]
Comment (guilhermespopolin) [reply apps/store/src/app/[locale]/new/confirmation/[shopSessionId]/page.tsx:23]: 🤔 now that I'm thinking more about it, I think we don't need that `dynamic` there. It will be dynamic because it uses `cookies()` (apollo client will use it to fetch shopSession).

image

Would you say the following way of thinking about your app makes sense?
* Shop session is utmost important for us. We rely on cookies to fetch it so "by default" our app is dynamic meaning every page that you create that instantiate will be dynamic because we fetch shop session on the root layout. If you want to create a dynamic page you need to make sure to add `dynamic = 'force-static'` to it.

apps/store/src/app/[locale]/new/checkout/CheckoutPage.tsx
Comment (guilhermespopolin) [reply apps/store/src/app/[locale]/new/checkout/CheckoutPage.tsx]: Updated it.

alebedev (approved):

guilhermespopolin:
### Merge activity

* **Jul 18, 4:08 AM EDT**: @guilhermespopolin started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4492).

PR #4504, opened 2024-07-18
fix(widget): form auto completion with suggested data

guilhermespopolin:
## Describe your changes

Ok, we were having an issue before where widget's price calcuator page was not pre-populating 'your address' form section. It was solved by this [PR](https://github.com/HedvigInsurance/racoon/pull/4435), which introduced another couple of issues:

1. When completing a purchase in Hedvig-dot-com, one could still see the number of items you have purchased in the header
2. When completing a purchase in widget and opening a tab with hedvig-dot-com, one would have it loaded with widget's shop session

This is the second attempt, which solves all of them. Here's the reasoning behind it:

We have two possible sources for shop session id in our app (and while debugging I think it makes sense to have 2): the URL and/or a cookie. The cookie option is the fallback one. If provided in the URL we should use it despite what we have in a cookie. `ShopSessionContext` is responsible to provide us the shop session and therefore should be initialised following the same criteria.

The issue we were having before was that even though shop session id was present for widget URLs,`ShopSessionContext` was using what we have as a cookie. Therefore, when entering your ssn in the widget we were updating the wrong shop session. My first attempt was to synchronyze both: make sure the cookie always have the same what we have in the URL, when provided. But as mentioned that causes other issues.

### DIFF
--- a/apps/store/src/services/shopSession/ShopSessionContext.tsx
+++ b/apps/store/src/services/shopSession/ShopSessionContext.tsx
@@ -46,16 +46,15 @@ export const useShopSession = (): ShopSessionResult => {
 
 const useShopSessionContextValue = (initialShopSessionId?: string) => {
   const { countryCode } = useCurrentCountry()
-  const apolloClient = useApolloClient()
-  const shopSessionServiceClientSide = useMemo(() => {
-    const service = setupShopSessionServiceClientSide(apolloClient)
-    if (initialShopSessionId) service.saveId(initialShopSessionId)
 
-    return service
-  }, [apolloClient, initialShopSessionId])
+  const apolloClient = useApolloClient()
+  const shopSessionServiceClientSide = useMemo(
+    () => setupShopSessionServiceClientSide(apolloClient),
+    [apolloClient],
+  )
 
   const [shopSessionId, setShopSessionId] = useState(
-    shopSessionServiceClientSide.shopSessionId() ?? initialShopSessionId,
+    initialShopSessionId ?? shopSessionServiceClientSide.shopSessionId(),
   )
 
   const queryResult = useShopSessionQuery({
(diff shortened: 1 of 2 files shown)


guilhermespopolin:
* **#4504** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4504?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

apps/store/src/services/shopSession/ShopSessionContext.tsx
```diff
@@ -96,7 +95,10 @@ const useShopSessionContextValue = (initialShopSessionId?: string) => {
 
   // Fetch client-side, service ensures it only happens once
   useEffect(() => {
-    if (isBrowser()) {
+    // We want to make sure the shopSession returned by this hook is:
+    // 1. The one which is is `initialShopSessionId`, when it's provided
```
gustaveen, line 99:
```suggestion
    // 1. The one which is `initialShopSessionId`, when it's provided
```

apps/store/src/services/shopSession/ShopSessionContext.tsx
```diff
@@ -46,16 +46,15 @@ export const useShopSession = (): ShopSessionResult => {
...
-  }, [apolloClient, initialShopSessionId])
+  const apolloClient = useApolloClient()
+  const shopSessionServiceClientSide = useMemo(
+    () => setupShopSessionServiceClientSide(apolloClient),
+    [apolloClient],
+  )
 
   const [shopSessionId, setShopSessionId] = useState(
-    shopSessionServiceClientSide.shopSessionId() ?? initialShopSessionId,
+    initialShopSessionId ?? shopSessionServiceClientSide.shopSessionId(),
```
gustaveen, line 57: So `initialShopSessionId` the id provided in the URL?

gustaveen (approved):
Looks good to me!

apps/store/src/services/shopSession/ShopSessionContext.tsx
guilhermespopolin: Yes. For "pages router" based pages we get shopSessionId provided in the URL for dynamic pages and use it to initialize `ShopSessionContext`. Take confirmation page as an example: We return the shop session from the URL [here](https://github.com/HedvigInsurance/racoon/blob/main/apps/store/src/pages/%5Blocale%5D/confirmation/%5BshopSessionId%5D.tsx#L54) and then intialize `ShopSessionContext` with it [here](https://github.com/HedvigInsurance/racoon/blob/main/apps/store/src/pages/_app.tsx#L107). So basically, whenever we intialise `ShopSessionContext` with a shop session id, it should take precedence over the one we store at cookie level.

guilhermespopolin:
### Merge activity

* **Jul 19, 5:41 AM EDT**: @guilhermespopolin started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4504).

PR #4544, opened 2024-07-30
GEN-2432 - fix(checkout-v2): invalidate router cache for checkout page

guilhermespopolin:
## Describe your changes

* Restructure how _Checkout_ page v2 is rendered so we don't have issues with caching

**The issues**

We have a couple of caching issues that comes from how _CheckoutPage_ is being rendered until now.

1. Recently added offers might not get shown in the _Checkout Page_

https://github.com/user-attachments/assets/b5c59f40-b53d-4421-9cbe-38f1536182f7

2. Checkout page _empty cart_ UI doesn't get shown when removing the last item from the cart

https://github.com/user-attachments/assets/39c59dff-3078-4ba0-9839-a3a8eabb4fe0

**Why they happen**

Current implementation of _CheckoutPage V2_ decides at the server level (via RSC) if we should show the _Empty cart UI_ or the _Shop breakdown_ based on the size of the cart entries. That works fine when browser navigating to checkout page but since Nextjs caches routes (even dynamic ones) at the client level the following can happen:

1. You visit checkout page and get's the _Empty cart_ UI. At this point Nextjs caches RSC payload for that route at client level (Router cache) for a certain amount of time (30 seconds if I'm mistaken)
2. You add a offer to the cart a client side navigate to checkout page. Cached RSC payload will be used which only includes the _Empty Cart UI_.

Additionally by keeping the logic between showing the _Empty Cart UI_ of the _Shop breakdown_ at the server level will cause the problem of not showing the _Empty Cart UI_ when removing the last offer because for that to happen we'd need to request the checkout page again. Since we keep mutations at the client side it makes sense to keep such decisions at the client side. In any case we're dealing with a highly dynamic and interactive page.

**The solution**

I've changed how _Checkout Page_ gets rendered by moving the decision between UI to the client via client components. Since everything is done the client side we'll not have caching issues as well because even if Nexjts caches RSC payload for the checkout page, that payload always refer to the that same client component that handle mutations and the decision between _Empty UI_ or _Shop breakdown_

[Oficial docs](https://nextjs.org/docs/app/building-your-application/caching#invalidation-1)

### DIFF
--- a/apps/store/src/app/[locale]/new/checkout/BonusOffer.tsx
+++ b/apps/store/src/app/[locale]/new/checkout/BonusOffer.tsx
@@ -2,10 +2,9 @@
 
 import { QuickAddOfferContainer } from '@/components/QuickAdd/QuickAddOfferContainer'
 import { useBonusOffer } from '@/components/QuickAdd/useBonusOffer'
-import { useShopSessionSuspense } from '@/services/shopSession/app-router/useShopSessionSuspense'
+import { type ShopSessionFragment } from '@/services/graphql/generated'
 
-export function BonusOffer({ shopSessionId }: { shopSessionId: string }) {
-  const shopSession = useShopSessionSuspense({ shopSessionId })
+export function BonusOffer({ shopSession }: { shopSession: ShopSessionFragment }) {
   const offerRecommendation = useBonusOffer()
 
   if (!offerRecommendation) {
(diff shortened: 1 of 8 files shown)


alebedev (commented):
Can you explore alternative solutions? This one does not look like it's going to be easy to scale. For example, would be nice to disable caching specifically for Apollo query that asks for cart items, in this case just `router.refresh()` could be enough

alebedev (approved):

guilhermespopolin:
### Merge activity

* **Aug 1, 4:26 AM EDT**: @guilhermespopolin started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4544).

PR #4579, opened 2024-08-09
GEN-2470 - styles(combobox): support size variants (small, medium, large)

guilhermespopolin:
## Describe your changes

* Combobox: support size variants: 'small', 'medium' and 'large'

image

With that change, Breed Selector now has the same sizing styles and other inputs on new Price Calculator pages:

image

Some of the changes included in Breed Selector [redesign](https://www.figma.com/design/5kmmDdh6StpXzbEfr7WevV/Hedvig-UI-Kit?node-id=15404-12523).

### DIFF
--- a/apps/store/src/components/Combobox/Combobox.css.ts
+++ b/apps/store/src/components/Combobox/Combobox.css.ts
@@ -1,81 +1,94 @@
-import { style } from '@vanilla-extract/css'
+import { style, createVar, styleVariants } from '@vanilla-extract/css'
 import { tokens, hoverStyles } from 'ui'
 import { zIndexes } from '@/utils/zIndex'
 
-export const wrapper = style({
-  position: 'relative',
+const wrapperBase = style({
   isolation: 'isolate',
-  selectors: {
-    '&[data-expanded=true]': {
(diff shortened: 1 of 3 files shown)


alebedev (changes requested):
Can we do it with CSS classes and specific selectors, without relying on HTML attributes? I think attributes should be the last solution when everything else fails. Here we're targeting things inside single component, so there's probably no need for attributes

guilhermespopolin:
> Can we do it with CSS classes and specific selectors, without relying on HTML attributes? I think attributes should be the last solution when everything else fails. Here we're targeting things inside single component, so there's probably no need for attributes

Aha, I see. The thing I like about this approach is that every UI related change/decision is under CSS realm. But I can change it by adding specific selectors 👍🏻

alebedev (approved):
Great!

guilhermespopolin:
### Merge activity

* **Aug 12, 9:08 AM EDT**: @guilhermespopolin started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4579).

PR #4599, opened 2024-08-15
refact(yStack/xStack): make arguments optional

guilhermespopolin:
## Describe your changes

* Update xStack/yStack apis so parameters became optional. Then `yStack()` would mean the same as `yStack({})` without the necessity of an empty object.
* Make `gap` opt in to `stack` by removing default `gap`. Then `yStack()` would be the same gapwise as `yStack({gap:'none'})`

Subjective. I think that API is easier to work with. Let me know what you think.

### DIFF
--- a/apps/store/src/features/priceCalculator/PurchaseFormV2.tsx
+++ b/apps/store/src/features/priceCalculator/PurchaseFormV2.tsx
@@ -45,7 +45,7 @@ export function PurchaseFormV2() {
       )
     case 'viewOffers':
       return (
-        <div className={yStack({})} style={{ gap: '2.75rem' }}>
+        <div className={yStack({ gap: 'md' })} style={{ gap: '2.75rem' }}>
           <InsuranceDataForm />
           <OfferPresenterV2 />
         </div>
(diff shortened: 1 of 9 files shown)


guilhermespopolin:
* **#4599** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4599?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

apps/store/src/features/priceCalculator/PurchaseFormV2.tsx
```diff
@@ -45,7 +45,7 @@ export function PurchaseFormV2() {
       )
     case 'viewOffers':
       return (
-        <div className={yStack({})} style={{ gap: '2.75rem' }}>
+        <div className={yStack()} style={{ gap: '2.75rem' }}>
```
alebedev, line 48: I think default gap above was 'md'. Having it `none` is better, I agree. Let's just update old uses to make sure visual look did not change

alebedev (approved):
Looks good, one comment about default value of old usage

apps/store/src/features/priceCalculator/PurchaseFormV2.tsx
guilhermespopolin: Have update it.

PR #4592, opened 2024-08-14
Use atomFamily to keep product-specific atoms scoped

alebedev:
## Describe your changes

Demo of 2nd alternative approach to scoping jotai atoms to React subtree. Original problem and 1st solution can be seen in https://github.com/HedvigInsurance/racoon/pull/4581

Used in next following PR to show multiple product variants next to each other

## Checklist before requesting a review

- [ ] I have performed a self-review of my code

PR #4626, opened 2024-08-21
RND-855 - fix(widget quote creation): proper initialize product data atom for sign page

guilhermespopolin:
## Describe your changes

* Fetch `productData` for widget calculate price page as we do for sign page.

The issue happens because during the flow we initialize productDataAtom without variants data for calculate price page and, while sign page fetches productData containing variants, productDataAtom never gets initialized with variants again. The issue goes away if we refresh the page because in that case productDataAtom is proper initialized. [Before](https://github.com/HedvigInsurance/racoon/pull/4592/files#diff-651670b13cc543b64b1e9754b49c434b56d73c21fa42e89788daa637b23cc225L45-L48) we're using `dangerouslyForceHydrate` option from Jotai so that's why it was working.

This solves it but it's still fragile as even though we're talking about different pages, the fact they share the same data dependency and that we're using jotai for it doesn't exclude the potential of having the same initialization issue again in the future unless we get `dangerouslyForceHydrate` back or use a whole different approach.

### DIFF
--- a/apps/store/src/pages/[locale]/widget/run/[flow]/[shopSessionId]/[priceIntentId]/calculate-price.tsx
+++ b/apps/store/src/pages/[locale]/widget/run/[flow]/[shopSessionId]/[priceIntentId]/calculate-price.tsx
@@ -4,6 +4,7 @@ import Head from 'next/head'
 import { useTranslation } from 'next-i18next'
 import { serverSideTranslations } from 'next-i18next/serverSideTranslations'
 import type { ComponentPropsWithoutRef, ReactNode } from 'react'
+import { fetchProductData } from '@/components/ProductData/fetchProductData'
 import { type ProductData } from '@/components/ProductData/ProductData.types'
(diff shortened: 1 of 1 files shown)


guilhermespopolin:
* **#4626** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4626?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

Youakeem (approved):

gustaveen (approved):
Tested and validated, good catch @guilhermespopolin 👌
Let's evaluate on the best approach to avoid inconsistency with atoms not being initialised properly

guilhermespopolin:
### Merge activity

* **Aug 21, 4:58 AM EDT**: @guilhermespopolin started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4626).

PR #4632, opened 2024-08-22
RND-858 - fix(checkout): display empty cart instead of 404 when loading checkout page without shop session cookie

guilhermespopolin:
## Describe your changes

* Display empty cart instead of 404 when loading checkout page without shop session cookie.

**Before**

https://github.com/user-attachments/assets/3e53e52d-f353-457a-bba8-54e45b391c0f

**After**

https://github.com/user-attachments/assets/719645c9-d027-4c3c-b82e-a9038eda9716

Better UX


guilhermespopolin:
* **#4632** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4632?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

apps/store/src/app/[locale]/new/checkout/actions.ts
```diff
@@ -0,0 +1,24 @@
+'use server'
```
Comment (alebedev) [inline apps/store/src/app/[locale]/new/checkout/actions.ts:1]: This looks more complicated vs having server component that then uses `shopSessionService.getOrCreate()`. Was there anything wrong with that approach?

Comment (guilhermespopolin) [reply apps/store/src/app/[locale]/new/checkout/actions.ts:1]: That's the thing: RSC components can't use `getOrCreate` because `cookies().set` is [introduces a side effect](https://github.com/vercel/next.js/issues/51875#issuecomment-1610196633):

> This is expected, it is not allowed to directly edit the cookie during server rendering. Only cookies().get() is allowed (otherwise it's the same as directly calling cookies().set() in Server Components).
This ensures that server rendering doesn't cause side effects and the result can be properly cached.

Comment (alebedev) [reply apps/store/src/app/[locale]/new/checkout/actions.ts:1]: Great explanation! Can you put short summary into code comments to make it clear why we're using effects?

apps/store/src/app/[locale]/new/checkout/CheckoutPage.tsx
```diff
@@ -16,23 +17,17 @@ import { CheckoutForm } from './CheckoutForm'
...
+export function CheckoutPage({ locale }: { locale: RoutingLocale }) {
   const { t } = useTranslation(['cart', 'checkout'])
-  const shopSession = useShopSessionSuspense({
-    shopSessionId,
-    options: {
-      context: { fetchOptions: 'no-store' },
-    },
-  })
   const productMetadata = useProductMetadata()
+  const { data } = useShopSession()
```
Comment (alebedev) [inline apps/store/src/app/[locale]/new/checkout/CheckoutPage.tsx:23]: nit: ` const {shopSession} = useShopSession()` is how we usually write it. Hook puts `shopSession` higher in result for ease of access

alebedev (approved):
Looks good!

guilhermespopolin:
### Merge activity

* **Aug 23, 4:26 AM EDT**: @guilhermespopolin started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4632).

PR #4634, opened 2024-08-22
feat(connect payment): load content from CMS

guilhermespopolin:
## Describe your changes

Update slug for getting Connect Payment CMS page to match changes on storyblok.

I understand both need to be in sync and that could be a pain but so far we have a couple of cases and I don't expect that changing so often. If that chagnes then we can think on an alternative way.

As part of new designs for checkout page, CMS content was added to connect payment page

image


guilhermespopolin:
### Merge activity

* **Aug 23, 4:54 AM EDT**: @guilhermespopolin started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4634).

PR #4647, opened 2024-08-27
Create ProductPageDataProviderV2

gustaveen:
## Describe your changes
Since the new ProductPage won't have a price form, remove price template from the provider. 

Feed product page data to the new product page

### DIFF
--- a/apps/store/src/components/ProductPage/ProductPageDataProviderV2.tsx
+++ b/apps/store/src/components/ProductPage/ProductPageDataProviderV2.tsx
@@ -1,0 +1,14 @@
+'use client'
+import type { ReactNode } from 'react'
+import { type ProductPageData } from '@/components/ProductPage/getProductPageData'
+import { useSyncProductPageData } from './ProductPageDataProvider'
+
+type Props = {
+  children: ReactNode
+  productPageData: ProductPageData
+}
+
+export function ProductPageDataProviderV2({ children, productPageData }: Props) {
+  useSyncProductPageData(productPageData)
+  return children
+}
--- a/apps/store/src/app/[locale]/[[...slug]]/ProductCmsPageV2.tsx
+++ b/apps/store/src/app/[locale]/[[...slug]]/ProductCmsPageV2.tsx
@@ -1,6 +1,11 @@
 import { StoryblokStory } from '@storyblok/react/rsc'
 import { storyblokBridgeOptions } from '@/appComponents/storyblokBridgeOptions'
 import { DefaultDebugDialog } from '@/components/DebugDialog/DefaultDebugDialog'
+import { fetchProductData } from '@/components/ProductData/fetchProductData'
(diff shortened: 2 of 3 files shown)


gustaveen:
* **#4647** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4647?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

apps/store/src/components/ProductPage/ProductPageDataProviderV2.tsx
```diff
@@ -0,0 +1,29 @@
+'use client'
+import { atom, useAtomValue } from 'jotai'
+import { useHydrateAtoms } from 'jotai/utils'
+import type { ReactNode } from 'react'
+import { type ProductPageData } from '@/components/ProductPage/getProductPageData'
+
+const productPageDataAtomV2 = atom<ProductPageData | null>(null)
```
alebedev, line 7: Why do we want new atom, is shape of data different from the old one?

alebedev: Also, if we need new atom, it should be atomFamily + React context for the key, similar to recently updated `ProductDataProvider`. This setup allows to not use  `dangerouslyForceHydrate` and makes navigation both cheaper and more stable

gustaveen: Ah right you are, the data is the same format. Should we reuse it between the 2 versions then you mean? @alebedev

alebedev: Yes, we can just use `productPageDataAtom` everywhere

apps/store/src/components/ProductPage/ProductPageDataProviderV2.tsx
```diff
@@ -0,0 +1,28 @@
...
+  children: ReactNode
+  productPageData: ProductPageData
+}
+
+export function ProductPageDataProviderV2({ children, productPageData }: Props) {
+  useSyncProductPageData(productPageData)
+  return children
+}
+
+const useSyncProductPageData = (productPageData: ProductPageData) => {
```
alebedev, line 18: If atom is the same, you can now reuse hooks from original `ProductPageDataProvider` - only provider itself should be different

alebedev (approved):

apps/store/src/components/ProductPage/ProductPageDataProviderV2.tsx
gustaveen: Ah but of course! 👌

gustaveen:
### Merge activity

* **Aug 28, 7:56 AM EDT**: @gustaveen merged this pull request with [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4647).

PR #4650, opened 2024-08-27
SEB leads debugger - lead creation

lucas-larsson:
## Describe your changes

Adds SEB leads debugger that allows to create lead entries in Insurely service manually

First version returns ID of created entry without processing it further

Since this is test-only thing, disabled it in production environment


apps/store/src/app/debugger/seb-leads/createSebLead.tsx
```diff
@@ -0,0 +1,105 @@
...
+    const errorMessage = error instanceof Error ? error.message : 'Unknown error'
+
+    return {
+      errors: {
+        generic: [errorMessage],
+      },
+    }
+  }
+
+  const message = `Lead created, sebSessionId=${sebSessionId}`
```
Youakeem, line 46: Nit: What do you think about moving this logic inside the `try` block?

apps/store/src/app/debugger/seb-leads/createSebLead.tsx
```diff
@@ -0,0 +1,105 @@
...
+  console.debug(`Creating SEB lead for ${params.personalNumber}`, body)
+
+  const response = await fetch(url, {
+    method: 'POST',
+    body: body,
+    headers: headers,
+  })
+
+  if (!response.ok) {
+    const json = (await response.json()) as CreateSebLeadsResponse
```
Youakeem, line 98: Seems like we are using `await response.json()` twice. Can't we get the json right after the response and just throw if it's not ok?

Youakeem (approved):
Great work, man! I left a couple of small comments!

alebedev:
### Merge activity

* **Sep 5, 7:14 AM EDT**: @alebedev started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4650).

PR #4684, opened 2024-09-10
fix: Add hidden input value to QuickPurchaseForm edit state

gustaveen:
## Describe your changes
Forgot to add the input value to be able to submit the form

### DIFF
--- a/apps/store/src/components/QuickPurchaseV2/QuickPurchaseForm/SsnField/SsnField.tsx
+++ b/apps/store/src/components/QuickPurchaseV2/QuickPurchaseForm/SsnField/SsnField.tsx
@@ -1,5 +1,5 @@
 import { useTranslation } from 'next-i18next'
-import { useCallback, useState } from 'react'
+import { type MouseEventHandler, useCallback, useState } from 'react'
 import { Button, sprinkles, Text } from 'ui'
 import { ChangeSsnWarningDialog } from '@/components/ChangeSsnWarningDialog/ChangeSsnWarningDialog'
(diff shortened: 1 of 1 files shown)


apps/store/src/components/QuickPurchaseV2/QuickPurchaseForm/SsnField/SsnField.tsx
```diff
@@ -30,6 +30,8 @@ export const SsnField = (props: Props) => {
   if (props.defaultValue) {
     return (
       <>
+        <input type="hidden" name={props.name} value={props.defaultValue} />
```
Youakeem, line 33: Passing `value` without `onChange` will cause a React error, adding a readonly attribute should take the error away

gustaveen: Hm, I can't see the error, where did it show for you? Afaik you either use `type="hidden"` or `readOnly={true}` or?

Youakeem: I didn't run this specific PR but this is the error that usually shows up:

```
Warning: You provided a `value` prop to a form field without an `onChange` handler. This will render a read-only field. If the field should be mutable use `defaultValue`. Otherwise, set either `onChange` or `readOnly`.
```
As far as I know, `hidden` inputs still need `readOnly`

alebedev: Let's just test it

alebedev (approved):

PR #4704, opened 2024-09-16
tech: Create BasePillow UI Component

Youakeem:
## Describe your changes
- Create a new `BasePillow` in the `ui` package
- Use `BasePillow` to build the `Pillow` component in

## Justify why they are needed
Will need to use `BasePillow` in the `ui` package while building the `Card` component.

## Checklist before requesting a review

- [x] I have performed a self-review of my code

PR #4703, opened 2024-09-16
tech: Rename ProductCard to ProductGridItem

Youakeem:
## Describe your changes
Subj

## Justify why they are needed
`ProductGridItem` is the CMS version of a product card while `ProductCard` will be used in the purchase flow.

- [x] I have performed a self-review of my code

PR #4705, opened 2024-09-16
tech: Initial Card UI component

Youakeem:
Building block for ProductCards and other type of cards

https://www.figma.com/design/5kmmDdh6StpXzbEfr7WevV/Hedvig-UI-Kit?node-id=16269-14623&t=EyGtApoX5yoK0pGI-1

### DIFF
--- a/packages/ui/src/components/Card/Card.css.ts
+++ b/packages/ui/src/components/Card/Card.css.ts
@@ -1,0 +1,33 @@
+import { style } from '@vanilla-extract/css'
+import { recipe } from '@vanilla-extract/recipes'
+import { xStack, yStack } from '../../patterns'
+import { sprinkles, tokens } from '../../theme'
+
+export const cardRoot = recipe({
+  base: [
+    yStack({ gap: 'md' }),
+    {
+      borderRadius: tokens.radius.xl,
+    },
+  ],
+  variants: {
+    variant: {
+      primary: {
+        background: tokens.colors.backgroundStandard,
+        border: '1px solid',
+        borderColor: tokens.colors.borderPrimary,
+        boxShadow: tokens.shadow.card,
+      },
+    },
+    size: {
+      md: sprinkles({ padding: 'lg' }),
+    },
+  },
+
+  defaultVariants: {
+    variant: 'primary',
+    size: 'md',
+  },
+})
+
+export const cardHeader = style([xStack({ gap: 'sm' })])
--- a/packages/ui/src/components/Card/Card.stories.tsx
+++ b/packages/ui/src/components/Card/Card.stories.tsx
@@ -33,7 +33,7 @@ export const Default: Story = {
       <Card.Root variant={args.variant}>
         <Card.Header>
           <Card.Media>
-            <BasePillow shouldFallback fill={tokens.colors.amber300} />
+            <BasePillow fill={tokens.colors.amber300} />
           </Card.Media>
           <Card.Heading>
             <Card.Title>Homeowner Insurance</Card.Title>
--- a/packages/ui/src/components/Card/Card.stories.tsx
+++ b/packages/ui/src/components/Card/Card.stories.tsx
@@ -19,7 +19,7 @@ const meta: Meta<Controls> = {
     design: {
       allowFullscreen: true,
       type: 'figma',
-      url: 'https://www.figma.com/file/5kmmDdh6StpXzbEfr7WevV/Hedvig-UI-Kit?type=design&node-id=15721-17893&mode=design&t=FQrv0wWXT2aihaTB-1',
+      url: 'https://www.figma.com/file/5kmmDdh6StpXzbEfr7WevV/Hedvig-UI-Kit?type=design&node-id=16269-14623&mode=design&t=AWGrq69c6GtZiBoY-1',
     },
   },
 }
--- a/packages/ui/src/components/Card/Card.stories.tsx
+++ b/packages/ui/src/components/Card/Card.stories.tsx
@@ -1,0 +1,50 @@
+import type { Meta, StoryObj } from '@storybook/react'
+import { type ComponentProps } from 'react'
+import { tokens } from '../../theme'
+import { BasePillow } from '../BasePillow/BasePillow'
+import { Card } from './Card'
+
+type Controls = ComponentProps<typeof Card.Root>
+
+const meta: Meta<Controls> = {
+  title: 'Card',
+  component: Card.Root,
+  argTypes: {
+    variant: {
+      options: ['primary', 'secondary'],
+      control: { type: 'select' },
+    },
+  },
+  parameters: {
+    design: {
+      allowFullscreen: true,
+      type: 'figma',
+      url: 'https://www.figma.com/file/5kmmDdh6StpXzbEfr7WevV/Hedvig-UI-Kit?type=design&node-id=15721-17893&mode=design&t=FQrv0wWXT2aihaTB-1',
+    },
+  },
+}
+export default meta
+
+type Story = StoryObj<Controls>
+
+export const Default: Story = {
+  render: (args: Controls) => (
+    <div style={{ maxWidth: '400px' }}>
+      <Card.Root variant={args.variant}>
+        <Card.Header>
+          <Card.Media>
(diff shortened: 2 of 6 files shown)


Youakeem:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4705?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#4707** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4707?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4706** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4706?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4705** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4705?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#4704** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4704?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4703** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4703?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

packages/ui/src/components/Card/Card.tsx
```diff
@@ -0,0 +1,60 @@
...
+  return <Slot>{children}</Slot>
+}
+
+type HeadingProps = ComponentProps<'div'>
+
+function CardHeading({ className, children }: PropsWithChildren<HeadingProps>) {
+  return <div className={className}>{children}</div>
+}
+
+type TitleProps = ComponentProps<typeof Heading>
```
gustaveen, line 33: What about composing the full type where it's declared?
```suggestion
type TitleProps = PropsWithChildren<ComponentProps<'div'>>
```

packages/ui/src/components/Card/Card.tsx
```diff
@@ -0,0 +1,60 @@
...
+import clsx from 'clsx'
+import { type ComponentProps, type PropsWithChildren } from 'react'
+import { Heading, Text } from 'ui'
+import { cardHeader, cardRoot } from './Card.css'
+
+type RootStyleProps = RecipeVariants<typeof cardRoot>
+type RootProps = ComponentProps<'div'> & RootStyleProps
+
+function CardRoot({ variant, size, className, children }: PropsWithChildren<RootProps>) {
+  return <article className={clsx(cardRoot({ variant, size }), className)}>{children}</article>
```
gustaveen, line 12: Just wondering if we will pass classNames to these components, their use cases would be covered by variants anyway right?

gustaveen (commented):
Nice! 👌 Just some comments

packages/ui/src/components/Card/Card.css.ts
```diff
@@ -0,0 +1,33 @@
+import { style } from '@vanilla-extract/css'
+import { recipe } from '@vanilla-extract/recipes'
+import { xStack, yStack } from '../../patterns'
+import { sprinkles, tokens } from '../../theme'
+
+export const cardRoot = recipe({
```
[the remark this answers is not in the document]
Youakeem: I wasn't sure how many we'll have for each. In a later PR we add a `secondary` variant and I suppose we'll need to support different sizes as well.

packages/ui/src/components/Card/Card.tsx
Youakeem: Sure! good point, will update!

packages/ui/src/components/Card/Card.tsx
Youakeem: Not sure I understand what you mean. This pattern normally merges the internal styles with whatever is being applied by consumers.

alebedev (approved):

packages/ui/src/components/Card/Card.stories.tsx
```diff
@@ -0,0 +1,50 @@
...
+    variant: {
+      options: ['primary', 'secondary'],
+      control: { type: 'select' },
+    },
+  },
+  parameters: {
+    design: {
+      allowFullscreen: true,
+      type: 'figma',
+      url: 'https://www.figma.com/file/5kmmDdh6StpXzbEfr7WevV/Hedvig-UI-Kit?type=design&node-id=16269-14623&mode=design&t=AWGrq69c6GtZiBoY-1',
```
guilhermespopolin, line 22: Learned something new here 🏆

packages/ui/src/components/Card/Card.tsx
```diff
@@ -0,0 +1,61 @@
...
+}
+
+type HeaderProps = ComponentProps<'header'>
+
+function CardHeader({ className, children }: PropsWithChildren<HeaderProps>) {
```
guilhermespopolin, line 17: nit:
I think this would be cleaner but I guess is personal preference so feel free to ignore this :)

```suggestion
type HeaderProps = PropsWithChildren<ComponentProps<'header'>>

function CardHeader({ className, children }: HeaderProps) {
```

packages/ui/src/components/Card/Card.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+type SubtitleProps = ComponentProps<typeof Text>
+
+function CardSubtitle({ className, children }: PropsWithChildren<SubtitleProps>) {
+  return (
+    <Text color="textSecondary" className={className}>
+      {children}
+    </Text>
+  )
+}
```
guilhermespopolin, line 52: If `CardSubtitle` is just a wrapper for `Text` with a pre defined `color` props, should we spread the reaming props here? The same applies to `CartTitle`.

```suggestion
type SubtitleProps = ComponentProps<typeof Text>

function CardSubtitle({ className, children, ...delegated }: PropsWithChildren<SubtitleProps>) {
  return (
    <Text color="textSecondary" className={className} {...delegated}>
      {children}
    </Text>
  )
}

// or even

function CardSubtitle(props: PropsWithChildren<SubtitleProps>) {
  return (
    <Text color="textSecondary" {...props} />
  )
}
```

Now if for us it's very important `color` don't be modified by the consumers we can do restrict it via TS:

```suggestion
// I believe `Text` already accepts a `children` so no need to use `PropsWithChildren`
type SubtitleProps = Omit<ComponentProps<typeof Text>, 'color'>

function CardSubtitle(props: SubtitleProps ) {
  return (
    <Text color="textSecondary" {...props} />
  )
}
```

guilhermespopolin (approved):
Nice one. Couple of comments inside.

packages/ui/src/components/Card/Card.tsx
Youakeem: Sure! Looks good, will update!

packages/ui/src/components/Card/Card.tsx
Youakeem: Good point, we should spread the props, I just forgot. Will update!

packages/ui/src/components/Card/Card.tsx
gustaveen: Precisely! Just wondered if we should do that by default for all components or if it should be on a need basis? Most components we have we don't need to pass a class from the consumer right?

packages/ui/src/components/Card/Card.css.ts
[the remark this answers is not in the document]
Youakeem: We also added another size

packages/ui/src/components/Card/Card.tsx
Youakeem: yeah, I think we should do it by default. We only need clsx if we are adding some custom classes, otherwise it'll just be applied as part of `...props`

Youakeem:
### Merge activity

* **Sep 19, 3:16 AM EDT**: @Youakeem started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4705).

PR #4706, opened 2024-09-16
tech: Add IconButton component

Youakeem:
## Describe your changes
Small wrapper around the `Button` component for icon only buttons similar to the one we have in Hope.

Reuse styles needed for icon buttons.

### DIFF
--- a/packages/ui/src/components/Button/Button.tsx
+++ b/packages/ui/src/components/Button/Button.tsx
@@ -23,7 +23,6 @@ type BaseProps = {
   loading?: boolean
   Icon?: ReactNode
   iconPosition?: 'left' | 'right'
-  isIconButton?: boolean
 }
 
 export type ButtonProps<C extends React.ElementType> = PolymorphicComponentPropsWithRef<
@@ -52,7 +51,6 @@ export const Button: PolymorphicComponent = forwardRef(function Button<
     children,
     Icon,
     iconPosition = 'left',
-    isIconButton,
     ...props
   }: ButtonProps<C>,
   ref?: PolymorphicRef<C>,
@@ -85,7 +83,7 @@ export const Button: PolymorphicComponent = forwardRef(function Button<
     >
       <span className={childrenWrapper} style={{ opacity: loading ? 0 : 1 }}>
         {iconPosition === 'left' && Icon}
-        {children && <span className={clsx(!isIconButton && textWrapper)}>{children}</span>}
+        {children && <span className={clsx(textWrapper)}>{children}</span>}
         {iconPosition === 'right' && Icon}
       </span>
       {hiddenText && hiddenText}
--- a/packages/ui/src/components/Button/Button.tsx
+++ b/packages/ui/src/components/Button/Button.tsx
@@ -23,6 +23,7 @@ type BaseProps = {
   loading?: boolean
   Icon?: ReactNode
   iconPosition?: 'left' | 'right'
+  isIconButton?: boolean
 }
 
 export type ButtonProps<C extends React.ElementType> = PolymorphicComponentPropsWithRef<
@@ -51,6 +52,7 @@ export const Button: PolymorphicComponent = forwardRef(function Button<
     children,
     Icon,
     iconPosition = 'left',
+    isIconButton,
     ...props
   }: ButtonProps<C>,
   ref?: PolymorphicRef<C>,
@@ -83,7 +85,7 @@ export const Button: PolymorphicComponent = forwardRef(function Button<
     >
       <span className={childrenWrapper} style={{ opacity: loading ? 0 : 1 }}>
         {iconPosition === 'left' && Icon}
-        {children && <span className={textWrapper}>{children}</span>}
+        {children && <span className={clsx(!isIconButton && textWrapper)}>{children}</span>}
         {iconPosition === 'right' && Icon}
       </span>
       {hiddenText && hiddenText}
--- a/packages/ui/src/components/Button/Button.tsx
+++ b/packages/ui/src/components/Button/Button.tsx
@@ -83,7 +83,7 @@ export const Button: PolymorphicComponent = forwardRef(function Button<
     >
       <span className={childrenWrapper} style={{ opacity: loading ? 0 : 1 }}>
         {iconPosition === 'left' && Icon}
-        {children && <span className={clsx(textWrapper)}>{children}</span>}
+        {children && <span className={textWrapper}>{children}</span>}
         {iconPosition === 'right' && Icon}
       </span>
       {hiddenText && hiddenText}
--- a/packages/ui/src/components/Button/IconButton.tsx
+++ b/packages/ui/src/components/Button/IconButton.tsx
@@ -1,17 +1,19 @@
 import { type ComponentProps, forwardRef } from 'react'
 import { Button } from './Button'
 
-type ButtonProps = Omit<ComponentProps<typeof Button<'button'>>, 'size'>
+type ButtonProps = ComponentProps<typeof Button<'button'>>
 
-export const IconButton = forwardRef<HTMLButtonElement, ButtonProps>((props, forwardedRef) => (
-  <Button
-    variant="ghost"
-    size="small"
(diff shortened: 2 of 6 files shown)


Youakeem:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4706?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#4707** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4707?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4706** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4706?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#4705** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4705?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4704** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4704?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4703** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4703?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

packages/ui/src/components/Button/IconButton.tsx
```diff
@@ -0,0 +1,17 @@
+import { type ComponentProps, forwardRef } from 'react'
+import { Button } from './Button'
+
+type ButtonProps = Omit<ComponentProps<typeof Button<'button'>>, 'size'>
+
+export const IconButton = forwardRef<HTMLButtonElement, ButtonProps>((props, forwardedRef) => (
```
alebedev, line 6: What is intended API? I can see two options`<IconButton Icon={<ArrowForwardIcon />} />` and `<IconButton><ArrowForwardIcon /></IconButton>`

alebedev (approved):

packages/ui/src/components/Button/IconButton.tsx
Youakeem: The intended API is nesting icons inside `IconButton`

```
<IconButton>
   <ArrowForwardIcon />
</IconButton>
```

And the purpose is to reuse the same button variants and styles like `ghost` and `aspectRatio` needed for icons.

`isIconButton` is needed because in `Button` we wrap the text with `marginTop: '-0.125rem'` but that's not needed for icons.

Youakeem: I am not sure I understand what you mean by implementing it internally with `Icon`

alebedev: Something like
```
const {children, ...forwardedProps} = props
return <Button Icon={children} {...forwardedProps} />
```

Main idea here is that we already have internal way to implement icon, let's use it instead of introducing another way

Youakeem: Ah, I see now! True! I guess I was confused by the `iconPosition` check in `Button` but realised now it defaults to `left` which is fine. Will update!

packages/ui/src/components/Button/Button.tsx
```diff
@@ -23,6 +23,7 @@ type BaseProps = {
   loading?: boolean
   Icon?: ReactNode
   iconPosition?: 'left' | 'right'
+  isIconButton?: boolean
```
guilhermespopolin, line 26: Can we derivate this from `Icon` prop? Having `Icon` defined means it's an _icon button_. Or that wouldn't work? 🤔

packages/ui/src/components/Button/Button.tsx
```diff
@@ -83,7 +85,7 @@ export const Button: PolymorphicComponent = forwardRef(function Button<
     >
       <span className={childrenWrapper} style={{ opacity: loading ? 0 : 1 }}>
         {iconPosition === 'left' && Icon}
-        {children && <span className={textWrapper}>{children}</span>}
+        {children && <span className={clsx(!isIconButton && textWrapper)}>{children}</span>}
```
guilhermespopolin, line 88: Just curious: why don't we want `textWrapper` being applied in case it's a icon button? 🤔

guilhermespopolin (commented):
Nice. Some comments inside

packages/ui/src/components/Button/Button.tsx
Youakeem: It should work! Alexandr also mentioned that. I'll update it soon.

packages/ui/src/components/Button/Button.tsx
Youakeem: It adds some negative top margin to compensate for text line-height and make it look centred but it does the opposite for the icon.

packages/ui/src/components/Button/Button.tsx
```diff
@@ -83,7 +83,7 @@ export const Button: PolymorphicComponent = forwardRef(function Button<
     >
       <span className={childrenWrapper} style={{ opacity: loading ? 0 : 1 }}>
         {iconPosition === 'left' && Icon}
-        {children && <span className={textWrapper}>{children}</span>}
+        {children && <span className={clsx(textWrapper)}>{children}</span>}
```
gustaveen, line 86: nit: do we need `clsx` here? 
```suggestion
        {children && <span className={textWrapper}>{children}</span>}
```

packages/ui/src/components/Button/IconButton.tsx
```diff
@@ -0,0 +1,19 @@
+import { type ComponentProps, forwardRef } from 'react'
+import { Button } from './Button'
+
+type ButtonProps = ComponentProps<typeof Button<'button'>>
+
+export const IconButton = forwardRef<HTMLButtonElement, ButtonProps>(
```
gustaveen, line 6: Just trying to think of the different use cases for this and if we need a wrapper for the Button with Icon. This seem like a specific case where we specify size and variant. Some Iconbuttons have variant secondary, like the button for mobile menu and are larger. For the mobile menu button we use it like this, but override to follow the mobile header size.

```
<Button
    className={buttonTrigger}
    variant="secondary"
    size="medium"
    aria-label={t('NAV_MENU_DIALOG_OPEN')}
    Icon={<HamburgerOpenIcon />}
  />
```

packages/ui/src/components/Button/IconButton.tsx
```diff
@@ -0,0 +1,19 @@
...
+  ({ children, ...props }, forwardedRef) => (
+    <Button
+      variant="ghost"
+      size="icon"
+      style={{ aspectRatio: '1', height: 'auto' }}
```
gustaveen, line 11: Should we use a className instead of inline styles?

packages/ui/src/components/Button/Button.tsx
Youakeem: Nope, probably a left over, will remove!

packages/ui/src/components/Button/IconButton.tsx
Youakeem: I think it's a little tricky with IconButton. In some cases we want a clickable icon with no or small padding. Currently the `Button` component adds inline padding (left and right) in combination with aspect ratio that makes icons even with the smallest padding too big.

Youakeem: I added the `icon` size which adds small radius and padding and IconButton can take any variant, it doesn't have to be a ghost.

packages/ui/src/components/Button/IconButton.tsx
Youakeem: Sure! I initially copied from Hope which had inline styles.

Youakeem:
### Merge activity

* **Sep 19, 3:16 AM EDT**: @Youakeem started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4706).

PR #4719, opened 2024-09-17
fix(cart-toast): send 'being_checkout' event

guilhermespopolin:
## Describe your changes

Update Cart toast so we send 'begin_checkout' when "see your cart" button get's clicked

When _Checkout_ and _Cart_ page was merged and old _Cart_ we didn't cover where we should send 'begin_checkout' event:

* User add a quote to the cart
* Cart toast get's displayed
* User clicks on "see your cart"

That scenario was being covered before because cart toast would redirect _Cart_ page which then had buttons that send that event. Since _Cart_ page is now gone and Cart toast redirects directly to _Checkout_ page, we need to send that event in the context of Cart toast.


guilhermespopolin:
* **#4719** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4719?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

alebedev (approved):

PR #4724, opened 2024-09-17
Remove react-wrap-balancer and use native css rule instead

gustaveen:
## Describe your changes
Use native CSS text balancer instead of package. The one thing I though was nice with the package was that you could specify a ratio for how much to balance. But as the say in their docs:

> Note that if the native CSS balance ([text-wrap: balance](https://developer.chrome.com/blog/css-text-wrap-balance/)) is available, React Wrap Balancer will use it instead. And the `ratio` option will be ignored in that case. 

So we should just use native CSS rule instead

Improve builds and bundle by using native css solution

### DIFF
--- a/apps/store/src/appComponents/RootLayout/RootLayout.tsx
+++ b/apps/store/src/appComponents/RootLayout/RootLayout.tsx
@@ -1,6 +1,5 @@
 import clsx from 'clsx'
 import { Suspense, type PropsWithChildren } from 'react'
-import { Provider as BalancerProvider } from 'react-wrap-balancer'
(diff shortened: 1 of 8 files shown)


gustaveen:
* **#4724** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4724?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

alebedev (approved):
This is great!

guilhermespopolin (approved):

PR #4764, opened 2024-09-23
GEN-2663 - feat: added new cart toast

guilhermespopolin:
## Describe your changes

Added new Cart Toast

https://github.com/user-attachments/assets/1ca6474f-1134-4f9e-bf0c-fee782db3252

* `@floating-ui/react` library is being used so we can easily place the notification ballon and its arrow
* `CartToastProvider` is a workaround the let us control the appearance of the new toast and in "global" way. I think the show state of that piece of UI can come from the `shopSession`. We could list for changes in `shopSession` with `useShopSession` and based on the length of cart entries field, decide among showing or not the notification. We can't do that yet for the following reasons:
* It's better to render the new cart toast closer to `ShoppingMenuCartItem` because we need to anchor those two together.
* Since `ShoppingMenuCartItem` get's shared among old and new price calculator pages, derivate show state based on shop session would make new cart toast to be displayed via old flow as well.

With that said I decided to keep that state in a jotai atom. But we need to be aware that we now that a page lever jotai provider for price calculator, which  `ShoppingMenuCartItem` doesn't rely on. That's the reason behind the `CartToastProvider`: make sure we read "show" atom from the global state despite being rendered bellow a different jotai provider.

Part of hedvig-com redesign.


guilhermespopolin:
* **#4764** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4764?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

apps/store/package.json
```diff
@@ -30,6 +30,7 @@
     "@emotion/is-prop-valid": "1.3.0",
     "@emotion/react": "11.13.3",
     "@emotion/styled": "11.13.0",
+    "@floating-ui/react": "0.26.24",
```
alebedev, line 33: I'm not sure about the library. We have components for popover already. Is there anything wrong it them? If this is about arrow, I doubt that it's worth extra dependency

apps/store/src/components/Header/ShoppingCartMenuItem/CartToast.css.ts
```diff
@@ -0,0 +1,30 @@
...
+  },
+  to: {
+    transform: 'translateY(0)',
+    opacity: 1,
+  },
+})
+
+export const TOAST_BG_COLOR = theme.colors.signalGreenFill
+export const toast = style({
+  width: 'min(35ch, 90vw)',
```
alebedev, line 17: Nit: why `ch` over `rem`?

apps/store/src/components/Header/ShoppingCartMenuItem/CartToast.tsx
```diff
@@ -0,0 +1,60 @@
...
+    ],
+    transform: false,
+    whileElementsMounted: autoUpdate,
+  })
+
+  if (!show) return null
+
+  return (
+    <output ref={setFloatingElement} className={toast} role="status" style={floatingStyles}>
+      <Text size="xs"> {t('CART_TOAST_HEADING')}</Text>
```
alebedev, line 46: Nit: is space before `{` significant?

apps/store/src/components/Header/ShoppingCartMenuItem/ShoppingCartMenuItem.tsx
```diff
@@ -1,25 +1,38 @@
...
 import Link from 'next/link'
+import { useState } from 'react'
 import { useShopSession } from '@/services/shopSession/ShopSessionContext'
 import { useRoutingLocale } from '@/utils/l10n/useRoutingLocale'
 import { PageLink } from '@/utils/PageLink'
 import { ShoppingBagIcon } from './ShoppingBagIcon'
 import { cartLink, wrapper } from './ShoppingCartMenuItem.css'
 
-export const ShoppingCartMenuItem = () => {
+const DynamicCartToast = dynamic(async () => {
```
alebedev, line 11: Not sure it's a good place for code splitting. Try testing when emulating slow network - is there extra delay before toast appears?

Speaking of this, is there a way to move toast into price calculator components and use cart menu item as positional anchor?

alebedev (commented):
Few comments, main one is if extra library makes sense

apps/store/package.json
guilhermespopolin: You mean radix Popover? It also include necessary features like anchoring the floating piece to an element, rendering an arrow, offset and colision detection but I think it's not intended to be used to implement something like a toast notification and it includes all sort of keyboard interactions and focus management that we don't want for a toast notification. By definition they are:

> [...] non-interactive, passive, and asynchronous short messages for users. Generally they are used as an interface feedback pattern for informing the user about the results of an action.

I don't think we can "turn off" those for radix Popover and even we could do I'd would be like a workaround. I guess we could implement our own following guides like [this one](https://web.dev/articles/building/a-toast-component) but we'd need to handle the complexity of place it in the correct place regardless viewport screen size.

alebedev: OK, sounds reasonable

apps/store/src/components/Header/ShoppingCartMenuItem/CartToast.css.ts
guilhermespopolin: Not a strong reason. For text heavy blocks, sometimes I like to measure using `ch` as there are some guidelines/best practices among maximum number of characters per line in order to enable a comfortable reading. I think WCAG recommends 80 characters per line or fewer, but 60-70 per line on desktop and 40-50 per line on mobile seems also to be a good number.

apps/store/src/components/Header/ShoppingCartMenuItem/ShoppingCartMenuItem.tsx
guilhermespopolin: I tested and it seems to work fine. The reasoning behind it that I didn't want to slow down product pages that don't care about this new cart toast as we'll be rolling out new price calculator pages in a gradual way. Maybe that's why you are asking about moving the new toast into price calculator components: that way we'd be rendering it only it needs to be rendered.

Now, I can think on a way to make this work but it's not pretty tho: I could get the cart menu item element using DOM API inside new cart toast; `getElementById` or something like this. What do you think?

alebedev: I'm fine with `getElementById()` in this case. It's still coupled, but at least dependency on menu side is tiny - just an ID

apps/store/src/features/priceCalculator/CartToast.css.ts
```diff
@@ -0,0 +1,30 @@
...
+export const TOAST_BG_COLOR = theme.colors.signalGreenFill
+export const toast = style({
+  width: 'min(35ch, 90vw)',
+  maxWidth: 'max-content',
+  borderRadius: theme.radius.sm,
+  paddingBlock: theme.space.xs,
+  paddingInline: theme.space.sm,
+  backgroundColor: TOAST_BG_COLOR,
+  pointerEvents: 'none',
+  lineHeight: '1.2',
```
alebedev, line 24: Why do we need it? line-height work in non-obvious ways, can this be handled with top / bottom paddings instead?

alebedev (approved):
Looks good

guilhermespopolin:
### Merge activity

* **Sep 24, 11:02 AM EDT**: @guilhermespopolin started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4764).

PR #4779, opened 2024-09-24
fix(priceIntentAtoms): avoid triggering the creation of multiple intents for the same product

guilhermespopolin:


alebedev (approved):
Great catch!

PR #4806, opened 2024-09-30
Adjust PriceCalculator layout

gustaveen:
## Describe your changes
Adjust PriceCalculator layout 
- Make sure layout is visually centered in desktop and not expanding viewport height
- Adjust position in mobile
- Replace inline styles with classnames

### Before
Screenshot 2024-09-30 at 14 50 09

Screenshot 2024-09-30 at 14 49 58

### After
Screenshot 2024-09-30 at 12 08 47

Screenshot 2024-09-30 at 14 58 47

Match design

### DIFF
--- a/apps/store/src/features/priceCalculator/InsuranceDataForm/InsuranceDataForm.css.ts
+++ b/apps/store/src/features/priceCalculator/InsuranceDataForm/InsuranceDataForm.css.ts
@@ -1,0 +1,25 @@
+import { style, styleVariants } from '@vanilla-extract/css'
+import { responsiveStyles } from 'ui'
+import { HEADER_HEIGHT_DESKTOP } from '@/components/Header/Header.css'
+
+export const gdprLink = style({
+  marginTop: '-1rem', // Offset from default gap of 2rem
+})
+
+export const formSection = styleVariants({
+  base: {
+    marginTop: '3.75rem',
+  },
+  ssn: {
+    ...responsiveStyles({
+      lg: {
+        position: 'absolute',
+        top: `calc(50%)`,
+        transform: `translateY(-50%)`,
+        maxWidth: '23rem',
+        // Visually center form section
+        marginTop: `calc(-1 * (${HEADER_HEIGHT_DESKTOP} / 2))`,
+      },
+    }),
+  },
+})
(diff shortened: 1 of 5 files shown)


gustaveen:
* **#4806** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4806?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

apps/store/src/features/priceCalculator/PriceCalculatorCmsPage.css.ts
```diff
@@ -6,7 +6,7 @@ import { HEADER_HEIGHT_MOBILE } from '@/components/Header/HeaderMenuMobile/Heade
 export const pageGrid = style({
   display: 'grid',
   gridTemplateRows: 'min-content 1fr',
-  gap: tokens.space.md,
+  columnGap: tokens.space.md,
```
gustaveen, line 9: Make sure we don't add row gap that makes the sections to tall

apps/store/src/features/priceCalculator/InsuranceDataForm/InsuranceDataForm.css.ts
```diff
@@ -0,0 +1,25 @@
...
+  marginTop: '-1rem', // Offset from default gap of 2rem
+})
+
+export const formSection = styleVariants({
+  base: {
+    marginTop: '3.75rem',
+  },
+  ssn: {
+    ...responsiveStyles({
+      lg: {
```
alebedev, line 15: Centering with absolute, transform and negative margins feels hacky. Can we do it with flex or grid?

apps/store/src/features/priceCalculator/InsuranceDataForm/InsuranceDataForm.tsx
```diff
@@ -84,7 +80,10 @@ export function InsuranceDataForm() {
       )
     }
     return (
-      <div key={section.id} className={yStack({ gap: 'lg' })} style={sectionStyle}>
+      <div
+        key={section.id}
+        className={clsx(formSection.base, isSsnSection && formSection.ssn, yStack({ gap: 'lg' }))}
```
alebedev, line 85: Should `yStack` come last for priority reasons?  It's usually first

alebedev (approved):
Looks good! Few comments

apps/store/src/features/priceCalculator/InsuranceDataForm/InsuranceDataForm.tsx
gustaveen: Shouldn't overlap but yeah that might be a good practice 👍

apps/store/src/features/priceCalculator/InsuranceDataForm/InsuranceDataForm.css.ts
gustaveen: We should use flex or grid for sure where possible. This is a bit of a edge case though where it's only one section that should be positioned vertically centered and it's nested in several containers that we would need to modify. But I'll see if it's possible to simplify 👍 @gustaveen

PR #4827, opened 2024-10-03
GEN-2651 - feat(price-calculator): enable addition of multiple products in purchase summary

guilhermespopolin:
## Describe your changes

Enable the addition of multi products (E.g: Pet and Car).

https://github.com/user-attachments/assets/699e1535-50d4-42a1-ba0a-4739024b9ef6

We can now from storefront if a particular product is a multi-product or not via `Product.multiple` field. However we also want to control we user should be presented with "Add another" button on the FE side. For that I'm using the price template (`TemplateV2`) so at the end, before presenting that option we check: `Product.multiple && priceTemplate.addMutliple`.

This is a first attempt which requires a full page refresh for new price calculator. Will be investigating a better way to implement this that doesn't require a full page refresh. 

Make it easier to add multiple products when enabled.

### DIFF
--- a/apps/store/src/components/ProductPage/PurchaseForm/priceIntentAtoms.ts
+++ b/apps/store/src/components/ProductPage/PurchaseForm/priceIntentAtoms.ts
@@ -1,8 +1,8 @@
 import { useApolloClient } from '@apollo/client'
 import { datadogLogs } from '@datadog/browser-logs'
-import { atom, useAtomValue, useSetAtom, useStore } from 'jotai'
+import { atom, useAtom, useAtomValue, useSetAtom, useStore } from 'jotai'
 import { atomFamily } from 'jotai/utils'
-import { useCallback, useEffect, useState } from 'react'
+import { useCallback, useEffect } from 'react'
 import { useProductData } from '@/components/ProductData/ProductDataProvider'
 import {
   priceTemplateAtom,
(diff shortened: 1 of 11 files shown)


apps/store/src/components/ProductPage/PurchaseForm/priceIntentAtoms.ts
```diff
@@ -293,15 +293,19 @@ export const usePriceIntentId = (): string => {
 export const useResetPriceIntent = () => {
   const shopSessionId = useShopSessionId()
   const apolloClient = useApolloClient()
-  const templateName = usePriceTemplate().name
+  const priceTemplate = usePriceTemplate()
   const [, setSelectedOffer] = useSelectedOffer()
   const setCurrentPriceIntentId = useSetAtom(currentPriceIntentIdAtom)
 
+  const priceIntentService = useMemo(
```
alebedev, line 300: `useMemo` is tricky. It works the way we want it today, but Reacts docs say developers should not rely on memo not throwing away its cache and recomputing it later. See more here: https://github.com/alexreardon/use-memo-one

Can we do it with ref and manual check for changes on every render?

apps/store/src/features/priceCalculator/PurchaseSummary.tsx
```diff
@@ -1,18 +1,33 @@
...
+  const priceTemplate = usePriceTemplate() as TemplateV2
+  const resetPriceIntent = useResetPriceIntent()
+
+  const handleAddMore = useCallback(() => {
+    resetPriceIntent()
+    // This is a workaround to reset the price intent for the new price calculator
+    window.location.reload()
+  }, [resetPriceIntent])
+
+  const showAddMoreButton = offer.product.multiple && priceTemplate.addMultiple
```
alebedev, line 30: Can you move it next to usage?

alebedev (approved):
Looks good, some comments inside

apps/store/src/features/priceCalculator/PurchaseSummary.tsx
guilhermespopolin: Not sure If I'm following..you mean creating an expectific button for it and add this verification there?

guilhermespopolin:
### Merge activity

* **Oct 7, 5:30 AM EDT**: @guilhermespopolin started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4827).

PR #4841, opened 2024-10-09
Feature/exclude new terms from debugger page

AlexSJ1990:

### DIFF
--- a/apps/store/src/app/debugger/terms/[productName]/page.tsx
+++ b/apps/store/src/app/debugger/terms/[productName]/page.tsx
@@ -18,6 +18,16 @@ const LOCALES = ['sv-SE', 'en-SE'] as const
 async function ProductTermsPage({ params }: Props) {
   const apolloClient = setupApolloClient({ locale: toRoutingLocale(LOCALES[0]) }).getApolloClient()
   const { productName } = params
+
+  const excludedVariants = [
+    'SE_APARTMENT_BRF_BAS',
+    'SE_APARTMENT_BRF_MAX',
+    'SE_APARTMENT_RENT_BAS',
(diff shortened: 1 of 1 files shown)


apps/store/src/app/debugger/terms/page.tsx
```diff
@@ -18,10 +18,21 @@ function TermsNavigationPage() {
...
+    "SE_HOUSE_MAX-HEDVIG-null-1"
+];
+
 async function AvailableProductSelector() {
-  const availableProducts = await fetchGlobalProductMetadata({
+  let availableProducts = await fetchGlobalProductMetadata({
     apolloClient: setupApolloClient({ locale: locales['sv-SE'].routingLocale }).getApolloClient(),
   })
+
+  availableProducts = availableProducts.filter((product) => !excludedProducts.includes(product.id))
```
alebedev, line 35: You probably want to exclude variants, not products.  For example, we'll still see Rent, but only middle tier of it

apps/store/src/app/debugger/terms/[productName]/page.tsx
```diff
@@ -18,6 +18,16 @@ const LOCALES = ['sv-SE', 'en-SE'] as const
 async function ProductTermsPage({ params }: Props) {
   const apolloClient = setupApolloClient({ locale: toRoutingLocale(LOCALES[0]) }).getApolloClient()
   const { productName } = params
+
+  const excludedVariants: Array<string> = [
```
Comment (alebedev) [inline apps/store/src/app/debugger/terms/[productName]/page.tsx:22]: Nit - use `Set` instead of array. And you probably don't need to declare type explicitly

alebedev (approved):
Looks good now!

PR #4860, opened 2024-10-16
fix(new-price-calculator): proper reset after adding offer to cart

guilhermespopolin:
## Describe your changes

_Reset_ new price calculator after adding offer to cart

### The issue

As can be seen in the following video, when you navigate back to price calculator page after adding an offer to the cart you end up in a misleading state of price calculator which might make you add multiple insurances for the same dog. Observe that for being able to add multiple products to the same dog, the user would need to take some actions that would make a **different** offer to be added to the cart like changing deductible for example. Changing _start date_ doesn't causes that, that's why you can still see only one offer in your cart when _add to cart_ button get's clicked:

https://github.com/user-attachments/assets/3df963db-ccf4-48bc-a344-88d32489ca8e

The same occurs for non multi products as well - like rent. The difference tho is that you can only have a single product of that type in your cart in all scenarios. So repeating the steps for rent would show a error dialog, which makes sense.

### The solution

What this solution does is:

* Update new price calculator so it works the same way as the old one when it comes about price intent reseting: whenever a new product get's added to the cart we create a new price intent for that product.
* `PurchaseSummary` is not rendered as a step of `PurchaseFormV2`. Instead we alternate between then inside `PriceCalculatorCmsPageContent`. With that, reseting the price intent and synchronizing price calculator atoms which it would cause any issues with `PurchaseSummary`.
* Code scout: since now old and new price calculator have the same "reset" price intent logic we don't need `useResetPriceIntent` hook anymore so I'm deleting it.

https://github.com/user-attachments/assets/4e8c085f-ecde-4c10-b8d7-43d74120574b

### DIFF
--- a/apps/store/src/components/ProductPage/PurchaseForm/priceIntentAtoms.ts
+++ b/apps/store/src/components/ProductPage/PurchaseForm/priceIntentAtoms.ts
@@ -4,10 +4,7 @@ import { atom, useAtom, useAtomValue, useSetAtom, useStore } from 'jotai'
 import { atomFamily } from 'jotai/utils'
 import { useCallback, useEffect } from 'react'
 import { useProductData } from '@/components/ProductData/ProductDataProvider'
-import {
-  priceTemplateAtom,
-  usePriceTemplate,
-} from '@/components/ProductPage/PurchaseForm/priceTemplateAtom'
+import { priceTemplateAtom } from '@/components/ProductPage/PurchaseForm/priceTemplateAtom'
 import { useSelectedOffer } from '@/components/ProductPage/PurchaseForm/useSelectedOffer'
 import { useCartEntryToReplace } from '@/components/ProductPage/useCartEntryToReplace'
 import {
@@ -19,7 +16,7 @@ import {
 import { setupForm } from '@/services/PriceCalculator/PriceCalculator.helpers'
 import type { Form } from '@/services/PriceCalculator/PriceCalculator.types'
 import { priceIntentServiceInitClientSide } from '@/services/priceIntent/PriceIntentService'
-import { useShopSession, useShopSessionId } from '@/services/shopSession/ShopSessionContext'
+import { useShopSession } from '@/services/shopSession/ShopSessionContext'
 import { getOffersByPrice } from '@/utils/getOffersByPrice'
 import { getAtomValueOrThrow } from '@/utils/jotaiUtils'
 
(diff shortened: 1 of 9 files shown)


guilhermespopolin:
* **#4860** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4860?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

apps/store/src/components/ProductPage/PurchaseForm/priceIntentAtoms.ts
```diff
@@ -291,22 +287,3 @@ export const usePriceIntentId = (): string => {
   }
   return priceIntentId
 }
-
-export const useResetPriceIntent = () => {
```
guilhermespopolin, line 295: We now reset price intent whenever adding a product to the cart as part of `useSyncPriceIntentState`. With that way of doing thins this hook becomes obsolete.

apps/store/src/features/priceCalculator/PurchaseFormV2/PurchaseSummary/PurchaseSummary.tsx
```diff
@@ -28,13 +29,15 @@ export function PurchaseSummary({ className }: { className?: string }) {
   const locale = useRoutingLocale()
   const offer = useSelectedOfferValueOrThrow()
   const priceTemplate = usePriceTemplate() as TemplateV2
-  const resetPriceIntent = useResetPriceIntent()
+  const setShowPurchaseSummary = useSetAtom(priceCalculatorShowPurchaseSummaryAtom)
```
alebedev, line 32: Please comment why it's not just another `priceCalculatorStep`. And how `selectedOffer` is the only thing that refers to previous priceIntent that's no longer current

guilhermespopolin: Updated

alebedev (approved):
Looks good after you explained it to me. Maybe we can rethink how we name and structure out atoms here later to be more clear. But for now, if it works, it's good

Youakeem (approved):

guilhermespopolin:
### Merge activity

* **Oct 21, 8:48 AM EDT**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4860).

PR #4882, opened 2024-10-22
fix(new-price-calculator): stop deriving purchase summary state from selected offer

guilhermespopolin:
## Describe your changes

* We need to keep an active query for `PriceIntent` for all price calculator context. So I'm moving `useSyncPriceIntentState` call from `PurchaseFormV2` to `PriceCalculatorCmsPageContent`
* Stop getting offer data used by `PurchaseSummary` form `selectedOfferAtom`. That one should be in sync with price intent changes to avoid unexpected behaviors.
* It's hard to know which offer was just added to the cart while reseting price intent whenever something get's added. We do need to reset price intent whenever something get's added to the cart so I'm replacing _show purchase summary_ state with an _added offer_ state which `PurchaseSummary` can use for its UI. Showing/hiding `PurchaseSummary` now is about changing that state. 

With that said this is how it works now:

* Price calculator get's initialized with price intent (P1)
* User adds it to the cart
* We store the offer that got added
* We create a new price intent P2 and synchronize all atoms with it
* We use the stored offer to render `PurchaseSummary`

The issue

https://github.com/user-attachments/assets/10a149f6-ced8-495d-8238-75d5d27412a2

`PurchaseHero` relies on `useIsPriceIntentStateReady()` which check the presence of a price intent. Since `useSyncPriceIntentState` was being called inside `PurchaseFormV2` we keep an active query for Price Intent until `PurchaseFormV2` is mounted. When something get's added to the cart we reset Price Intent and swap what's gets rendered at the right section: `PurchaseFormV2` --> `PurchaseSummary` marking the end of the flow. However `useIsPriceIntentStateReady()` will never return `true` because we don't sync priceIntentAtoms with the new Price Intent because `PurchaseFormV2` was unmounted. This causes the `PurchaseHero` never being rendered.

### DIFF
--- a/apps/store/src/features/priceCalculator/PriceCalculatorCmsPage/PriceCalculatorCmsPageContent/PriceCalculatorCmsPageContent.tsx
+++ b/apps/store/src/features/priceCalculator/PriceCalculatorCmsPage/PriceCalculatorCmsPageContent/PriceCalculatorCmsPageContent.tsx
@@ -1,7 +1,8 @@
 'use client'
 
 import { useAtomValue } from 'jotai'
-import { priceCalculatorShowPurchaseSummaryAtom } from '@/features/priceCalculator/priceCalculatorAtoms'
+import { useSyncPriceIntentState } from '@/components/ProductPage/PurchaseForm/priceIntentAtoms'
+import { priceCalculatorAddedOffer } from '@/features/priceCalculator/priceCalculatorAtoms'
 import { PurchaseSummary } from '@/features/priceCalculator/PurchaseFormV2/PurchaseSummary/PurchaseSummary'
 import { useResponsiveVariant } from '@/utils/useResponsiveVariant'
 import { ProductHeroV2 } from '../../ProductHeroV2/ProductHeroV2'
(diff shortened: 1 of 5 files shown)


guilhermespopolin:
* **#4882** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4882?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

apps/store/src/features/priceCalculator/PriceCalculatorCmsPage/PriceCalculatorCmsPageContent/PriceCalculatorCmsPageContent.tsx
```diff
@@ -17,9 +18,11 @@ import {
 
 export function PriceCalculatorCmsPageContent() {
   const variant = useResponsiveVariant('lg')
-  const showPurchaseSummary = useAtomValue(priceCalculatorShowPurchaseSummaryAtom)
+  const addedOffer = useAtomValue(priceCalculatorAddedOffer)
```
gustaveen, line 21: nit: maybe clarify name
```suggestion
  const addedOfferToCart = useAtomValue(priceCalculatorAddedOffer)
```

gustaveen (approved):
Good job! 🥳

guilhermespopolin:
### Merge activity

* **Oct 23, 3:24 AM EDT**: A user merged this pull request with [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4882).

PR #4883, opened 2024-10-22
Fix issue where ref to hidden product pillow would cause sticky header to be visible

gustaveen:
## Describe your changes
Since we hide the product pillow when `isCondensedProductHero` is true, the behaviour became inconsistent when using a ref to check if you had scrolled past the pillow since the height would be 0 of that empty div. 

We can rather simplify it by just checking scroll position in StickyHeader instead.

### Issue:
StickyHeader is visible when user hasn't scrolled
Screenshot_20241022-091927


guilhermespopolin (approved):
Nicely done. Tested on mobile and it works great.

gustaveen:
### Merge activity

* **Oct 23, 3:41 AM EDT**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4883).

PR #4897, opened 2024-10-24
RND-939 - refact(checkout): improve error handling

guilhermespopolin:
## Describe your changes

* Keep polling in case of _NetworkErrors_ instead of failing sign early. 
* Leverages only `BankIdDialog` to handle error cases. That way we'll have a standard way to handle error during sign.

https://github.com/user-attachments/assets/295b0907-4c8c-43a6-96cc-4ba03d687be3

A better UI for handling _no internet_ scenarios will be added next.

We want to keep pooling if network errors happen during sign instead of showing error dialog, by doing that we can:
* Still redirect users to _connect payment_ page when internet connection get's back
* Avoid confusing cases where we show an error dialog during signing but internally user has successfully signed.

We currently show double dialogs for error cases during signing which is confusing and unnecessary:

https://github.com/user-attachments/assets/56a85bdf-16c6-4109-8d1d-0355d1e0b7c8

I think `BankIdDialog` should be the one responsible for handling error cases.


guilhermespopolin:
* **#4897** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4897?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

apps/store/src/services/bankId/useBankIdCheckoutSign.ts
```diff
@@ -114,11 +114,11 @@ const useBankIdCheckoutSignApi = ({ dispatch }: Options) => {
               }
             },
             onError(error) {
-              signingResult.stopPolling()
               if (error.networkError) {
                 bankIdLogger.warn('SigningQuery | Network error', { error: error.message })
```
Youakeem, line 118: I think we need to top polling if it's not a network error, otherwise we'll keep hitting a server that's already struggling

guilhermespopolin: But we'll stop pulling if it's not a network error. It's on the else statement below.

Youakeem: True, my bad, I missed it! Great work!

Youakeem (approved):

guilhermespopolin:
### Merge activity

* **Oct 25, 10:01 AM EDT**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4897).

PR #4904, opened 2024-10-28
Add Dialog.Content closeOnPointerDownOutside prop

alebedev:
## Describe your changes

Allow to close dialog on click outside via prop. Use it in information dialogs about tier coverage and deductible in Price Calculator page

Since many of current dialogs are modals which we don't want to close accidentally, prop defaults to false


alebedev:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4904?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#4904** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4904?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#4903** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4903?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

packages/ui/src/components/Dialog/Dialog.tsx
```diff
@@ -10,11 +11,25 @@ type ContentProps = {
...
   frostedOverlay?: boolean
   centerContent?: boolean
+  closeOnPointerDownOutside?: boolean
+  asChild?: boolean
 }
 
 const Content = (props: ContentProps) => {
   const handleClose = () => props.onClose?.()
 
+  const handleInteractOutside = useCallback(
```
Youakeem, line 21: I am not sure if `useCallback` is needed

alebedev: Good point, we did not have it

packages/ui/src/components/Dialog/Dialog.tsx
```diff
@@ -10,11 +11,21 @@ type ContentProps = {
...
+  asChild?: boolean
 }
 
 const Content = (props: ContentProps) => {
   const handleClose = () => props.onClose?.()
 
+  const handleInteractOutside = (
+    event: Parameters<NonNullable<DialogContentProps['onInteractOutside']>>[0],
+  ) => {
+    if (event.type !== 'dismissableLayer.pointerDownOutside' || !props.closeOnPointerDownOutside) {
```
Youakeem, line 24: I am not sure about defaulting to false, I feel like the natural behaviour is to click outside to close, are we sure we want to have this as the default behaviour?

On a side note: Don't think we really care about what the event is called we can omit that from the prop name.

alebedev: If we had no existing usages, I would agree. But allowing existing dialogs like BankID to close by random click outside might break things in unexpected way. So, making it opt-in is safer

Youakeem: Can't we op into the "modal" behaviour instead?

alebedev: Do you mean `Dialog.Root.modal` (https://www.radix-ui.com/primitives/docs/components/dialog#root). It's already on by default and does not affect mouse interaction, unfortunately - only keyboard focus

So, I'd love to use built-in library feature for this, but don't see one that does what we want

Youakeem (approved):

alebedev:
### Merge activity

* **Oct 29, 6:00 AM EDT**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4904).

PR #4919, opened 2024-10-29
GEN-2777 / Add tagline to ProductHero

gustaveen:
## Describe your changes
Add tagline to ProductHero

Note that we should not set default text alignment in in Heading since we cant make it inherit value from a wrapper then. The browser default is `start` so it will still default to left if no align value is passed

Potential improvement, add as a field in storybook on price calculator page if it should differ from tagline from product metadata 

Screenshot 2024-10-29 at 17 19 40

Screenshot 2024-10-29 at 17 19 48

Match design


gustaveen:
* **#4919** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4919?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

alebedev (approved):
Looks good

packages/ui/src/components/Heading/Heading.tsx
```diff
@@ -35,7 +35,7 @@ export function Heading({
   className,
   color = 'textPrimary',
   variant = 'standard.32',
-  align = 'left',
+  align,
```
guilhermespopolin, line 38: Agreed. We just need to make sure this doesn't break uses cases that was relying on this default setting.

guilhermespopolin (approved):
Have changed that file recently by adding `DiscountTooltip`. I think it should be a simple conflict resolution but ping me if that proves to be wrong :)

packages/ui/src/components/Heading/Heading.tsx
gustaveen: For sure! I looked around on the site to try to find cases, looked fine. But might have missed some so let's keep eyes open for that 👍

PR #4916, opened 2024-10-29
GEN-2773 - fix(InputDay): make sure it's required and invalidades the form when empty

guilhermespopolin:
## Describe your changes

- Pet flow: Mark _birth date_ field as required.

https://github.com/user-attachments/assets/d782bc5c-fe2a-48cb-bc5d-4e484fc55778

## Justify why they are needed

We want to make _pet birth date_ a required field and invalidate the form when submitting it with that field empty, as we do for all fields of _Your Dog_ section.

https://github.com/user-attachments/assets/eb8b39b7-bebf-40a5-a8e0-76c4268d854e

PR #4924, opened 2024-10-30
GEN-2773 - feat(InputDay): improve keyboard interaction

guilhermespopolin:
## Describe your changes

* `InputDay`: improve keyboard interaction
* Updated focus style to match with other fields

The idea for this PR was to change `InputDay` so it gets opened when focused via keyboard as mentioned [here](https://github.com/HedvigInsurance/racoon/pull/4916#discussion_r1820975522) as an improvement of keyboard interaction for that field. However I couldn't accomplish that without causing more problems and based on WAI/ARIA [patterns](https://www.w3.org/WAI/ARIA/apg/patterns/menu-button/) - menu button would be the one that closely relates with how this field works - the way it works today "makes sense" and would provide good accessibility support. So I've just did some small changes to make the keyboard interaction to work again.

FYI @alebedev


guilhermespopolin:
* **#4924** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4924?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

apps/store/src/components/InputDay/InputDay.tsx
```diff
@@ -115,9 +115,6 @@ export const InputDay = (props: InputDayProps) => {
           sideOffset={-4}
           alignOffset={-4}
         >
-          {/* Fix issue on iOS where month dropdown is focused and can't be tapped */}
-          <div tabIndex={1} aria-hidden={true} />
```
guilhermespopolin, line 119: It was tested again on iphone and the issue doesn't happen anymore. I'm removing this so date picker get's focused when `InputDay` field get's triggered open. That's necessary for keyboard navigation for that field to work.

apps/store/src/components/InputDay/InputDay.tsx
```diff
@@ -144,6 +141,8 @@ export const InputDay = (props: InputDayProps) => {
         {...inputProps}
         id={inputId}
         className={visuallyHidden}
+        // This input is only used for form submission and should not be focusable
+        tabIndex={-1}
```
guilhermespopolin, line 145: Fixes this:
image

PR #4975, opened 2024-11-07
set `isStudent` flag to `false` on Accident

lucas-larsson:
## Describe your changes

- Set `isStudent` flag to always be `false`

- We want to do this change to enable different behaviour on the backend. 
- Without going into much detail here, we want to use the same Product name in SF to always create a student tier offer if not `isStudent` set to `false`

## Anything else?
- I am not sure this is the right way of doing this please feel free to suggest a different way of doing this.
- Do we not sell Student Tier Accident on hedvig.com?


lucas-larsson:
Closed since we will not have it.

PR #4976, opened 2024-11-07
Use regular header menu on intermediate cart screen

gustaveen:
## Describe your changes
Set a data attribute in `PriceCalculatorCmsPageContent` so we can use the state `addedOfferToCart` to determine wether to hide the menu items or not. 

### After
https://github.com/user-attachments/assets/da59c339-7ffa-4d74-9c28-7f6353ad7e13

https://github.com/user-attachments/assets/31a63cc2-ce85-4025-8447-de6ac78f80a5

For the "fake toast" screen, we want to show the regular header menu with the cart icon since you completed the price calculation and have added the product to cart. 


gustaveen:
* **#4976** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4976?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

guilhermespopolin (commented):
Nice usage of `:has` selector. Just wondering if we want it in a way that's not dependent on the `tag` used. E.g: `body:has([data-hide-menu=true])`. That way one could easily change HTML structure of price calculator page without causing any unintended issues with this functionality. I've checked this [article](https://www.joshwcomeau.com/css/has/#global-detection-6) and it seems we don't need to worry a bunch with performance as browsers are very good to solve those selectors real quick. On his article he add a "disable scroll" feature for modals. Observe he uses `html:has(...)` for the selector and as he [mentions](https://www.joshwcomeau.com/css/has/#global-detection-6:~:text=What%20about%20performance,to%20worry%20about.) there practically no performance issues with that. 

With that said, it seems there are some issues with the navigation items pop up 🤔 

image

gustaveen:
> Nice usage of `:has` selector. Just wondering if we want it in a way that's not dependent on the `tag` used. E.g: `body:has([data-hide-menu=true])`. That way one could easily change HTML structure of price calculator page without causing any unintended issues with this functionality. I've checked this [article](https://www.joshwcomeau.com/css/has/#global-detection-6) and it seems we don't need to worry a bunch with performance as browsers are very good to solve those selectors real quick. On his article he add a "disable scroll" feature for modals. Observe he uses `html:has(...)` for the selector and as he [mentions](https://www.joshwcomeau.com/css/has/#global-detection-6:~:text=What%20about%20performance,to%20worry%20about.) there practically no performance issues with that.
> 
> With that said, it seems there are some issues with the navigation items pop up 🤔
> 

Thanks! That was my thought, solve as much as possible with CSS 🥳 No need to use the main selector, just maybe something to point to where it's being set :) Mostly I followed to same pattern as we have in other places with selecting body, something that one level down from html feels more logical. 

Good find! I'l have a look at the sub menu overlap 👌

apps/store/src/components/Header/Header.css.ts
```diff
@@ -204,7 +212,7 @@ export const navigationContent = style({
 
       selectors: {
         // Make sure the menu item to the right doesn't expand outside the viewport
-        [`${navigationItem}:nth-child(2) &`]: {
+        [`${navigationItemSupportMenu} &`]: {
```
gustaveen, line 215: @guilhermespopolin this fixes the issue since we have a `backLink` on price calculator page, targeting nth-child caused picking the wrong element. Here we should target the support menu item since that's the one we want specificallty

guilhermespopolin: Aha, got it 🕵🏻

guilhermespopolin (approved):
🚢

gustaveen:
### Merge activity

* **Nov 11, 3:49 AM EST**: [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/racoon/4976) rebased this pull request as part of a merge.

